PR #1, opened 2019-01-24
🌱 Create hello world, CI, coverage and linting setup

juliaogris:
### Changes Proposed in this PR:
- Create hello world program
- Create tests for hello world
- Create CI setup
- Create coverage setup
- Create Linter setup
- Update `.gitignore` with coverage output
- Update `README.md` with prerequisite and build sections

### DIFF
--- a/0_hello_world/main.go
+++ b/0_hello_world/main.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, "Hallo du schöne Welt!")
+}
--- a/0_hello_world/main_test.go
+++ b/0_hello_world/main_test.go
@@ -2,7 +2,7 @@ package main
 
 import (
 	"bytes"
-	"strings"
+	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -18,7 +18,7 @@ func TestMainOutput(t *testing.T) {
 	main()
 
 	// Then
-	expected := "Hallo du schöne Welt!\\n" // escape newline for easier string comparison
-	actual := strings.Replace(buf.String(), "\n", "\\n", -1)
+	expected := strconv.Quote("Hallo du schöne Welt!\n")
+	actual := strconv.Quote(buf.String())
 	r.Equalf(expected, actual, "Unexpected output in main()")
 }
--- a/0_hello_world/main_test.go
+++ b/0_hello_world/main_test.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMainOutput(t *testing.T) {
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := "Hallo du schöne Welt!\\n" // escape newline for easier string comparison
+	actual := strings.Replace(buf.String(), "\n", "\\n", -1)
+	r.Equalf(expected, actual, "Unexpected output in main()")
+}
--- a/README.md
+++ b/README.md
@@ -1,2 +1,31 @@
-# go-training
-Go programming language training
+# Go Training
+
+[![Build Status](https://travis-ci.com/anz-bank/go-training.svg?branch=master)](https://travis-ci.com/anz-bank/go-training)
+[![Coverage](https://codecov.io/gh/anz-bank/go-training/branch/master/graph/badge.svg)](https://codecov.io/gh/anz-bank/go-training)
+[![GolangCI](https://golangci.com/badges/github.com/anz-bank/go-training.svg)](https://golangci.com/r/github.com/anz-bank/go-training)
+
+This project is a playground for hands-on golang training.
+
+## Prerequisites
+
+-   Install `go 1.11` according to [official installation instruction](https://golang.org/doc/install)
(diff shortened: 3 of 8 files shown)


0_hello_world/main.go
```diff
@@ -0,0 +1,13 @@
+package main
```
anzboi, line 1: Interesting choice for the name of this file. What's the logic behind it?

0_hello_world/main.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, "Hallo du schöne Welt!")
```
anzboi, line 12: I don't like using a global `out` variable for testing purposes. Here's a thought I had. Replace `os.Stdout` for tests instead. Don't know whether this is a good idea or not, but it would look as follows.

Add this function to your test file.
```
func captureStdout(call func(), buf *bytes.Buffer) {
	realStdout := os.Stdout
	defer func() { os.Stdout = realStdout }()
	r, w, _ := os.Pipe()
	os.Stdout = w

	call()
	// os.Stdout = realStdout //EDIT: this line is redundant

	w.Close()
	io.Copy(buf, r)
}
```

Then the `//when` block in your test looks like
```
captureStdout(main, &buf)
```

README.md
```diff
@@ -1,2 +1,31 @@
...
+-   Clone this project outside your `$GOPATH` to enable [Go Modules](https://github.com/golang/go/wiki/Modules)
+-   Install `golangci-lint` according to [instructions](https://github.com/golangci/golangci-lint#local-installation)
+
+## Build, execute, test, lint
+
+Build and install this project with
+
+    go install ./...
+
+execute its binary with
```
anzboi: Capitalize execute -> Execute

README.md
```diff
@@ -1,2 +1,31 @@
...
+
+    0_hello_world
+
+test it with
```
anzboi: Capitalize

anzboi (approved):
LGTM. Don't know whether my idea for testing `stdout` is good or horrible, feel free to dismiss it.

0_hello_world/main.go
anzdaddy: So crazy, it might just work. The (very small) danger is other packages reading and storing os.Stdout on startup. The definitive approach is to build and run with stdin/stdout pipes.

camscale: I don't think this will work if the amount of output is larger than the pipe buffer size. The writer will block until the reader clears some space. The reading should be done in a goroutine concurrently with the call() func.

0_hello_world/main_test.go
```diff
@@ -0,0 +1,24 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := "Hallo du schöne Welt!\\n" // escape newline for easier string comparison
```
camscale: the meaning of this comment escapes me. How is it easier to compare a string with an escaped newline than without?

Is it just so that the error output is more readable when the test fails? If so, `strconv.Quote()` might be a better way to do that in general.

README.md
```diff
@@ -1,2 +1,31 @@
...
+
+## Build, execute, test, lint
+
+Build and install this project with
+
+    go install ./...
+
+execute its binary with
+
+    0_hello_world
```
camscale, line 23: I'm not sure where `go install ./...` installs a binary when using modules (no `GOPATH`), but if it is not on your `PATH`, then you will need some sort of path on this line, if only `./0_hello_world`. If it is on your `PATH`, you're assuming the user has their `PATH` set up the same way (`$GOPATH/bin` I assume).

camscale (approved):

0_hello_world/main.go
juliaogris: You need a package main with a function main ([specs](https://golang.org/ref/spec#Program_execution)) to create an executable program, so no choice whatsoever - am i missing something?

0_hello_world/main.go
juliaogris: @anzboi, @anzdaddy, @camh-anz  - I read that stackoverflow too, but it's kinda not [recommended practice](https://stackoverflow.com/a/10474375) @anzdaddy, I'm not really sure what you mean by the "definitive approach" - I'd love to know though.

Also, please remember this meant to be `go-training` - I don't want to overwhelm people with pipes etc.
All I want is 100% coverage. In the real world we wouldn't work with stdout in this way.

I do agree that `out` as a private global var is not ideal, how about I add a comment as a compromise? 
Or do you have any other good ideas on how to achieve 100% overage?

0_hello_world/main_test.go
juliaogris: Exactly @camh-anz:
```
        	Error:      	Not equal:
        	            	expected: "\"Hallo du schöne Welt!\""
        	            	actual  : "\"Hallo du schöne Welt!\\n\""
```
Shame on me for not searching for `strconv.Quote()` - kinda obvious that go would provide this. And this is exactly why I (also) :heart: codereviews.

README.md
juliaogris: Done, and the ones below as well assuming you meant for them to be Capitalized too.

README.md
juliaogris: Do you know where go installs a project's dependencies if the `GOAPTH` isn't set? With`GOPATH` they seem to end up in `$GOPATH/pkg/mod`.
I think people keep thinking that `go mod` doesn't require `GOPATH` at all any more, but I'm not sure that that's true, it just means _your_ development doesn't have to happen in `$GOPATH/src/github.com/your/repo`any longer .... I might be mistaken though.

camscale (approved):

PR #19, opened 2019-02-11
Lab 1 - Fibonacci

mohankrishna:
Fixes #24 

- Add 01_fib/main.go for printing first 7 digits of Fibonacci series extending to negatives as well
- Add 01_fib/main_test.go unit tests the implementation of Fibonacci series in 01_fib/main.go


01_fib/main.go
```diff
@@ -0,0 +1,34 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+
+	n1 := 0
+	n2 := 1
```
anzdaddy: You can fold this into one line: `n1, n2 := 0, 1`.

01_fib/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+	fibSeries := ""
+
+	for i := 0; i < n; i++ {
+		if n1 == 0 {
+			fibSeries = fmt.Sprintf("%s%d", fibSeries, n1)
```
anzdaddy: fibSeries is guaranteed to be `""` at this point. No need to include in the `fmt.Sprintf()` call.

01_fib/main.go
```diff
@@ -0,0 +1,34 @@
...
+		} else {
+			fibSeries = fmt.Sprintf("%d %s %d", n1*-1, fibSeries, n1)
```
anzdaddy: `-n1` suffices.

01_fib/main.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+		sum := n1 + n2
+		n1 = n2
+		n2 = sum
```
anzdaddy: You can update in one line: `n1, n2 = n2, sum`.

01_fib/main_test.go
```diff
@@ -0,0 +1,24 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := "-8 -5 -3 -2 -1 -1 0 1 1 2 3 5 8"
```
anzdaddy: Cute. 😄

01_fib/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+func fib(n int) {
+
+	n1 := 0
+	n2 := 1
+
+	fibSeries := ""
+
+	for i := 0; i < n; i++ {
+		if n1 == 0 {
```
anzdaddy: Better to initialise `fibSeries` to avoid the need for this edge case.

01_fib/main.go
mohankrishna: Updated

01_fib/main.go
mohankrishna: Good point and as you suggested in another comment, initialised the `fibSeries` to `"0"` and the loop now starts from `1` so that we don't need to handle the edge case in the loop.

01_fib/main.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+
```
juliaogris: Remove blank lines add beginning/end of block.
Could be further simplified to:
```go
	a, b := 1, 1
	for i := 1; i <= n; i++ {
		fmt.Fprintln(out, a)
		a, b = b, a+b
	}
```

01_fib/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+func TestFibOutputWithZero(t *testing.T) {
+
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	fib(0)
```
juliaogris: NICE!
Also test for -1, 1, and very large number.

juliaogris (commented):
First PR I read with properly filled in PR description and more than one test case - definite bonus points for this!

Please fix up  you first commit messages with `git rebase -i HASH` according to [recommendations](https://chris.beams.io/posts/git-commit/):
* Capitalize the subject line
* Use the imperative mood in the subject line

01_fib/main.go
mohankrishna: Updated the logic to print Negafibonacci numbers for negative `n`. And addressed your comments. Please review.

01_fib/main_test.go
mohankrishna: Added tests for -1 and 1. For very large number the tests timeout(`30s`), how do we test this?

01_fib/main.go
```diff
@@ -0,0 +1,41 @@
...
+		if n > 0 {
+			//Print Fibonacci numbers for positive n
+			fibSeries = fmt.Sprintf("%s%d\n", fibSeries, n1)
+		} else {
+			//Print Negafibonacci numbers for negative n
+			fibSeries = fmt.Sprintf("%v\n%s", math.Pow(-1, float64(i+1))*float64(n1), fibSeries)
+		}
+		n1, n2 = n2, n1+n2
+	}
+	fmt.Fprint(out, fibSeries)
```
anz-rfc: It's better to decouple fibonacci sequence calculation from IO. https://github.com/anz-bank/go-training/issues/37

01_fib/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := "1\n1\n2\n3\n5\n8\n13\n"
```
paraganz: You can make it readable like `1
1
2
3
5
8
13'

mohankrishna: Updated!

01_fib/main.go
mohankrishna: Done

01_fib/main_test.go
mohankrishna: As discussed we are leaving out the tests for very large numbers.

paraganz (dismissed):

01_fib/main.go
juliaogris: We haven't learned arrays yet, so will keep it for now and revisit later.

madhavansu: if you want to collect values, collect them in a slice and return it. Let main handle the printing

madhavansu (commented):
if you want to collect values, collect them in a slice and return it. Let main handle the printing

01_fib/main.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func abs(x int) int {
```
juliaogris: Use `math.Abs` ?

01_fib/main.go
mohankrishna: As the assignment suggested, we are to print the values inside Fib() method. Otherwise, it is just that we need to return the string and print it in main.

juliaogris (commented):
Pretty close to perfect, one little change or comment and we are there.
Also: nice commit messages and pr description!

01_fib/main.go
mohankrishna: `math.Abs` needs a `float64` thought this is simpler than converting the `int` to `float64` when absolute value of a int is required

juliaogris:
-   Click `Update branch` and locally `git pull origin Lab_1` before adding new commits.
-  Move files to folder `01_fib/mohankrishna`

mohankrishna:
@juliaogris All done! Please review.

juliaogris (approved):
LGTM

PR #31, opened 2019-02-11
-  implemented fibonacci series

osrnsree:
Fixes #

implemented fibonacci series

-
-

### DIFF
--- a/01_fib/main.go
+++ b/01_fib/main.go
@@ -9,15 +9,13 @@ import (
 var out io.Writer = os.Stdout
 
 func main() {
-	fib(5)
-
+	fib(7)
 }
 
-func fib(n int) int {
+func fib(n int) {
 	n1, n2 := 0, 1
 	for i := 0; i < n; i++ {
 		n1, n2 = n2, n1+n2
 		fmt.Fprint(out, " ", n1)
 	}
-	return n1
 }
--- a/01_fib/main.go
+++ b/01_fib/main.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(5)
+
+}
+
+func fib(n int) int {
+	var n1, n2, n3 = 0, 1, 0
+	for i := 0; i < n; i++ {
+		n1 = n2
+		n2 = n3
+		n3 = n1 + n2
+		fmt.Fprint(out, n3, " ")
+	}
+	return n3
+}
--- a/01_fib/main.go
+++ b/01_fib/main.go
@@ -14,12 +14,10 @@ func main() {
 }
 
 func fib(n int) int {
-	var n1, n2, n3 = 0, 1, 0
+	n1, n2 := 0, 1
 	for i := 0; i < n; i++ {
-		n1 = n2
-		n2 = n3
-		n3 = n1 + n2
-		fmt.Fprint(out, n3, " ")
+		n1, n2 = n2, n1+n2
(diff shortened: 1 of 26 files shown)


01_fib/main.go
```diff
@@ -0,0 +1,25 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(5)
+
+}
+
+func fib(n int) int {
+	var n1, n2, n3 = 0, 1, 0
```
anzdaddy: Use `:=` instead of `var` here.

01_fib/main.go
```diff
@@ -0,0 +1,25 @@
...
+	for i := 0; i < n; i++ {
+		n1 = n2
+		n2 = n3
```
anzdaddy: You can make this a one-liner: `n1, n2 = n2, n3`.

01_fib/main.go
golangcibot: ineffectual assignment to `n1` (from `ineffassign`)

01_fib/main.go
```diff
@@ -0,0 +1,25 @@
...
+		n3 = n1 + n2
```
anzdaddy: I think this should be the first assignment. Otherwise, `n2` becomes zero in the first pass. Also, you can do `n3 := n1 + n2` since n3 doesn't need to be tracked from one pass of the loop to the next. More so, you don't even need `n3` if you fold the sum into the one-liner suggested above.

01_fib/main.go
osrnsree: thanks for the suggestion, got the difference between := and var, 
in Go, := is for declaration + assignment, whereas = is for assignment only

osrnsree: removed ineffectual assignment

01_fib/main.go
osrnsree: changed the logic to one-liner which was quite efficient

01_fib/main.go
osrnsree: changed the logic to one-liner

juliaogris:
Please update PR description and fix broken build.

01_fib/main_test.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import (
+	"strings"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

01_fib/main.go
```diff
@@ -0,0 +1,23 @@
...
+
+}
+
+func fib(n int) int {
+	n1, n2 := 0, 1
+	for i := 0; i < n; i++ {
+		n1, n2 = n2, n1+n2
+		fmt.Fprint(out, " ", n1)
+	}
+	return n1
```
anz-rfc: It's better to decouple fibonacci sequence calculation from IO. https://github.com/anz-bank/go-training/issues/37

01_fib/main_test.go
osrnsree: able to do the format by updating the settings file as follows.
"editor.formatOnSave": true

juliaogris:
-   Click `Update branch` and locally `git pull origin lab2` before adding new commits.
-   Please fix up PR description according to [template](https://github.com/anz-bank/go-training/blob/master/.github/pull_request_template.md)
-   Please edit your commits to follow [recommendations](https://chris.beams.io/posts/git-commit/)
    -   Capitalize the subject line
    -   Use the imperative mood in the subject line
    -   Fold fixup commits into previous ones with `git rebase -i HASH`

juliaogris:
Hi,
Thanks!
Julia

Issue #63, opened 2019-02-12
Lab 3 - Letter frequency


eobjects:
- [ ] Create an executable go program in directory **03_letters**
- [ ] Write a function that returns a mapping of each letter to its frequency:
```go
func letters(s string) map[rune]int
```
- [ ] Write a function that returns a sorted slice of strings with elements "{key}:{val}". Use package [sort:](https://golang.org/pkg/sort/)
```go
func sortLetters(m map[rune]int) []string
```
Call **print(letters("aba"))** in **main** to print:
```go
a:2
b:1
```
- [ ] Bonus points: comprehensive tests

PR #68, opened 2019-02-13
Implement Lab 3 feature 'Letter Frequency Check'

jatheesh:
Fixes #63 

LETTER FREQUENCY CHECK FEATURE
-Add Feature Letter Frequency check and associated Test Cases

### DIFF
--- a/03_LabJat/letterFreequencyCount.go
+++ b/03_LabJat/letterFreequencyCount.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(input string) map[rune]int {
+	m := map[rune]int{}
+	for _, c := range input {
+		m[c]++
+	}
+	return m
+}
+func sortedLetters(m map[rune]int) []string {
+	var keys []string
+	for k, v := range m {
+		keys = append(keys, string(k)+":"+strconv.Itoa(v))
+	}
+	sort.Strings(keys)
+	return keys
+}
+func main() {
+	fmt.Fprintln(out, strings.Join(sortedLetters(letters("This is a String")), "\n"))
+}
--- a/03_LabJat/letterFreequencyCount.go
+++ b/03_LabJat/letterFreequencyCount.go
@@ -19,7 +19,7 @@ func letters(input string) map[rune]int {
 	return m
 }
 func sortedLetters(m map[rune]int) []string {
-	var keys []string
+	var keys = make([]string, 0)
 	for k, v := range m {
 		keys = append(keys, string(k)+":"+strconv.Itoa(v))
 	}
--- a/03_LabJat/letterFreequencyCount_test.go
+++ b/03_LabJat/letterFreequencyCount_test.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


03_LabJat/letterFreequencyCount.go
```diff
@@ -0,0 +1,31 @@
...
+
+func letters(input string) map[rune]int {
+	m := map[rune]int{}
+	for _, c := range input {
+		m[c]++
+	}
+	return m
+}
+func sortedLetters(m map[rune]int) []string {
+	var keys []string
```
golangcibot: Consider preallocating `keys` (from `prealloc`)

03_LabJat/letterFreequencyCount.go
```diff
@@ -0,0 +1,31 @@
...
+
+func letters(input string) map[rune]int {
+	m := map[rune]int{}
+	for _, c := range input {
+		m[c]++
+	}
+	return m
+}
+func sortedLetters(m map[rune]int) []string {
+	var keys = make([]string, 0)
```
camscale, line 22: this could be done a little more simply with: `keys := []string{}`

03_LabJat/letterFreequencyCount.go
```diff
@@ -0,0 +1,31 @@
...
+var out io.Writer = os.Stdout
+
+func letters(input string) map[rune]int {
+	m := map[rune]int{}
+	for _, c := range input {
+		m[c]++
+	}
+	return m
+}
+func sortedLetters(m map[rune]int) []string {
```
camscale, line 21: nit: put a blank line between the end of one function and the start of the next. Do this across the whole file.

03_LabJat/letterFreequencyCount_test.go
```diff
@@ -0,0 +1,60 @@
...
+	// Given
+	r := require.New(t)
+
+	// When
+	m := letters("aaa")
+
+	// Then
+	expected := 1
+	actual := len(m)
+	r.Equalf(expected, actual, "Unexpected output in main()")
```
camscale, line 34: the test description is wrong - this test is not testing main. I suggest using more specific descriptions, in this case, something like "unexpected number of elements in result".

This applies to the other test functions too.

03_LabJat/letterFreequencyCount_test.go
```diff
@@ -0,0 +1,60 @@
...
+	actual := m['a']
+	r.Equalf(expected, actual, "Unexpected output in main()")
+}
+func TestSortedLetters(t *testing.T) {
+	// Given
+	r := require.New(t)
+
+	// When
+	m := letters("abaa")
+	k := sortedLetters(m)
```
camscale, line 54: This test can fail if the `letters()` function breaks, because it is relying on it working properly. Ideally test cases are kept small and test one thing only. Since `sortedLetters()` takes a map as input, you should specify that map in this test case so you don't need to call `letters()`.

```
    k := sortedLetters(map[rune]int{'a': 3, 'b': 1})
```

camscale (commented):
Pretty close! Just a few tweaks and you'll be there.

camscale:
Sorry, two more comments I forgot.

Can you please reorganise your code so that it is in `03_letters/jatheesh/` rather than being in `03_LabJat`. We want to keep the structure uniform for all students.

You have some commits in this PR that add the `01_fib` files and then remove them in a later commit. It would be best if you could rebase to fix this (Julia can probably help if you can get some of her time). Otherwise just take care to avoid this in future labs.

juliaogris:
Hi,
Thanks!
Julia

Issue #67, opened 2019-02-13
Lab 4 - numeronym


mohankrishna:
- [ ] Create an executable go program in directory **04_numeronym/USERNAME**
- [ ] Write a function that returns a slice of numeronyms for its input strings:
```go
func numeronyms(vals ...string) []string
```
- [ ] Call **fmt.Println(numeronyms("accessibility", "Kubernetes", "abc"))** in main to print:
```go
[a11y K8s abc]
```

PR #81, opened 2019-02-13
Lab 04 numeronym

sahidakhtar1:
Fixes #67

- `numeronyms ()` to accepts slice of string and return slice with numeronyms of each string. It's a variadic functions.
- Unit test to test `numeronyms ()`


04_numeronym/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,67 @@
...
+	testData := []struct {
+		in  []string
+		out []string
+	}{
+		{[]string{"accessibility", "Kubernetes", "abc"}, []string{"a11y", "K8s", "abc"}},
+		{[]string{"accessibility", "Kubernetes", "abc", "accessibility"}, []string{"a11y", "K8s", "abc", "a11y"}},
+		{[]string{"123456"}, []string{"146"}},
+		{[]string{"-123456"}, []string{"-56"}},
+		{[]string{"aaaaaaaaaaaaaaaa"}, []string{"a14a"}},
+		{[]string{"  accessibility  ", "Kubernetes  ", "  abc"}, []string{"a11y", "K8s", "abc"}},
```
paraganz, line 34: add test case for unicode characters

sahidakhtar1: Done.

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,49 @@
...
+		result[i] = createNumeronym(v)
+	}
+	return result
+}
+
+func createNumeronym(s string) string {
+	trimedString := strings.TrimSpace(s)
+	var numeronym string
+	//If the string contains space or \t or \n then it will return the same string
+	if strings.Contains(trimedString, " ") ||
```
juliaogris: https://golang.org/pkg/strings/#ContainsAny is probably the cleaner choice.

It has been specified clearly so I guess this fine, but I think the nicer solution is to work with the `Trim`ed string and treat internal space as any other character.

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,49 @@
...
+		strings.Contains(trimedString, "\t") ||
+		strings.Contains(trimedString, "\n") {
+		return s
+	}
+	var count int
+	var s1, s2 string
```
juliaogris: Better convert to `runes := []rune(trimedString)` and `runes[0]`, `runes[len(runes) - 1]`

juliaogris (changes requested):
A couple of minor things.

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: Done.

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: Done

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,42 @@
...
+	if strings.ContainsAny(trimedString, " & \t &\n") {
+		return s
+	}
+	runes := []rune(trimedString)
+	count := len(runes)
+	if count <= 3 {
+		numeronym = trimedString
+	} else {
+		s1 := runes[0]
+		s2 := runes[len(runes)-1]
```
madhavansu: This can be clubbed into single liner like 
`s1, s2 := runes[0], runes[len(runes)-1]`

04_numeronym/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,72 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
```
madhavansu, line 6: Remove the blank line

sahidakhtar1: @madhavansu it gets auto formatted on save. Also no linter error on this. So we should skip this.

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: Done

04_numeronym/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+		{"abc  ", "abc"},
+		{"abc efhjb", "abc efhjb"},
+		{"abc\\tefhjb", "a8b"},
+		{"abc\nefhjb", "abc\nefhjb"},
+		{"abcefhjb\n", "a6b"},
+		{"世🖖界", "世🖖界"},
+		{"世", "世"},
+		{"a世🖖界bc", "a4c"},
+		{"世abc", "世2c"},
+		{"abc世", "a2世"},
```
paraganz, line 65: Please check if this works for `⌘abcef♥️`

sahidakhtar1: Added and works fine.

04_numeronym/sahidakhtar1/main_test.go
anzboi: Blank lines between native imports and third party imports is part of the golang conventions and `gofmt` formatting rules.

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,40 @@
...
+		result[i] = createNumeronym(v)
+	}
+	return result
+}
+
+func createNumeronym(s string) string {
+	trimedString := strings.TrimSpace(s)
+	var numeronym string
+	//If the string contains space or \t or \n then it will return the same string
+	if strings.ContainsAny(trimedString, " & \t &\n") {
```
anzboi: If your intent is to only check for space, `\t` and `\n`, then this is incorrect, it will also match `&`.
Use `" \t\n"` instead.

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,40 @@
...
+func numeronyms(vals ...string) []string {
+	result := make([]string, len(vals))
+	for i, v := range vals {
+		result[i] = createNumeronym(v)
+	}
+	return result
+}
+
+func createNumeronym(s string) string {
+	trimedString := strings.TrimSpace(s)
```
anzboi: You can trim more using `strings.Trim(s, <charsToTrim>)`.

04_numeronym/sahidakhtar1/main.go
```diff
@@ -0,0 +1,40 @@
...
+	//If the string contains space or \t or \n then it will return the same string
+	if strings.ContainsAny(trimedString, " & \t &\n") {
+		return s
+	}
+	runes := []rune(trimedString)
+	count := len(runes)
+	if count <= 3 {
+		numeronym = trimedString
+	} else {
+		numeronym = string(runes[0]) + strconv.Itoa(count-2) + string(runes[len(runes)-1])
```
anzboi: The last term here: `string(runes[len(runes)-1])`, you have already stored `len(runes)` as `count` so you can write `string(runes[count-1])`.
Not a big issue but makes it a little more readable.

04_numeronym/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+		{"⌘abcef♥", "⌘5♥"},
+	}
```
anzboi, line 67: Very comprehensive tests 👍

anzboi (commented):
Very close, just a few things to consider.

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: The intent here is if the string contains more than one word then do nothing and return the same string.  So not using `Trim()`

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: Done

04_numeronym/sahidakhtar1/main.go
sahidakhtar1: Done

04_numeronym/sahidakhtar1/main.go
juliaogris: [`strings.Trim`](https://golang.org/pkg/strings/#Trim) takes as second parameter a `cutset`:
> Trim returns a slice of the string s with all leading and trailing Unicode code points contained in cutset removed.

So what @anzboi might be suggesting is to use it as `strings.Trim(s, "\n\t ")`

juliaogris (dismissed):
LGTM 👍 

There might have been some misunderstanding around `strings.Trim`:

[`strings.Trim`](https://golang.org/pkg/strings/#Trim) takes as second parameter a `cutset`:
> Trim returns a slice of the string s with all leading and trailing Unicode code points contained in cutset removed.

So what @anzboi might be suggesting is to use it as `strings.Trim(s, "\n\t ")`

juliaogris (approved):
LGTM

Issue #70, opened 2019-02-13
Lab 5 - Stringer


mohankrishna:
- [ ] Create an executable go program in directory 05_stringer/USERNAME

- [ ] Make the IPAddr type implement fmt.Stringer to print the address as a dotted quad
- [ ] Find hints at [Tour of Go Exercise: Stringers](https://tour.golang.org/methods/18)
- [ ] Call **fmt.Println(IPAddr{127, 0, 0, 1})** in **main** to print:
```
127.0.0.1
```

PR #84, opened 2019-02-13
Lab 4 and 5: Numeronym and Stringers implementation

madhavansu:
Fixes #67, fixes #70  

- Write a function that returns a slice of numeronyms for its input strings

### DIFF
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 )
 
 var out io.Writer = os.Stdout
@@ -14,12 +15,12 @@ func main() {
 func numeronyms(vals ...string) []string {
 	var strcopy []string
 	for _, val := range vals {
+		val := []rune(strings.TrimSpace(val))
 		length := len(val)
 		if length > 3 {
-			first, last := val[0:1], val[length-1:]
-			strcopy = append(strcopy, fmt.Sprint(first, length-2, last))
+			strcopy = append(strcopy, fmt.Sprint(string(val[0:1]), length-2, string(val[length-1:])))
 		} else {
-			strcopy = append(strcopy, val)
+			strcopy = append(strcopy, string(val))
 		}
 	}
 	return strcopy
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
+	var strcopy []string
+	for _, val := range vals {
+		length := len(val)
+		if length > 3 {
+			first, last := val[0:1], val[length-1:]
+			strcopy = append(strcopy, fmt.Sprint(first, length-2, last))
+		} else {
+			strcopy = append(strcopy, val)
+		}
+
+	}
+	return strcopy
+}
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -13,14 +13,14 @@ func main() {
 	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
 }
 func numeronyms(vals ...string) []string {
-	var strcopy []string
-	for _, val := range vals {
+	strcopy := make([]string, len(vals))
+	for k, val := range vals {
 		val := []rune(strings.TrimSpace(val))
 		length := len(val)
 		if length > 3 {
-			strcopy = append(strcopy, fmt.Sprint(string(val[0:1]), length-2, string(val[length-1:])))
+			strcopy[k] = fmt.Sprint(string(val[0]), length-2, string(val[length-1]))
 		} else {
-			strcopy = append(strcopy, string(val))
+			strcopy[k] = string(val)
 		}
 	}
 	return strcopy
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -13,15 +13,15 @@ func main() {
 	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
 }
 func numeronyms(vals ...string) []string {
-	strcopy := make([]string, len(vals))
+	s := make([]string, len(vals))
 	for k, val := range vals {
-		val := []rune(strings.TrimSpace(val))
+		val := []rune(strings.Trim(val, "\n\t\\ "))
 		length := len(val)
 		if length > 3 {
-			strcopy[k] = fmt.Sprint(string(val[0]), length-2, string(val[length-1]))
+			s[k] = fmt.Sprint(string(val[0]), length-2, string(val[length-1]))
 		} else {
-			strcopy[k] = string(val)
+			s[k] = string(val)
 		}
 	}
-	return strcopy
+	return s
 }
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -11,7 +11,6 @@ var out io.Writer = os.Stdout
 func main() {
 	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
 }
-
 func numeronyms(vals ...string) []string {
 	var strcopy []string
 	for _, val := range vals {
@@ -22,7 +21,6 @@ func numeronyms(vals ...string) []string {
 		} else {
 			strcopy = append(strcopy, val)
 		}
-
 	}
 	return strcopy
 }
--- a/04_numeronym/madhavansu/main.go
+++ b/04_numeronym/madhavansu/main.go
@@ -12,13 +12,14 @@ var out io.Writer = os.Stdout
 func main() {
 	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
 }
+
 func numeronyms(vals ...string) []string {
 	s := make([]string, len(vals))
 	for k, val := range vals {
 		val := []rune(strings.Trim(val, "\n\t\\ "))
 		length := len(val)
 		if length > 3 {
-			s[k] = fmt.Sprint(string(val[0]), length-2, string(val[length-1]))
+			s[k] = fmt.Sprintf("%c%d%c", val[0], length-2, val[length-1])
 		} else {
 			s[k] = string(val)
 		}
--- a/04_numeronym/madhavansu/main_test.go
+++ b/04_numeronym/madhavansu/main_test.go
@@ -33,7 +33,6 @@ func TestMainOutput(t *testing.T) {
 	actual := strconv.Quote(buf.String())
 	r.Equalf(expected, actual, "Unexpected output in main()")
 }
-
 func TestNumeronyms(t *testing.T) {
 	// Given
 	r := require.New(t)
--- a/04_numeronym/madhavansu/main_test.go
+++ b/04_numeronym/madhavansu/main_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"bytes"
 	"fmt"
-	"strconv"
 	"testing"
 
 	"github.com/stretchr/testify/require"
@@ -13,10 +12,11 @@ var s = []struct {
 	input  string
 	output string
 }{
-	{"accessibility", "[a11y]"},
+	{"accessibility  ", "[a11y]"},
 	{"Kubernetes", "[K8s]"},
 	{"World Wide Web", "[W12b]"},
-	{"abc", "[abc]"},
+	{"மாதவ", "[ம2வ]"},
+	{"ΩΟΞψωξ", "[Ω4ξ]"},
 }
 
 func TestMainOutput(t *testing.T) {
@@ -29,8 +29,8 @@ func TestMainOutput(t *testing.T) {
 	main()
 
 	// Then
-	expected := strconv.Quote(`[a11y K8s abc]`)
-	actual := strconv.Quote(buf.String())
+	expected := `[a11y K8s abc]`
+	actual := buf.String()
 	r.Equalf(expected, actual, "Unexpected output in main()")
 }
 func TestNumeronyms(t *testing.T) {
--- a/04_numeronym/madhavansu/main_test.go
+++ b/04_numeronym/madhavansu/main_test.go
@@ -2,11 +2,22 @@ package main
 
 import (
 	"bytes"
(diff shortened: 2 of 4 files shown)


04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,26 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+func numeronyms(vals ...string) []string {
+	var strcopy []string
```
mohankrishna: Since we know the length of the input array, it is a good practice to pre-allocate the size of array `strcopy`.

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,26 @@
...
+	for _, val := range vals {
+		length := len(val)
+		if length > 3 {
+			first, last := val[0:1], val[length-1:]
```
mohankrishna: you can directly access first and last values with `val[0]` and `val[length-1]`

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+func numeronyms(vals ...string) []string {
+	var strcopy []string
+	for _, val := range vals {
```
mohankrishna: After preallocating the size you can use the `position` that comes from `range` as 
```go
for pos, val := range vals
```
and assign the values as 
```go
strcopy[pos] = fmt.Sprint(first, length-2, last)
```

04_numeronym/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote(`[a11y K8s abc]`)
```
mohankrishna: you probably dont need to wrap both `expected` and `actual` in `strconv.Quote`. Compare them directly.

04_numeronym/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var s = []struct {
+	input  string
+	output string
+}{
+	{"accessibility", "[a11y]"},
+	{"Kubernetes", "[K8s]"},
+	{"World Wide Web", "[W12b]"},
+	{"abc", "[abc]"},
```
mohankrishna: Please add tests for unicode characters

05_stringer/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote(`127.0.0.1`)
```
mohankrishna: Same as above `strconv.Quote` is not required

05_stringer/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	r.Equalf(expected, actual, "Unexpected output in main()")
+}
+
+var s = []struct {
+	input  IPAddr
+	output string
+}{
+	{IPAddr{127, 0, 0, 1}, "127.0.0.1"},
+	{IPAddr{10, 10, 40, 1}, "10.10.40.1"},
+	{IPAddr{7, 20, 210, 12}, "7.20.210.12"},
```
mohankrishna: Add some more tests use something like
```go
IPAddr{127}
IPAddr{127,1}
IPAddr{127,3,4}
```

mohankrishna (dismissed):

04_numeronym/madhavansu/main.go
madhavansu: The functional implementation uses the append string, when you use append and also pre-allocate the size, it will start appending at the end of the allocated size and you will see an unnecessary space preponed to the string.

04_numeronym/madhavansu/main.go
madhavansu: This is another approach to access the character and I think we are more used to do with `val[0]` so tried something different.

04_numeronym/madhavansu/main.go
madhavansu: I have followed different approach.

04_numeronym/madhavansu/main_test.go
madhavansu: This is for the readability purpose, as a user we will know what is expected and what is the actual value.

05_stringer/madhavansu/main_test.go
madhavansu: This is for the readability purpose, as a user we will know what is expected and what is the actual value.

04_numeronym/madhavansu/main_test.go
madhavansu: Updated.

05_stringer/madhavansu/main_test.go
madhavansu: Updated

04_numeronym/madhavansu/main.go
mohankrishna: Exactly, if we pre-allocate we don't need to append we can just assign the value at the position. As suggested in another comment.

mohankrishna: Or, you can preallocate `strcopy` with a size of 0 and capacity of len(vals) 
```go
strcopy := make([]string, 0, len(vals))
```
and continue appending values to it.

My thought here is when we don't init a slice with the required capacity and keep appending values to the slice everytime the size of slice exceeds its capacity it will have to 
a) reallocate a new slice with higher capacity b)copy its own values and c) perform append.
Maybe @juliaogris can confirm on this.

However for a smaller data set like this it shouldn't matter much :smiley:

madhavansu: Updated

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+func numeronyms(vals ...string) []string {
+	strcopy := make([]string, len(vals))
```
juliaogris: It's not really a copy. Use `s` for slice or `result` instead.

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,27 @@
...
+	for k, val := range vals {
+		val := []rune(strings.TrimSpace(val))
```
juliaogris: Maybe use `strings.Trim(val, "\n\t\ ")`.

04_numeronym/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+var s = []struct {
+	input  string
+	output string
+}{
+	{"accessibility  ", "[a11y]"},
+	{"Kubernetes", "[K8s]"},
+	{"World Wide Web", "[W12b]"},
+	{"மாதவ", "[ம2வ]"},
+	{"ΩΟΞψωξ", "[Ω4ξ]"},
```
juliaogris, line 19: Nice!
Add no string input, more than one string input, empty strings, white space strings.
```go
var s = []struct {
	input  []string
	expected []string
}{ /*  ...   */}
// Then down below:

r.Equal(v.expected, numeronyms(v.input...))

```

04_numeronym/madhavansu/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+}
+func TestNumeronyms(t *testing.T) {
+	// Given
+	r := require.New(t)
+
+	// When
+	for _, v := range s {
+		expected := fmt.Sprint(numeronyms(v.input))
+		actual := v.output
+		r.Equalf(expected, actual, "Unexpected output in numeronyms()")
```
juliaogris: Swap expected with actual: actual is the actual, calculated output. Expected is what you expect it to be.
The following should do:
```go
r.Equal(v.expected, numeronyms(v.input...))
```

05_stringer/madhavansu/main.go
```diff
@@ -0,0 +1,20 @@
...
+var out io.Writer = os.Stdout
+
+// IPAddr export
+type IPAddr [4]byte
+
+func main() {
+	fmt.Fprint(out, IPAddr{127, 0, 0, 1})
+}
+
+func (ipAddr IPAddr) String() string {
```
juliaogris: Use single or two character receiver name (see [recommendation](https://github.com/golang/go/wiki/CodeReviewComments#variable-names))

05_stringer/madhavansu/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	input  IPAddr
+	output string
+}{
+
+	{IPAddr{127, 0, 0, 1}, "127.0.0.1"},
+	{IPAddr{10, 10, 40, 1}, "10.10.40.1"},
+	{IPAddr{7, 20, 210}, "7.20.210.0"},
+	{IPAddr{7, 20}, "7.20.0.0"},
+	{IPAddr{0}, "0.0.0.0"},
+	{IPAddr{}, "0.0.0.0"},
```
juliaogris, line 34: Nice!

05_stringer/madhavansu/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Then
+	expected := `127.0.0.1`
+	actual := buf.String()
+	r.Equalf(expected, actual, "Unexpected output in main()")
+}
+
+var s = []struct {
+	input  IPAddr
+	output string
```
juliaogris: I'd rename this to `expected` or at least `expectedOutput`

04_numeronym/madhavansu/main.go
madhavansu: Updated.

04_numeronym/madhavansu/main.go
madhavansu: Updated.

04_numeronym/madhavansu/main_test.go
madhavansu: Updated.

04_numeronym/madhavansu/main_test.go
madhavansu: Updated.

05_stringer/madhavansu/main.go
madhavansu: Updated.

05_stringer/madhavansu/main_test.go
madhavansu: Updated.

05_stringer/madhavansu/main_test.go
madhavansu: Thank you.

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,27 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+func numeronyms(vals ...string) []string {
+	s := make([]string, len(vals))
+	for k, val := range vals {
+		val := []rune(strings.Trim(val, "\n\t\\ "))
```
anzboi, line 19: Nice 👍

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"io"
+	"os"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
```
anzboi, line 14: Add a newline after closing `}`

04_numeronym/madhavansu/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	r.Equalf(`[a11y K8s abc]`, buf.String(), "Unexpected output in main()")
+}
```
anzboi, line 39: Add newline after closing `}`

05_stringer/madhavansu/main.go
```diff
@@ -0,0 +1,20 @@
...
+
+// IPAddr export
+type IPAddr [4]byte
+
+func main() {
+	fmt.Fprint(out, IPAddr{127, 0, 0, 1})
+}
+
+func (i IPAddr) String() string {
+	return fmt.Sprintf("%d.%d.%d.%d", i[0], i[1], i[2], i[3])
```
anzboi, line 19: Perfect 👍

04_numeronym/madhavansu/main.go
```diff
@@ -0,0 +1,27 @@
...
+func numeronyms(vals ...string) []string {
+	s := make([]string, len(vals))
+	for k, val := range vals {
+		val := []rune(strings.Trim(val, "\n\t\\ "))
+		length := len(val)
+		if length > 3 {
+			s[k] = fmt.Sprint(string(val[0]), length-2, string(val[length-1]))
```
anzboi: This works fine. An interesting alternative I thought of
`s[k] = fmt.Sprintf("%c%d%c", val[0], length-1, val[length-1])`
Advantage: removes casts
Disadvantage: runtime cost of string formatting

anzboi (commented):
Almost perfect, just a few formatting issues.

04_numeronym/madhavansu/main.go
anzboi: I see this has been changed. Always prefer single access over slice access when just accessing one element at a known index.

04_numeronym/madhavansu/main_test.go
anzboi: This is really a personal preference thing. If they are complicated to construct we usually prefer `expected` and `actual`, otherwise its really up to the programmer.

04_numeronym/madhavansu/main.go
madhavansu: Updated.

04_numeronym/madhavansu/main_test.go
madhavansu: Updated.

04_numeronym/madhavansu/main.go
madhavansu: Updated.

anzboi (approved):
LGTM

PR #94, opened 2019-02-13
05 stringer

sahidakhtar1:
Fixes #70 

* Created type `IPAddr` to capture all 4 components of an IP address .
* Implemented `Stringer` interface to print the IP address in dotted quad.

### DIFF
--- a/05_stringer/sahidakhtar1/main.go
+++ b/05_stringer/sahidakhtar1/main.go
@@ -8,9 +8,7 @@ import (
 
 var out io.Writer = os.Stdout
 
-type MyByte byte
-
-type IPAddr [4]MyByte
+type IPAddr [4]byte
 
 func (i IPAddr) String() string {
 	return fmt.Sprintf("%d.%d.%d.%d", i[0], i[1], i[2], i[3])
--- a/05_stringer/sahidakhtar1/main.go
+++ b/05_stringer/sahidakhtar1/main.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type IPAddr [4]byte
+
+func (i IPAddr) String() string {
+	return fmt.Sprintf("%v.%v.%v.%v", i[0], i[1], i[2], i[3])
+}
+
+func main() {
+	fmt.Fprintln(out, IPAddr{127, 0, 0, 1})
+}
--- a/05_stringer/sahidakhtar1/main.go
+++ b/05_stringer/sahidakhtar1/main.go
@@ -8,10 +8,12 @@ import (
 
 var out io.Writer = os.Stdout
 
-type IPAddr [4]byte
+type MyByte byte
+
+type IPAddr [4]MyByte
 
 func (i IPAddr) String() string {
-	return fmt.Sprintf("%v.%v.%v.%v", i[0], i[1], i[2], i[3])
+	return fmt.Sprintf("%d.%d.%d.%d", i[0], i[1], i[2], i[3])
 }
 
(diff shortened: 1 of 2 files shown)


05_stringer/sahidakhtar1/main.go
```diff
@@ -0,0 +1,19 @@
...
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type IPAddr [4]byte
+
+func (i IPAddr) String() string {
+	return fmt.Sprintf("%v.%v.%v.%v", i[0], i[1], i[2], i[3])
```
anzboi: Since you know these are type `byte`, you should use `%d` here, not `%v`. May not seem like an issue but can cause issues later down the track.
If you add a type `type MyByte byte` and change `IPAddr` to `type IPAddr [4]MyByte`, the rest of the code will still work as intended. Someone can then add their own stringer method `func (m MyByte) String() string` and it will change the behaviour of this stringer method. Try out the following code
```
type MyInt int

// See the difference with and without this commented
// func (i MyInt) String() string {
// 	return "BAD"
// }

type MyMultiInt [4]MyInt

func (m MyMultiInt) String() string {
	return fmt.Sprintf("%d.%d.%v.%v", m[0], m[1], m[2], m[3])
}

func main() {
	var m = MyMultiInt{1, MyInt(1), 1, MyInt(1)}
	fmt.Println(m)
}
```

05_stringer/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+	{IPAddr{255, 255, 255, 255}, "255.255.255.255"},
+	{IPAddr{1}, "1.0.0.0"},
+}
+
+func TestIPAddrFmtInterfaceImpl(t *testing.T) {
+	r := require.New(t)
+	for _, tt := range testData {
+		r.Equalf(tt.out, tt.in.String(), "Convert IP address to dotted quad fails")
+	}
+}
```
anzboi, line 40: Great tests 👍

anzboi (commented):
Very close, just need to address the usage of `%v` in string formatting.

05_stringer/sahidakhtar1/main.go
sahidakhtar1: Thanks for the nice suggestion. Changes done accordingly.

05_stringer/sahidakhtar1/main.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type MyByte byte
```
anzboi: I probably should have been more clear in my previous comment. The issue is if you add a type like this with it's own stringer, it can mess up the behaviour of your own stringer implementation using `%v`. Using `%d` protects you from that in the future.
You don't need `MyByte`, the code I gave you is just meant to illustrate the danger of using `%v` for what is expected to be literals.

anzboi (commented):
Sorry to bother you with one more change as a result of my previous comment.
Otherwise LGTM 👍

05_stringer/sahidakhtar1/main.go
sahidakhtar1: Got it. If define type `MyByte` with its own stringer and use %v then it will call `MyByte`'s stringer instead of `IPAddr` stringer. Right?

anzboi: Almost, `%v` will look at the `MyByte` stringer, whereas `%d` only cares about the underlying `byte`.

anzboi (approved):
LGTM

PR #99, opened 2019-02-14

### LINKED ISSUE #88: Lab 6 - CRUD puppy with interface

crud_puppy

mohankrishna:
Fixes #88 

- Write MapStore implementation backed by the map
- Write SyncStore implementation backed by the sync.map
- Write tests against the Storer interface and run in a suite with both implementations


06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+
+type MapStore struct {
+	m map[uint32]Puppy
+}
+
+func CreateStore() *MapStore {
+	return &MapStore{make(map[uint32]Puppy)}
+}
+
+func (ms *MapStore) CreatePuppy(puppy *Puppy) {
```
juliaogris: Puppy is small, you can work without pointers as it's being copied on map insertion (better double check that I'm saying the right thing here ;) )

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+	ReadPuppy(ID uint32) *Puppy
+	UpdatePuppy(ID uint32, pet *Puppy)
+	DeletePuppy(ID uint32) bool
+}
+
+type MapStore struct {
+	m map[uint32]Puppy
+}
+
+func CreateStore() *MapStore {
```
juliaogris: Go convention for constructors is to use  the `New` prefix, so I'd call it `NewMapStore`.

camscale: Idiomatic Go calls for naming constructors as `NewSomething()` (when not `something.New()`). You can see this across the Go standard library, and is also alluded to here: https://golang.org/doc/effective_go.html#composite_literals

So this should be called `NewMapStore()`.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+	return &MapStore{make(map[uint32]Puppy)}
+}
+
+func (ms *MapStore) CreatePuppy(puppy *Puppy) {
+	ms.m[puppy.ID] = *puppy
```
juliaogris: Also, typically Create is responsible for assigning new IDs - user of this package will probably no know which IDs are still free.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+	Breed  string
+	Colour string
+	Value  int
+}
+
+type Storer interface {
+	CreatePuppy(puppy *Puppy)
+	ReadPuppy(ID uint32) *Puppy
+	UpdatePuppy(ID uint32, pet *Puppy)
+	DeletePuppy(ID uint32) bool
```
juliaogris: I think you can work completely without pointers here for simplicity.

06_puppy/mohankrishna/storage_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// when
+	actual := *s.store.ReadPuppy(0x1234)
+
+	// then
+	expected := puppy1()
+	assert.Equal(expected, actual)
+}
+
+func (s *storerSuite) TestMapStoreReadPuppy_IDDoesNotExist() {
```
juliaogris: No underscores in golang variable and function names.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+func CreateStore() *MapStore {
+	return &MapStore{make(map[uint32]Puppy)}
+}
+
+func (ms *MapStore) CreatePuppy(puppy *Puppy) {
+	ms.m[puppy.ID] = *puppy
+}
+
+func (ms *MapStore) ReadPuppy(id uint32) *Puppy {
+	puppy := ms.m[id]
```
camscale: This will cause a zero value puppy to be returned if the `id` is not in the map. I'd suggest a better way to deal with that is to check if it exists and if not, return `nil`. In future when you get to errors, you may add an additional return value for the error, and in that case, you'll also be returning `nil` for the `*Puppy`, along with a non-nil error.

06_puppy/mohankrishna/storage_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// when
+	s.store.CreatePuppy(&newPup)
+	createdPup := *s.store.ReadPuppy(0x1236)
+	createdPup.Value = 12323
+	s.store.UpdatePuppy(createdPup.ID, &createdPup)
+
+	// then
+	updatedPup := *s.store.ReadPuppy(createdPup.ID)
+	assert.Equal(createdPup, updatedPup, "Updated puppy should be identical to the one passed to Update")
```
juliaogris: Possibly except for the ID.

juliaogris (changes requested):
Looks pretty solid, you are my first review for this lab so might have more feedback later...

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (ms *MapStore) UpdatePuppy(id uint32, puppy *Puppy) {
+	ms.m[id] = *puppy
+}
+
+func (ms *MapStore) DeletePuppy(id uint32) bool {
+	_, ok := ms.m[id]
+	if ok {
+		delete(ms.m, id)
```
camscale, line 40: There is a race condition here where another thread/goroutine could delete the puppy in the mean time, so your delete will "fail". The end result is that both calls to `DeletePuppy()` would return true, when only one actually deleted the value.

With the tools you have so far, there's not much you can do about this, but it is worth at least adding a comment that this race condition exists so it doesn't go unnoticed if you extend this program later.

06_puppy/mohankrishna/map_store.go
camscale: This function accepts a `*Puppy` but stores a copy of it. I think normal expectations would be that if the interface accepts a pointer to a struct on insertion and returns a pointer on lookup, then the pointers will be the same and no copies are being made. This is not a universal assumption though (and in the case of an external data store behind the interface, it would not be possible).

There are two ways you could address this:
1) Change the interface to work with values (`Puppy`) instead of pointers (`*Puppy`)
2) Continue to do what you're doing, but add a godoc comment for the function that describes that it takes a copy and when looked up, the returned value will not be the same struct as passed in originally.

Since in another comment I suggest returning `nil` when the `ID` does not exist in the data store, option 2 would make the most sense (even though @juliaogris is suggesting you use option 1). But option 1 means you can ignore my other comment :) so that might be easier.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+	ms.m[puppy.ID] = *puppy
+}
+
+func (ms *MapStore) ReadPuppy(id uint32) *Puppy {
+	puppy := ms.m[id]
+	return &puppy
+}
+
+func (ms *MapStore) UpdatePuppy(id uint32, puppy *Puppy) {
+	ms.m[id] = *puppy
```
camscale: This has the potential for getting IDs out of sync. What happens if `id` is different to `puppy.ID`? Should you do something special here, like changing `puppy.ID` to match `id`? Or something else? If you do, a godoc comment would then be a good place to explain that.

06_puppy/mohankrishna/sync_store.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"sync"
+)
+
+type SyncStore struct {
+	sm sync.Map
+}
+
+func CreateSyncStore() *SyncStore {
```
camscale: Idiomatic naming for this would be `NewSyncStore()`.

However it would also be perfectly acceptable to remove this function altogether, since you are just returning the zero value for the type. In those cases, the caller can just use `&SyncStore{}` when they want a new SyncStore. It is common in Go to make your types work well/properly as a zero value, which reduces the amount of code you need to write.

06_puppy/mohankrishna/sync_store.go
```diff
@@ -0,0 +1,40 @@
...
+
+func (ss *SyncStore) ReadPuppy(id uint32) *Puppy {
+	v, ok := ss.sm.Load(id)
+	if ok {
+		if puppy, ok := v.(Puppy); ok {
+			return &puppy
+		}
+	}
+	//This should return an error, we will be implementing the errors in next lab
+	return &Puppy{}
```
camscale: It would make more sense to return `nil` here, so the caller can easily tell if the read failed. When you get to errors, you'll want to return `nil` here anyway, as that is the zero value for pointers.

06_puppy/mohankrishna/sync_store.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (ss *SyncStore) UpdatePuppy(id uint32, puppy *Puppy) {
+	ss.sm.Store(id, *puppy)
```
camscale: Same comment here as I made on `MemStore.UpdatePuppy()` - what happens if the IDs don't match.

06_puppy/mohankrishna/sync_store.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func (ss *SyncStore) DeletePuppy(id uint32) bool {
+	_, ok := ss.sm.Load(id)
+	if ok {
+		ss.sm.Delete(id)
```
camscale, line 46: Same comment here as I made on `MemStore.DeletePuppy()` - a race condition means you may return `true` when you didn't actually delete the puppy.

06_puppy/mohankrishna/storage_test.go
```diff
@@ -0,0 +1,154 @@
...
+	case maps:
+		// create a map store
+		s.store = CreateStore()
+		puppy := puppy1()
+		s.store.CreatePuppy(&puppy)
+	case smaps:
+		// create a sync map store
+		s.store = CreateSyncStore()
+		puppy := puppy1()
+		s.store.CreatePuppy(&puppy)
```
camscale: The body of the `case` should probably just create storer object, and you can move `puppy := puppy1(); s.store.CreatePuppy(&puppy)` to outside the `switch`. This makes it obvious to the reader that the different stores are still set up the same way with the same data.

06_puppy/mohankrishna/storage_test.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (s *storerSuite) TestMapStoreReadPuppy_IDDoesNotExist() {
+	// given
+	assert := tassert.New(s.T())
+
+	// when
+	actual := *s.store.ReadPuppy(0x123)
+
+	// then
+	assert.Equal(Puppy{0, "", "", 0}, actual)
```
camscale: `Puppy{0, "", "", 0}` is the zero value for `Puppy`. You should be able to use just `Puppy{}` here - it is more explicit to the reader that is is the zero value.

06_puppy/mohankrishna/map_store.go
camscale: What happens if `puppy.ID` is already in the store? Is it right to erase another puppy from existence just because another wants its ID? Poor puppy. :)

This could be alleviated by generating the IDs yourself in the store, rather than requiring the caller to know what IDs to use.

06_puppy/mohankrishna/storage_test.go
```diff
@@ -0,0 +1,154 @@
...
+	//given
+	assert := tassert.New(s.T())
+
+	// when
+	deleted := s.store.DeletePuppy(0x1234)
+
+	// then
+	assert.True(deleted, "Delete should return true indicating a puppy was deleted")
+	actual := *s.store.ReadPuppy(0x1234)
+	assert.Equal(Puppy{0, "", "", 0}, actual)
```
camscale: Use `Puppy{}` for the zero value.

camscale (commented):
This is pretty good. Many of my comments are asking questions about how to handle cases that will make the code more complicated. It's fine not to do that if you don't feel it is worth it, in which case the best course of action is to add comments that describe the shortcuts that are being taken. That way the reader/reviewer knows you have considered the issue and actively made the choice you did, rather than have missed it altogether.

I would rename `storage_test.go` to `storer_test.go` as it is running the tests of the `Storer` interface.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,75 @@
+package main
+
+import "fmt"
+
+// Error codes
+const (
+	ErrUnknown int = iota
```
golangcibot: `ErrUnknown` is unused (from `deadcode`)

06_puppy/mohankrishna/main.go
```diff
@@ -0,0 +1,61 @@
...
+			ID:     11,
+			Breed:  "Sheep herder",
+			Colour: "Brown",
+			Value:  1000,
+		}
+	}
+
+	//Map store
+	newPup := puppy()
+	mapStorage.CreatePuppy(&newPup)
```
golangcibot: Error return value of `mapStorage.CreatePuppy` is not checked (from `errcheck`)

06_puppy/mohankrishna/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	pup, err := mapStorage.ReadPuppy(newPup.ID)
+	if err == nil {
+		fmt.Fprintln(out, *pup)
+	}
+
+	pup.Value = 10000
+	mapStorage.UpdatePuppy(pup.ID, pup)
```
golangcibot: Error return value of `mapStorage.UpdatePuppy` is not checked (from `errcheck`)

06_puppy/mohankrishna/main.go
```diff
@@ -0,0 +1,61 @@
...
+
+	ok, _ := mapStorage.DeletePuppy(pup.ID)
+	if ok {
+		_, err = mapStorage.DeletePuppy(11)
+		fmt.Fprintln(out, err)
+	}
+
+	//Sync.map store
+	newPup = puppy()
+	syncStorage.CreatePuppy(&newPup)
```
golangcibot: Error return value of `syncStorage.CreatePuppy` is not checked (from `errcheck`)

06_puppy/mohankrishna/main.go
```diff
@@ -0,0 +1,61 @@
...
+	pup, err = syncStorage.ReadPuppy(11)
+	if err == nil {
+		fmt.Fprintln(out, *pup)
+	}
+
+	pup.Value = 10000
+
+	syncStorage.UpdatePuppy(pup.ID, pup)
```
golangcibot: Error return value of `syncStorage.UpdatePuppy` is not checked (from `errcheck`)

06_puppy/mohankrishna/storer_test.go
```diff
@@ -0,0 +1,217 @@
...
+		// create a map store
+		s.store = NewMapStore()
+	case smaps:
+		// create a sync map store
+		s.store = NewSyncStore()
+	default:
+		panic("Unrecognised storer implementation")
+	}
+	puppy := puppy1()
+	s.store.CreatePuppy(&puppy)
```
golangcibot: Error return value of `s.store.CreatePuppy` is not checked (from `errcheck`)

mohankrishna:
@camh-anz @juliaogris 
I have updated PR with some changes
1. Add error handling
2. No more empty puppies, made everything pointer access
3. Even though `CreatePuppy` method is not assigning the IDs yet, it will return an error if a puppy with the same id already exists. Now I can safely say "No puppies will be harmed during the execution of this program" :)
4. Added a little check in the `UpdatePuppy` method to check whether the passed `id` and `puppy.id` are equal or not.

06_puppy/mohankrishna/map_store.go
```diff
@@ -0,0 +1,44 @@
...
+	return nil
+}
+
+func (ms *MapStore) DeletePuppy(id uint32) (bool, error) {
+	_, ok := ms.m[id]
+	if ok {
+		delete(ms.m, id)
+		return ok, nil
+	}
+	return ok, &Error{ErrDuplicate, fmt.Sprintf("No puppy exists with id %d", id)}
```
camscale, line 43: I think this error code should be `ErrNotFound`.

06_puppy/mohankrishna/storer_test.go
```diff
@@ -0,0 +1,220 @@
...
+	tassert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+type storerImpl int
+
+const (
+	maps  storerImpl = iota
+	smaps storerImpl = iota
```
camscale, line 16: while this works, it is more idiomatic go to just say `smaps` here, so the whole const block reads:
```
const (
    maps storerImpl = iota
    smaps
)
```

See https://github.com/golang/go/wiki/Iota for an example and links to specs, etc.

06_puppy/mohankrishna/storer_test.go
```diff
@@ -0,0 +1,220 @@
...
+	_, err := s.store.DeletePuppy(0x123234)
+
+	// then
+	assert.Error(err, "Should get an error deleting a non existant ID")
+}
+
+func TestStorerImpls(t *testing.T) {
+	mapSuite := storerSuite{
+		impl: maps,
+	}
```
camscale, line 197: Idiomatic go would usually be to have this on one line:
```
    mapSuite := storerSuite{impl: maps}
```

Vertical whitespace can be valuable in making code clearer, but should only be used when necessary (judgement call, of course).

06_puppy/mohankrishna/sync_store.go
```diff
@@ -0,0 +1,50 @@
...
+	return nil
+}
+
+func (ss *SyncStore) DeletePuppy(id uint32) (bool, error) {
+	_, ok := ss.sm.Load(id)
+	if ok {
+		ss.sm.Delete(id)
+		return ok, nil
+	}
+	return ok, &Error{ErrDuplicate, fmt.Sprintf("No puppy exists with id %d", id)}
```
camscale, line 49: This error code should be `ErrNotFound`.

06_puppy/mohankrishna/types.go
```diff
@@ -0,0 +1,31 @@
...
+	Breed  string
+	Colour string
+	Value  int
+}
+
+type Storer interface {
+	CreatePuppy(puppy *Puppy) error
+	ReadPuppy(ID uint32) (*Puppy, error)
+	UpdatePuppy(ID uint32, puppy *Puppy) error
+	DeletePuppy(ID uint32) (bool, error)
```
camscale, line 30: It is questionable as to whether `DeletePuppy` should return two values. The way it is currently being used means there is no overlap between the bool and the error, so you may as well just return an error. If the error is `nil`, then all is ok. If it is not, then it's not ok.

This might change if you decide that it is not an error to delete a puppy that does not exist. Then `ok` returns something valuable. The error would then be for implementations where trying to delete leads to errors outside your control, such as when using an external database. If the connection to the database is lost or some other database error, an error would be returned. This would mean your current implementations have no cases where it would return an error.

camscale (approved):
LGTM. Thanks for adding errors. It's a worthwhile point to get to as you'll soon learn, with Go you'll be dealing with errors a lot :)

My comments are mostly just about style and idiomatic go, so there's no need to update this PR again to address those. Just something to take on board as you carry on more Go development. The best thing to do at this point is read lots of Go code (the standard Go library would be a good place to start) to see how Go code is typically written. It has a more terse style than many other languages.

There are a couple of points where I think you're using the wrong error code - again not so important to get fixed up for the lab, since nothing actually uses the error code.

I'll double-check with @juliaogris to see if she's happy to have her review dismissed, in which case you can then merge.

juliaogris (approved):
Great work @mohankrishna and yet another thank you to @camh-anz for incredible patience and so many hours dedicated.

mohankrishna:
Thank you, @juliaogris & @camh-anz. I am planning on working through all the remaining labs. Would you be able to review them?

camscale:
Sure @mohankrishna . Happy to review any PRs you raise.

PR #101, opened 2019-02-14
Lab 06 CRUD Puppy

### LINKED ISSUE #88: Lab 6 - CRUD puppy with interface

sahidakhtar1:
Fixes #88 

- Create Struct `Puppy`
- Create interface `Storer` with CRUD methods
- Implement `MapStore` on `Storer` backed by a map
- Implement `SyncStore` on `Storer` backed by a sync.Map

### DIFF
--- a/06_puppy/sahidakhtar1/main.go
+++ b/06_puppy/sahidakhtar1/main.go
@@ -24,7 +24,7 @@ type MapStore struct {
 
 func newMapStore() *MapStore {
 	ms := MapStore{}
-	ms.m = make(map[uint]Puppy, 0)
+	ms.m = make(map[uint]Puppy)
 	return &ms
 }
 func (ms *MapStore) CreatePuppy(p Puppy) {
@@ -52,7 +52,7 @@ func (ms *MapStore) DeletePuppy(id uint) bool {
 
 //Map Store Ends
 
-//Sync Store Ends
+//Sync Store Starts
 type SyncStore struct {
 	sync.Mutex
 	sync.Map
@@ -94,7 +94,7 @@ func (m *SyncStore) DeletePuppy(id uint) bool {
 //Sync Store Ends
 
 func main() {
-	ms := newMapStore()
+	var ms Storer = newMapStore()
 	p1 := Puppy{1, "Bulldog", "White", "100"}
 	p2 := Puppy{2, "Poddle", "Black", "200"}
 	ms.CreatePuppy(p1)
@@ -107,11 +107,10 @@ func main() {
 	fmt.Println(rp1)
 	rp2 := ms.ReadPuppy(2)
 	fmt.Println(rp2)
-	ms.DeletePuppy(2)
-	rp2 = ms.ReadPuppy(2)
-	fmt.Println(rp2)
+	delete := ms.DeletePuppy(2)
+	fmt.Println(delete)
 
-	s := newSyncStore()
+	var s Storer = newSyncStore()
 	p4 := Puppy{4, "Beagle", "White", "400"}
 	p5 := Puppy{5, "Pug", "Black", "500"}
 	s.CreatePuppy(p4)
@@ -122,8 +121,7 @@ func main() {
 	s.UpdatePuppy(4, p4)
 	rp4 = s.ReadPuppy(4)
 	fmt.Println(rp4)
-	s.DeletePuppy(5)
-	rp5 := s.ReadPuppy(5)
-	fmt.Println(rp5)
+	delete = s.DeletePuppy(5)
+	fmt.Println(delete)
 
 }
--- a/06_puppy/sahidakhtar1/main.go
+++ b/06_puppy/sahidakhtar1/main.go
@@ -12,9 +12,9 @@ type Puppy struct {
 
 type Storer interface {
 	CreatePuppy(Puppy)
-	ReadPuppy(int) Puppy
-	UpdatePuppy(int, Puppy)
-	DeletePuppy(int) bool
+	ReadPuppy(uint) Puppy
+	UpdatePuppy(uint, Puppy)
+	DeletePuppy(uint) bool
 }
 
 //Map Store Starts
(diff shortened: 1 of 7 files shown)


06_puppy/sahidakhtar1/main.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+type Storer interface {
+	CreatePuppy(Puppy)
+	ReadPuppy(uint) Puppy
+	UpdatePuppy(uint, Puppy)
+	DeletePuppy(uint) bool
+}
+
+//Map Store Starts
```
juliaogris: I'd create 3 more files for this:
`storer.go`,  `mapstore.go` and `sunystore.go`.
As long as they are all in the same directory and contained in package `main` it shouldn't be an issue.

06_puppy/sahidakhtar1/main.go
```diff
@@ -0,0 +1,127 @@
...
+	m map[uint]Puppy
+}
+
+func newMapStore() *MapStore {
+	ms := MapStore{}
+	ms.m = make(map[uint]Puppy)
+	return &ms
+}
+func (ms *MapStore) CreatePuppy(p Puppy) {
+	_, ok := ms.m[p.ID]
```
juliaogris: Create should take care of `ID`.
Consider adding a `maxID uint` field to `MapStore` and using here like this:
```go
ms.maxID++
p.ID = maxID
ms.m[p.ID] = p
```
This is not thread/go routine save, but it will suffice as a "toy" implementation.

06_puppy/sahidakhtar1/main.go
```diff
@@ -0,0 +1,127 @@
...
+	if ok {
+		return
+	}
+	ms.m[p.ID] = p
+}
+func (ms *MapStore) ReadPuppy(id uint) Puppy {
+	return ms.m[id]
+}
+func (ms *MapStore) UpdatePuppy(id uint, p Puppy) {
+	ms.m[id] = p
```
juliaogris: maybe add: `p.ID = id`

06_puppy/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,140 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMainOutPut(t *testing.T) {
```
juliaogris, line 11: This is not testing anything. Remove.

06_puppy/sahidakhtar1/main_test.go
```diff
@@ -0,0 +1,140 @@
...
+	//Just not checking the cosole out put
+	r := require.New(t)
+	main()
+	// Then
+	r.Equalf("", "", "Unexpected output in main()")
+}
+
+func TestCreatePuppyInMapStore(t *testing.T) {
```
juliaogris: Split the test cases into files in analogy to the implementation (`mapstore_test.go`, etc.)

juliaogris (changes requested):
Looks pretty solid!

Maybe just update the file structure as suggested in comments.

06_puppy/sahidakhtar1/main_test.go
sahidakhtar1: Just added one line out put test to achieve 100% coverage.

06_puppy/sahidakhtar1/main.go
sahidakhtar1: Done

06_puppy/sahidakhtar1/main.go
sahidakhtar1: Created separate files for each storer and its corresponding  testing file.

06_puppy/sahidakhtar1/main.go
sahidakhtar1: As pointer is not used.. it won't update the Puppy. Here `Puppy{1,"Bulldog","White","100"}` will be updated against the `id` passed.

06_puppy/sahidakhtar1/main_test.go
sahidakhtar1: Done

06_puppy/sahidakhtar1/mapstore.go
```diff
@@ -0,0 +1,33 @@
...
+
+type MapStore struct {
+	m     map[uint]Puppy
+	maxID uint
+}
+
+func newMapStore() *MapStore {
+	ms := MapStore{}
+	ms.m = make(map[uint]Puppy)
+	return &ms
```
camscale, line 11: I would write this function as:
```
    return &MapStore{m: make(map[uint]Puppy)}
```
or even
```
    return &MapStore{m: map[uint]Puppy{}}
```

Whether you use `make` (option 1) or a map literal (option 2) is a matter of style, but most Go code I see uses the map literal form.

Putting the make/literal inside the Puppy literal is described more here: https://golang.org/doc/effective_go.html#composite_literals

There is no need to make this change for this lab. What you have is fine. This is just something to keep in mind for next time.

06_puppy/sahidakhtar1/main_test.go
camscale: This test still isn't right, because it is not testing that the output of `main()` is correct, just a small part of it. `main()` should be changed to use `fmt.Fprint*(out, ...)` calls so all output can be captured by this test case.

Furthermore, you should not be creating mapStores, puppies or anything else in this test function because they are not the things this test is for. This test is for `main()` so it should only call `main()` verify that it performed correctly.

camscale (commented):
The code itself here is great, but the tests still need a little work. The `main()` test needs to be re-worked as described, but more importantly, `mapstore_test.go` and `syncstore_test.go` should be a single set of tests for `storer_test.go` as describe in the Lab 6 slide.

That is, rather than testing `mapStore` and `syncStore` directly, you set up two test suites where each creates an instance of each of those stores, but the tests run against a `Storer` interface. This verifies that each implementation has the same behaviour (and is correct) when used as the interface. Sometimes you may add tests for the struct types if there are features that are not available through the `Storer` interface, but in this lab this is not the case.

juliaogris:
Hi,
Thanks!
Julia

PR #105, opened 2019-02-14
Lab 6 - Create CRUD puppy with interface

### LINKED ISSUE #88: Lab 6 - CRUD puppy with interface

madhavansu:
Fixes #88 

- Create an executable go program in directory 06_puppy/USERNAME (see hints)
- Implement a Puppy struct containing ID, Breed, Colour, Value.
- Create Storer interface with CRUD methods for Puppy
- Write a MapStore implementation of Storer backed by a map
- Write a SyncStore implementation of Storer backed by a sync.Map
- Keep all implementation files in the same folder and in package main
- Test against the Storer interface and run in suite with both implementations

### DIFF
--- a/06_puppy/madhavansu/main.go
+++ b/06_puppy/madhavansu/main.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+var out io.Writer = os.Stdout
+
+//Sync store
+type syncStore struct {
+	sync.Mutex
+	sync.Map
+}
+
+func newSyncStore() *syncStore {
+	return &syncStore{}
+}
+
+func (s *syncStore) createPuppy(in Puppy) {
+	s.Store(in.id, in)
+}
+
+func (s *syncStore) readPuppy(id uint32) Puppy {
+	pd, ok := s.Load(id)
+	if !ok {
+		fmt.Printf("No puppy exists\n")
+	}
+	p, _ := pd.(Puppy)
+	return p
+}
+
+func (s *syncStore) updatePuppy(id uint32, in Puppy) {
+	s.Store(in.id, in)
+}
+
+func (s *syncStore) deletePuppy(id uint32) {
+	s.Delete(id)
+}
+
+// Map Store
+type mapStore struct {
+	ms map[uint32]Puppy
+}
+
+func newMapStore() *mapStore {
+	return &mapStore{make(map[uint32]Puppy)}
+}
+
+func (m *mapStore) createPuppy(in Puppy) {
+	m.ms[in.id] = in
+}
+
+func (m *mapStore) readPuppy(id uint32) Puppy {
+	p, ok := m.ms[id]
+	if !ok {
+		fmt.Printf("No puppy exists\n")
+	}
+	return p
+}
+
+func (m *mapStore) updatePuppy(id uint32, in Puppy) {
+	m.ms[id] = in
+}
+
+func (m *mapStore) deletePuppy(id uint32) {
+	_, ok := m.ms[id]
+	if ok {
+		delete(m.ms, id)
+	}
+}
+
+func main() {
+	// Sync storer
+	var s Storer = newSyncStore()
+	fmt.Fprintln(out, "~~~~~~~~~~")
+	fmt.Fprintln(out, "Sync Store")
+	fmt.Fprintln(out, "~~~~~~~~~~")
+	// Create/Read puppy
+	s.createPuppy(Puppy{101, "Poodle", "red", 18000})
+	fmt.Fprintln(out, "101 : ", s.readPuppy(101))
+	// Delete puppy
+	s.createPuppy(Puppy{102, "Bulldog", "brownish", 9999})
+	s.deletePuppy(102)
+	fmt.Fprintln(out, "102 : ", s.readPuppy(102))
+	// Update puppy
+	s.createPuppy(Puppy{103, "Labrador Retriever", "purple", 987})
+	s.updatePuppy(103, Puppy{103, "German Shepherd", "red", 4533})
+	fmt.Fprintln(out, "103 : ", s.readPuppy(103))
+
+	// Map storer
+	var m Storer = newMapStore()
+	fmt.Fprintln(out, "~~~~~~~~~")
+	fmt.Fprintln(out, "Map Store")
+	fmt.Fprintln(out, "~~~~~~~~~")
+	// Create/Read puppy
+	m.createPuppy(Puppy{104, "Pug", "brown", 0.91})
+	fmt.Fprintln(out, "104 : ", m.readPuppy(104))
+	// Delete puppy
+	m.createPuppy(Puppy{105, "Beagle", "yellowish", 1233})
+	m.deletePuppy(105)
+	fmt.Fprintln(out, "105 : ", m.readPuppy(105))
+	// Update puppy
+	m.createPuppy(Puppy{106, "Boxer", "black", 45000.98})
+	m.updatePuppy(106, Puppy{106, "Beagle", "brown", 0.91})
+	fmt.Fprint(out, "106 :  ", m.readPuppy(106))
+}
--- a/06_puppy/madhavansu/main.go
+++ b/06_puppy/madhavansu/main.go
@@ -42,15 +42,19 @@ func (s *syncStore) deletePuppy(id uint) {
 
 // Map Store
 type mapStore struct {
-	ms map[uint]Puppy
+	ms    map[uint]Puppy
+	mapID uint
 }
 
 func newMapStore() *mapStore {
-	return &mapStore{make(map[uint]Puppy)}
+	return &mapStore{make(map[uint]Puppy), 0}
 }
 
-func (m *mapStore) createPuppy(in Puppy) {
+func (m *mapStore) createPuppy(in Puppy) uint {
+	m.mapID++
+	in.id = m.mapID
 	m.ms[in.id] = in
+	return m.mapID
 }
 
 func (m *mapStore) readPuppy(id uint) Puppy {
@@ -91,19 +95,20 @@ func main() {
 	fmt.Fprintln(out, "103 : ", s.readPuppy(103))
 
 	// Map storer
-	var m Storer = newMapStore()
+	var m StorerMap = newMapStore()
+	var puppyID uint
 	fmt.Fprintln(out, "~~~~~~~~~")
 	fmt.Fprintln(out, "Map Store")
 	fmt.Fprintln(out, "~~~~~~~~~")
 	// Create/Read puppy
-	m.createPuppy(Puppy{104, "Pug", "brown", 0.91})
-	fmt.Fprintln(out, "104 : ", m.readPuppy(104))
+	puppyID = m.createPuppy(Puppy{104, "Pug", "brown", 0.91})
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
 	// Delete puppy
-	m.createPuppy(Puppy{105, "Beagle", "yellowish", 1233})
-	m.deletePuppy(105)
-	fmt.Fprintln(out, "105 : ", m.readPuppy(105))
+	puppyID = m.createPuppy(Puppy{105, "Beagle", "yellowish", 1233})
+	m.deletePuppy(puppyID)
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
 	// Update puppy
-	m.createPuppy(Puppy{106, "Boxer", "black", 45000.98})
-	m.updatePuppy(106, Puppy{106, "Beagle", "brown", 0.91})
-	fmt.Fprint(out, "106 :  ", m.readPuppy(106))
+	puppyID = m.createPuppy(Puppy{106, "Boxer", "black", 45000.98})
+	m.updatePuppy(puppyID, Puppy{puppyID, "Beagle", "brown", 0.91})
+	fmt.Fprint(out, puppyID, " :  ", m.readPuppy(puppyID))
 }
--- a/06_puppy/madhavansu/main.go
+++ b/06_puppy/madhavansu/main.go
@@ -23,7 +23,7 @@ func (s *syncStore) createPuppy(in Puppy) {
 	s.Store(in.id, in)
 }
 
-func (s *syncStore) readPuppy(id uint32) Puppy {
+func (s *syncStore) readPuppy(id uint) Puppy {
 	pd, ok := s.Load(id)
 	if !ok {
 		fmt.Printf("No puppy exists\n")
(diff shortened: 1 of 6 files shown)


06_puppy/madhavansu/types.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+// Puppy defines the attributes corresponding to puppy
+type Puppy struct {
```
golangcibot, line 4: struct of size 48 bytes could be of size 40 bytes (from `maligned`)

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,206 @@
+package main
+/**
+* Map store implemented in below scenarios
+* Dynamic Map ID: MapStore implementation of Storer backed by a map
+* Static Map ID: SyncStore implementation of Storer backed by a sync.Map
+*/
+import (
```
golangcibot, line 3: File is not `goimports`-ed (from `goimports`)

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+	"io"
+	"os"
+	"sync"
+)
+
+var out io.Writer = os.Stdout
+
+//Sync store
+type syncStore struct {
+	sync.Mutex
```
camscale: This embedded struct, `sync.Mutex` is unused. Please either remove it or justify its inclusion with a comment.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+	sync.Map
+}
+
+func newSyncStore() *syncStore {
+	return &syncStore{}
+}
+
+func (s *syncStore) createPuppy(in Puppy) {
+	s.Store(in.id, in)
```
camscale: Consider whether or not you should be generating the IDs for puppys instead of requiring the called provide it. How does the caller know what IDs are already taken?

I see you have `mapID` in the `mapStore` implementation. You should use a similar pattern here.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (s *syncStore) readPuppy(id uint) Puppy {
+	pd, ok := s.Load(id)
+	if !ok {
+		fmt.Printf("No puppy exists\n")
```
camscale: This should not be printing anything. This is a "library" function so all input and output should be via the parameters and the return values.

The current code will return the zero value of a puppy due to the type assertion failing (since a `nil` `interface{}` is not a `Puppy`), so you could be explicit and instead of printing the error, just `return Puppy{}`.

See https://tour.golang.org/methods/15 for the return values for failed type assertions/checks.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// Map Store
+type mapStore struct {
+	ms    map[uint]Puppy
+	mapID uint
+}
+
+func newMapStore() *mapStore {
+	return &mapStore{make(map[uint]Puppy), 0}
```
camscale: There is no need to initialise fields to their zero value - go does that for us already. To partially initialise a struct, you need to use named fields, like this:
```
    return &mapStore{ms: make(map[uint]Puppy)}
```

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+	m.mapID++
+	in.id = m.mapID
+	m.ms[in.id] = in
+	return m.mapID
+}
+
+func (m *mapStore) readPuppy(id uint) Puppy {
+	p, ok := m.ms[id]
+	if !ok {
+		fmt.Printf("No puppy exists\n")
```
camscale: Same comment applies here regarding printing this message as in the `syncStore` implementation.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,114 @@
...
+	puppyID = m.createPuppy(Puppy{104, "Pug", "brown", 0.91})
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
+	// Delete puppy
+	puppyID = m.createPuppy(Puppy{105, "Beagle", "yellowish", 1233})
+	m.deletePuppy(puppyID)
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
+	// Update puppy
+	puppyID = m.createPuppy(Puppy{106, "Boxer", "black", 45000.98})
+	m.updatePuppy(puppyID, Puppy{puppyID, "Beagle", "brown", 0.91})
+	fmt.Fprint(out, puppyID, " :  ", m.readPuppy(puppyID))
```
camscale: use `fmt.Fprintln` here so it properly outputs a newline at the end. For reasons I don't understand, this also adds the extra space so the colons line up, so also remove that extra space after the colon that you have that lines up the values. Don't forget to update your tests.

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,207 @@
...
+
+	tassert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/stretchr/testify/suite"
+)
+
+/**
+* Map store implemented in below scenarios
+* Dynamic Map ID: MapStore implementation of Storer backed by a map
+* Static Map ID: SyncStore implementation of Storer backed by a sync.Map
```
camscale: This static/dynamic comment may change how you respond to my previous comments. However that means these comments belong on the structs you define and not here in the tests. Please move them if you retain the static/dynamic division.

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,207 @@
...
+	puppy108 = func() Puppy {
+		return Puppy{
+			id:    108,
+			breed: "Bulldog",
+			color: "red",
+			value: 842.50,
+		}
+	}
+	puppyEmpty = func() Puppy {
+		return Puppy{
```
camscale: Change this to `return Puppy{}` - let Go fill in the zero values.

06_puppy/madhavansu/types.go
```diff
@@ -0,0 +1,28 @@
...
+	id    uint
+	breed string
+	color string
+	value float32
```
camscale: Since `value` is presumably an amount of money, using a `float32` is not a good idea. Here's one article on that, but there are plenty more: https://dzone.com/articles/never-use-float-and-double-for-monetary-calculatio

06_puppy/madhavansu/types.go
```diff
@@ -0,0 +1,28 @@
...
+	createPuppy(Puppy)
+	readPuppy(uint) Puppy
+	updatePuppy(uint, Puppy)
+	deletePuppy(uint)
+}
+
+// StorerMap defines standard CRUD operations for puppy
+// This has been created for dynamic ID implementation for each puppy
+// I would like to know about extending the "Storer" interface "createPuppy" method with a return integer,
+// Otherwise I can use the same interface for MapStore and SyncStore
```
camscale: I suggest that you define a single interface with `createPuppy(Puppy) uint` and in the other implementation, just return `puppy.id`.

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,207 @@
...
+* Map store implemented in below scenarios
+* Dynamic Map ID: MapStore implementation of Storer backed by a map
+* Static Map ID: SyncStore implementation of Storer backed by a sync.Map
+ */
+type storerImpl int
+
+var mapID uint
+
+const (
+	mem storerImpl = iota
```
camscale: This value is meant to determine which of the two implementations are tested, so there should be two consts defined, one for `mapStore` and one for `syncStore`. The implementation is then tested via the `Storer` interface.

Then when you set up the test suite, you set up two test suites, one for each implementation.

When you make this change, you can update your tests to do all the operations once only, instead of twice as it is now.

camscale (commented):
This is reasonably close, but the structure of the tests needs some work. The rest are some general comments on how you could improve the main code. Please feel free to ask any questions if I wasn't clear in any of my comments.

06_puppy/madhavansu/main.go
madhavansu: I have purposefully followed two different approach. SyncStore keeps whatever receives and mapStore generates a dynamic ID

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main_test.go
madhavansu: Updated.

06_puppy/madhavansu/main_test.go
madhavansu: Updated.

06_puppy/madhavansu/types.go
madhavansu: Updated.

06_puppy/madhavansu/types.go
madhavansu: Updated.

06_puppy/madhavansu/main_test.go
madhavansu: Updated.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (s *syncStore) updatePuppy(id uint, in Puppy) {
+	s.Store(in.id, in)
+}
+
+func (s *syncStore) deletePuppy(id uint) {
+	s.Lock()
+	defer s.Unlock()
+	s.Delete(id)
```
anzboi: The `Delete` method for `sync.Map` is already thread-safe, you don't need to `Lock` and `Unlock` this call.

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+func (m *mapStore) updatePuppy(id uint, in Puppy) {
+	m.ms[id] = in
+}
+
+func (m *mapStore) deletePuppy(id uint) {
+	_, ok := m.ms[id]
+	if ok {
+		delete(m.ms, id)
```
anzboi: `delete` works whether or not something was actually deleted. You don't need to check first.

06_puppy/madhavansu/main.go
anzboi: use `FPrintln`, otherwise you get yucky terminal output.

anzboi (commented):
This is pretty good, a few more comments to address.

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main.go
madhavansu: Updated.

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,244 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := "~~~~~~~~~~\nSync Store\n~~~~~~~~~~\n101 :  {101 Poodle red 18000}\n102 :  {0   }\n103 :  {103 German Shepherd red 4533}\n~~~~~~~~~\nMap Store\n~~~~~~~~~\n1  :  {1 Pug brown 0.91}\n2  :  {0   }\n3  :  {3 Beagle brown 0.91}\n"
```
golangcibot: line is 238 characters (from `lll`)

juliaogris (changes requested):
Looks pretty solid - the only recommendation I have is to break up your source code into a few different files:

- `main.go`
- `storer.go` including `Puppy` definition
- `storer_test.go`
- `mapstore.go`
- `syncstore.go`

madhavansu:
@juliaogris break up of source code is updated

camscale (commented):
The code itself here is great. But the tests still need some more work. One of the important parts of this lab is "* Test against the Storer interface and run in suite with both implementations". The tests you have are testing against the specific implementations by using separate `mapstore` and `syncstore` objects of the concrete types. Your tests then operate on one or the other of these.

Instead, set up your tests like this:
```
type storerImpl int
const (
    smem storerImpl = iota
    mmem
)

type storerSuite struct {
    suite.Suite
    storer Storer
    impl storerImpl
}

func TestStorer(t *testing.T) {
    mmemSuite := storerSuite{impl: mmem}
    suite.Run(t, &mmemSuite)
    smemSuite := storerSuite{impl: smem}
    suite.Run(t, &smemSuite)
}

func (s *storerSuite) SetupTests() {
    switch s.impl {
    case smem:
        s.storer = newSyncStore()
    case mmem:
        s.storer = newMemStore()
    default:
        panic("Unknown storer implementation")
    }
    s.storer.createPuppy(puppyIN())
}
```
 
Then each test just runs against `s.storer` and remove the duplication that exists, such as with `TestReadPuppySuccessful()` and `TestMapReadPuppySuccessful()`. Each of those tests do the same thing against different implementations, and can be reduced to a single test against the `Storer` interface instead.

This is a rough outline, and I haven't compiled the above code, so there may be some bugs in it. But when done, it should reduce your test suite by half.

madhavansu:
@camh-anz @juliaogris I have updated the test suite and created a separate file for each map store and sync store

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+/**
```
juliaogris, line 11: Godocs typically start with // FunctionName:
https://blog.golang.org/godoc-documenting-go-code - I'm surprised the linter didn't pick this up. I think your comments actually belong to the top of the file/ package level....

06_puppy/madhavansu/main.go
```diff
@@ -0,0 +1,51 @@
...
+	puppyID = m.createPuppy(Puppy{104, "Pug", "brown", "0.91"})
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
+	// Delete puppy
+	puppyID = m.createPuppy(Puppy{105, "Beagle", "yellowish", "1233"})
+	m.deletePuppy(puppyID)
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
+	// Update puppy
+	puppyID = m.createPuppy(Puppy{106, "Boxer", "black", "45000.98"})
+	m.updatePuppy(puppyID, Puppy{puppyID, "Beagle", "brown", "0.91"})
+	fmt.Fprintln(out, puppyID, " : ", m.readPuppy(puppyID))
```
juliaogris, line 50: I would suggest to keep `main()` pretty light weight. You have got it all covered in the tests. Next step would be to start up an HTTP server in main....

06_puppy/madhavansu/main_test.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+)
+
+type storerSuite struct {
+	suite.Suite
+	storer Storer
+	impl   storerImpl
+}
+
+func (s *storerSuite) SetupTest() {
```
juliaogris, line 49: Nice.

juliaogris (approved):
Looks pretty good to me! Just a couple of minor comments, feel free to ignore.

PR #117, opened 2019-02-27
Lab 03 - Letter frequency

JackLeeHal:
Fixes #63

* Implement the letter frequency
* Add test cases
* Fix CI issue

### DIFF
--- a/03_letters/jacklee/letter.go
+++ b/03_letters/jacklee/letter.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	m := map[rune]int{}
+	for _, v := range s {
+		m[v]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	l := len(m)
+	keys := make([]int, 0, l)
+	for v := range m {
+		keys = append(keys, int(v))
+	}
+	sort.Ints(keys)
+
+	s := make([]string, 0, l)
+	for _, k := range keys {
+		s = append(s, fmt.Sprintf("%c:%d", k, m[rune(k)]))
+	}
+	return s
+}
--- a/03_letters/jacklee/letter_test.go
+++ b/03_letters/jacklee/letter_test.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


03_letters/jacklee/letter.go
```diff
@@ -0,0 +1,38 @@
...
+	m := map[rune]int{}
+	for _, v := range s {
+		m[v]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	l := len(m)
+	keys := make([]int, 0, l)
```
anzboi, line 27: I was going to ask you to change this to use a fixed size array, populated directly with an index variable, then did some research and found out this method isn't much slower (benchmarks ~20% slower, pretty sure no resizing is going on). Nice work on finding this method 👍. I'll leave it up to you whether you want to change this to gain that 20% speedup.

anzboi (approved):
LGTM 👍
See the comment about your array populating method. It's a little slower than optimal but not a big issue.

03_letters/jacklee/letter.go
anzdaddy, line 27: Why not `[]rune`?

03_letters/jacklee/letter_test.go
```diff
@@ -0,0 +1,54 @@
...
+	r.Equal(expected, actual, "Unexpected output")
+}
+
+func TestLetters(t *testing.T) {
+	for _, d := range testData {
+		// When
+		l := letters(d.in)
+
+		// Then
+		require.Equal(t, d.letter, l, "unexpected output")
```
anzdaddy, line 42: No need to explicitly say what went wrong ("unexpected output") since `require.Equal` provides a detailed description and the maintainer can read the code. Instead, include any necessary context in the error output to make it easy for the engineer to identify where the problem occurred:
```go
for i, d … {
	⋮
	require.Equal(t, d.letter, l, "%d", i)
```

03_letters/jacklee/letter_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+}
+
+func TestSortLetters(t *testing.T) {
+	for _, d := range testData {
+		// When
+		s := sortLetters(d.letter)
+
+		// Then
+		require.Equal(t, d.out, s, "unexpected output")
```
anzdaddy, line 52: No need to explicitly say what went wrong ("unexpected output") since `require.Equal` provides a detailed description and the maintainer can read the code. Instead, include any necessary context in the error output to make it easy for the engineer to identify where the problem occurred:
```go
for i, d … {
	⋮
	require.Equal(t, d.letter, l, "%d", i)
```

anzdaddy (changes requested):

juliaogris:
Sorry, I will close this PR now in preparation for the next training in a few weeks in Melbourne - feel free to re-open it if you want to continue work on it.

PR #127, opened 2019-05-09
Improve lab README code blocks

anzboi:
So I actually had a look at the README's in github and some of the code blocks were broken up due to the extra back-ticks from the slides. Didn't notice it in an editor.

Also added back-ticks around lab1 directory for consistency.


juliaogris (approved):
Awesome, thanks for fixing this up.

Merge as `Improve lab...` rather than `Improving lab...` please.

LGTM

PR #129, opened 2019-05-15
Remove testify from hello world test

### LINKED ISSUE #128: Remove testify from hello world test
We want to encourage a basic understanding of testing in go works, which has been identified as a problem and is made worse by the reliance on rich third party libraries.

We want to start with the native `testing` package that all tests must use, before introducing `testify` that hides `testing` behind a richer framework.

anzboi:
This encourages people to start by using the testing.T object
directly and gain a better understanding of how testing works in go
before moving to richer test libraries.

### DIFF
--- a/00_hello_world/juliaogris/main_test.go
+++ b/00_hello_world/juliaogris/main_test.go
@@ -4,21 +4,18 @@ import (
 	"bytes"
 	"strconv"
 	"testing"
-
-	"github.com/stretchr/testify/require"
 )
 
 func TestMainOutput(t *testing.T) {
-	// Given
-	r := require.New(t)
 	var buf bytes.Buffer
 	out = &buf
 
-	// When
 	main()
 
-	// Then
 	expected := strconv.Quote("Hallo du schöne Welt!\n")
 	actual := strconv.Quote(buf.String())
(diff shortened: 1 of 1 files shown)


00_hello_world/juliaogris/main_test.go
```diff
@@ -4,21 +4,18 @@ import (
...
-	r := require.New(t)
 	var buf bytes.Buffer
 	out = &buf
 
-	// When
 	main()
 
-	// Then
 	expected := strconv.Quote("Hallo du schöne Welt!\n")
 	actual := strconv.Quote(buf.String())
```
juliaogris, line 16: I've read a few times `got` and `want` instead of `actual` and `expected`- up to you if want to change it.
See:

 - https://github.com/golang/go/wiki/TableDrivenTests#example-of-a-table-driven-test
- https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

juliaogris (approved):
LGTM, just one minor comment, up to you if you want to address this.

00_hello_world/juliaogris/main_test.go
anzboi: I really don't think there's much of a difference, less characters I suppose.

PR #130, opened 2019-05-20
Lab 1 - Fibonacci and Negafibonacci

kasunfdo:
Fixes #24 

#### Changes proposed in this PR:
* Implement Fibonacci and Negafibonacci series
* Add unit tests

PR #131, opened 2019-05-20
Lab01 - Implement Fibonacci series

### LINKED ISSUE #146: Lab 1 - fibonacci series
Lab 1:

Provide solution for: https://github.com/anz-bank/go-course/tree/master/01_fib

n0npax:
Provide Fibonacci and Negafibonacci functions.
Provide dispatcher to return proper series base on input.
Provide tests for implementation.

Fixes #146 

Review of colleague's PR #130 

* Implement Fibonacci and Negafibonacci series
* Add unit tests

### DIFF
--- a/01_fib/n0npax/main.go
+++ b/01_fib/n0npax/main.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+)
+
+// Allow to overwtite stream by tests
+var out io.Writer = os.Stdout
+
+// This function is useless and it's provided just due to requirements
+func main() {
+	fib(7)
+}
+
+// Wrapper function to print computed Fibonacci numbers
+func fib(num int) {
+	for _, v := range dispatchFib(num) {
+		fmt.Fprintln(out, v)
+	}
+}
+
+// Dispatch and return proper numbers
+func dispatchFib(num int) []int {
+	switch {
+	case num == 0:
+		return []int{0}
+	case num < 0:
+		return computeNegafibonacci(num)
+	default:
+		return computeFibonacci(num)
+
+	}
+}
+
+// Compute Fibonacci numbers
+func computeFibonacci(num int) []int {
+	fibNumbers := make([]int, num)
+	a, b := 1, 1
+	for i := range fibNumbers {
+		fibNumbers[i] = a
+		a, b = b, a+b
+	}
+	return fibNumbers
+}
+
+// Compute negafibonacci numbers
+func computeNegafibonacci(num int) []int {
+	fibNumbers := computeFibonacci(num * -1)
+	for i, v := range fibNumbers {
+		fibNumbers[i] = v * int(math.Pow(-1, float64(i%2)))
+	}
+	return fibNumbers
+}
--- a/01_fib/n0npax/main.go
+++ b/01_fib/n0npax/main.go
@@ -3,11 +3,10 @@ package main
 import (
 	"fmt"
 	"io"
-	"math"
 	"os"
 )
 
-// Allow to overwtite stream by tests
+// Allow to overwrite stream by tests
 var out io.Writer = os.Stdout
 
 // This function is useless and it's provided just due to requirements
@@ -20,6 +19,7 @@ func fib(num int) {
 	for _, v := range dispatchFib(num) {
 		fmt.Fprintln(out, v)
 	}
+
 }
 
 // Dispatch and return proper numbers
@@ -50,7 +50,7 @@ func computeFibonacci(num int) []int {
 func computeNegafibonacci(num int) []int {
 	fibNumbers := computeFibonacci(num * -1)
 	for i, v := range fibNumbers {
-		fibNumbers[i] = v * int(math.Pow(-1, float64(i%2)))
+		fibNumbers[i] = v * [2]int{1, -1}[i%2]
 	}
 	return fibNumbers
 }
--- a/01_fib/n0npax/main_test.go
+++ b/01_fib/n0npax/main_test.go
@@ -81,5 +81,8 @@ func TestFibOutputWithOne(t *testing.T) {
 	out = &buf
 
 	fib(1)
-	r.Equalf("1\n", buf.String(), "Unexpected output in fib()")
+	expected := strconv.Quote(`1
+`)
+	actual := strconv.Quote(buf.String())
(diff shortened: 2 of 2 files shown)


pfannerj (commented):
I like the fact the you created an array containing the numbers and printed the array rather than printing the numbers as the were being calculated... it separates the concerns of the functions better.

pfannerj (dismissed):
In the test cases, I think the last case (TestFibOutputWithOne) should really be written in the same way as the others with the expected and actual values defined rather than just have the values within the Equalf statement..

01_fib/n0npax/main.go
```diff
@@ -0,0 +1,56 @@
...
+		a, b = b, a+b
+	}
+	return fibNumbers
+}
+
+// Compute negafibonacci numbers
+func computeNegafibonacci(num int) []int {
+	fibNumbers := computeFibonacci(num * -1)
+	for i, v := range fibNumbers {
+		fibNumbers[i] = v * int(math.Pow(-1, float64(i%2)))
```
camscale: Since you are already using slices in this code, you could optimise this expression to remove `math.Pow` and the float/int conversions:
```
    fibNumbers[i] = v * [2]int{1, -1}[i%2]
```

i.e. index a static array instead of computing -1^0 and -1^1 each time.

Just an idea to demonstrate a different way. No need to make this change.

01_fib/n0npax/main.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+)
+
+// Allow to overwtite stream by tests
```
juliaogris: s/overwtite/overwrite

01_fib/n0npax/main.go
```diff
@@ -0,0 +1,56 @@
...
+var out io.Writer = os.Stdout
+
+// This function is useless and it's provided just due to requirements
```
juliaogris, line 12: FYI - don't change anything here but the comment `This function is useless` strikes me as strange.
The use of the function `main` in the package `main` is to be automatically run in an executable. It is the entry point of the executable.
When you use `go run main.go` it will print what has been asked to do in this lab.

01_fib/n0npax/main_test.go
```diff
@@ -0,0 +1,85 @@
+package main
```
juliaogris, line 1: Tests look pretty complete, well done!

01_fib/n0npax/main.go
```diff
@@ -0,0 +1,56 @@
...
+func main() {
+	fib(7)
+}
+
+// Wrapper function to print computed Fibonacci numbers
+func fib(num int) {
+	for _, v := range dispatchFib(num) {
```
juliaogris, line 19: If you insist on creating an array you could use a one-liner here if you want
`fmt.Fprintln(strings.Join(dispatchFib(num), "\n"))`

juliaogris (changes requested):
You need to create a [new issue](https://github.com/anz-bank/go-course/issues/new) that you reference in your PR description:
`Fixes: #<ISSUE-NUMBER>` means it will automatically be closed when the PR is merged.

I can understand that one might be offended by functions printing to a stream rather than returning a result which is then been printed.
However, IMO four functions to solve this lab seems a bit more than necessary.

You could get a pretty clean solution doing something like this:
```go
func fib(n int) {
	if n == 0 {
		// print and return
	}
	sign := // 1 for positive n ; -1 for negative n
	n1, n2 := 1, sign
	for j := 0; j < n*sign; j++ {
		fmt.Fprintln(out, n1)
		n1, n2 = n2, n1+sign*n2
	}
}
```

01_fib/n0npax/main.go
n0npax: Good hint. Thanks for this!

01_fib/n0npax/main.go
n0npax: Hi @juliaogris,
Thanks for the review.
My comment wasn't clear. Sorry about that.
I totally understand the purpose of the `main` function. I'm just thinking the `library` may be a better fit for this use-case. The `main` function doesn't consume any input - it's just hardcoded number `7`. In this case, the `main` invokes function `fib`. The `main` is required by lab description, so comment should be rather like: ```//This function is provided due to lab requirements```

01_fib/n0npax/main.go
n0npax: `strings.Join` definition looks like: `func Join(a []string, sep string) string`. https://golang.org/pkg/strings/#Join

In this case `dispatchFib(num)` returns `[]int`.

01_fib/n0npax/main_test.go
n0npax: Thanks!

n0npax:
@juliaogris due to a new issue in the GitHub.
Should I create an issue that contains just a lab name like `Lab 1 - fibonacci series` or rather `Lab 1 - fibonacci series - ${USERNAME}` ? Which approach is prefered?

01_fib/n0npax/main.go
n0npax: fixed. Thanks!

juliaogris:
> Should I create an issue that contains just a lab name like Lab 1 - fibonacci series or rather Lab 1 - fibonacci series - ${USERNAME} ? Which approach is prefered?

@n0npax - it's up you :), but it needs to be your own issue (don't reuse somebody else's) and it should be clear - maybe refer to the lab slide/README and the body? 
If you don't create your own issue somebody else might close it.

01_fib/n0npax/main.go
juliaogris: Oh! of course, apologies!

n0npax:
Ok, so it's already done. Cheers :)

juliaogris (approved):
LGTM - you solved the Lab1 🌞 

I still think you could have kept the guts of the `fib` function simpler with the following structure:

```go
func fib(n int) {
	if n == 0 { /* print 0 and return */ }
	sign := // 1 for positive n ; -1 for negative n
	n1, n2 := 1, sign
	for j := 0; j < n*sign; j++ {
		fmt.Fprintln(out, n1)
		n1, n2 = n2, n1+sign*n2
	}
}
```
I understand simplicity and its merits are subjective

PR #134, opened 2019-05-21
Lab 4 - Numeronym

### LINKED ISSUE #168: Lab 4 - Numeronym
Create an executable go program that returns a slice of numeronyms for input strings.

kasunfdo:
Fixes #168

Review of colleague's PR #265

* Implement numeronym
* Support unicode characters, improve unit tests


juliaogris (changes requested):
Please create a new issue that you reference in your PR description: Fixes #<ISSUE-NUMBER> means it will be automatically closed when the PR is merged.

Table driven tests might be a good idea here too. (https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

It's a grey zone, but IMO it's incorrect to omit non alphanumeric characters (how about languages with umlauts or accents? If you work with runes you will get results for all characters, however you could consider trimming whitespace on either end of the word.

Use some unicode characters in your tests.

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+			expected: "➳8⤆",
+		},
+	}
+
+	for name, test := range tests {
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			actual := getNumeronym(input)
+			fmt.Println(actual)
+			assert.Equalf(t, expected, actual,"Unexpected output")
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,63 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"github.com/stretchr/testify/assert"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

04_numeronym/kasunfdo/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(s ...string) []string {
+	out := make([]string, len(s))
+	for i, str := range s {
+		out[i] += getNumeronym(str)
```
camh-: The `+=` here seems unnecessary here. The slice is initialised to `len(s)` with each element being the zero value for a string - i.e. the empty string. Using `+=` on a known empty string doesn't really make much sense.

04_numeronym/kasunfdo/main.go
```diff
@@ -0,0 +1,37 @@
...
+	}
+	return out
+}
+
+func getNumeronym(str string) string {
```
camh-: Naming a function "getSomething" usually implies a "getter" function. In Go, these are named without the Get prefix (https://golang.org/doc/effective_go.html#Getters). However this isn't even a getter, so it makes sense to not use the Get prefix.

Since you have `numeronyms` as the name of the function generating multiple numeronyms, I think a more sensible name for this function that is more in line with idiomatic Go would be just `numeronym`. That makes `numeronyms` produce multiple numeronyms and `numeronym` produce one numeronym. It is a more consistent naming.

04_numeronym/kasunfdo/main.go
```diff
@@ -0,0 +1,37 @@
...
+	trimmedString := strings.TrimSpace(str)
+
+	//  Unicode string could not be accessed using index
+	//  Therefore, string is converted into an array of runes
+	runeArr := []rune(trimmedString)
```
camh-: Variable names in Go are not usually named with the type (`Arr` in this case) (https://dave.cheney.net/practical-go/presentations/qcon-china.html#_dont_name_your_variables_for_their_types). Also, this is a slice not an array so this is even misleading.

I would just call this `runes`.

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+			input:    " ➳↪▲✓₷$€₡😀⤆ ",
+			expected: "➳8⤆",
+		},
+	}
+
+	for name, test := range tests {
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			actual := getNumeronym(input)
+			assert.Equalf(t, expected, actual, "Unexpected output")
```
camh-: In the other tests you just use a Go expression and `t.Errorf`. Please keep the tests consistent - either use testify `assert` everywhere, or nowhere.

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	actual := numeronyms("testString", "www", "W3school")
+	expected := []string{"t8g", "www", "W6l"}
+	for i, v := range actual {
+		if expected[i] != v {
+			t.Errorf("Unexpected output. Expected: %q - Actual: %q", expected[i], v)
+		}
+	}
+}
+
+func TestGetNumeronyms(t *testing.T) {
```
camh-: This should really be called `TestGetNumeronym` (singular, not plural) as that is the name of the function being tested. But if you do rename the function as I suggest, don't forget to rename this test function too.

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote(`[a11y K8s abc]
```
camh-: I think when there is a single `\n` terminator on the string, this would read better as `"[a11y K8s abc]\n"`, rather than have the embedded newline in a raw string. Raw strings can be easier to read when there are multiple lines, but I think this code will flow better by keeping the indentation consistent.

04_numeronym/kasunfdo/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+		"ASCII string": {
+			input:    " Multilingualization",
+			expected: "M17n",
+		},
+		"Rune string": {
+			input:    " ➳↪▲✓₷$€₡😀⤆ ",
+			expected: "➳8⤆",
+		},
+	}
```
camh-: You should add some additional test cases for the boundary conditions - strings of length 3 and 4 to make sure the conditional logic in the function is correct. I know you have "www" in the earlier test case, but it really belongs here as this is the function with the condition on string length, so this is where the test should be.

camh- (changes requested):
A few comments about software engineering in general (testing, consistency and readability) that would pass because I only focus "Request changes" on changes related to Go itself, but since there are a couple of changes requested related to idiomatic Go, it's worth fixing them all.

04_numeronym/kasunfdo/main.go
camscale: The `+=` here seems unnecessary here. The slice is initialised to `len(s)` with each element being the zero value for a string - i.e. the empty string. Using `+=` on a known empty string doesn't really make much sense.

04_numeronym/kasunfdo/main.go
camscale: Naming a function "getSomething" usually implies a "getter" function. In Go, these are named without the Get prefix (https://golang.org/doc/effective_go.html#Getters). However this isn't even a getter, so it makes sense to not use the Get prefix.

Since you have `numeronyms` as the name of the function generating multiple numeronyms, I think a more sensible name for this function that is more in line with idiomatic Go would be just `numeronym`. That makes `numeronyms` produce multiple numeronyms and `numeronym` produce one numeronym. It is a more consistent naming.

04_numeronym/kasunfdo/main.go
camscale: Variable names in Go are not usually named with the type (`Arr` in this case) (https://dave.cheney.net/practical-go/presentations/qcon-china.html#_dont_name_your_variables_for_their_types). Also, this is a slice not an array so this is even misleading.

I would just call this `runes`.

04_numeronym/kasunfdo/main_test.go
camscale: I think when there is a single `\n` terminator on the string, this would read better as `"[a11y K8s abc]\n"`, rather than have the embedded newline in a raw string. Raw strings can be easier to read when there are multiple lines, but I think this code will flow better by keeping the indentation consistent.

04_numeronym/kasunfdo/main_test.go
camscale: This should really be called `TestGetNumeronym` (singular, not plural) as that is the name of the function being tested. But if you do rename the function as I suggest, don't forget to rename this test function too.

04_numeronym/kasunfdo/main_test.go
camscale, line 56: You should add some additional test cases for the boundary conditions - strings of length 3 and 4 to make sure the conditional logic in the function is correct. I know you have "www" in the earlier test case, but it really belongs here as this is the function with the condition on string length, so this is where the test should be.

04_numeronym/kasunfdo/main_test.go
camscale, line 62: In the other tests you just use a Go expression and `t.Errorf`. Please keep the tests consistent - either use testify `assert` everywhere, or nowhere.

camscale (changes requested):
A few comments about software engineering in general (testing, consistency and readability) that would pass because I only focus "Request changes" on changes related to Go itself, but since there are a couple of changes requested related to idiomatic Go, it's worth fixing them all.

camscale:
Sorry about the double review, Kasun. I used the wrong account the first time.

camscale (approved):
LGTM. Apologies for the delay in getting back to re-reviewing.

juliaogris (approved):
LGTM - really nice to read!

PR #135, opened 2019-05-22
Add collated labs markdown

anzboi:
Collect all labs into one markdown for easy reference.


labs.md
```diff
@@ -0,0 +1,181 @@
...
+
+Create directory `11_notify/USERNAME` containing a copy of upstream master `10_rest/USERNAME`
+Create `cmd/lostpuppy-service/main.go` running single endpoint:
+```
+POST   /api/lostpuppy/          Payload: { id: PUPPY_ID }
+This stubbed endpoint returns with 2 second delay:
+HTTP status 201 for even IDs
+HTTP status 500 for odd IDs
+```
+Update Puppy Delete method to notify lostpuppy-service in a goroutine and log response code asynchronously.
```
camscale: There is no newline at the end of the file. Every line (even the last) should be terminated with a new line.

Unfortunately, it looks (to me) that GitHub is no longer showing this - it used to with a red marker. Please configure your editor to always include a newline at the end of every line, including the last.

camscale (commented):
Can you add to the commit message how you generated this file?

The following did it for me:
```
for i in */README.md ; do cat $i; echo; echo; done > labs.md
```

This is useful as if any of the `README.md` files are updated, we can regenerate the combined `labs.md` file "automatically". We can point people at this commit to find the command.

Also, something like this should be done automatically, and any commit with auto-generated "code" should include the command to generate that code.

camscale (approved):
LGTM

PR #136, opened 2019-05-27
Fix formatting for lab README files

samuong:
A lot of the lab exercises are written with dot points, but GitHub's Markdown
engine unwraps all the lines, making it hard to read in the GitHub UI. I've
added a `-` to lines where this unwanted wrapping happens, to improve
readability, and also fixed some capitalisation errors and a broken link.

### DIFF
--- a/01_fib/README.md
+++ b/01_fib/README.md
@@ -1,13 +1,13 @@
 # Lab 1 - Fibonacci
 
-Create an executable go program in directory `01_fib/USERNAME`
-Write a function that prints the first n Fibonacci numbers
+- Create an executable go program in directory `01_fib/USERNAME`
+- Write a function that prints the first n Fibonacci numbers
 
 ```
 func fib(n int)
 ```
 
-Call fib(7) in main to print
+Call `fib(7)` in `main` to print
 
 ```
 1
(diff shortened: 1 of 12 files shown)


camscale (changes requested):
Thanks Sam. The bad formatting was due to the format originally being gopresent format with an improper conversion to markdown.

Could you also run:
```
for i in [01]*/README.md; do cat "$i"; echo; echo > labs.md; done
```
to update the `labs.md` file which is just a concatenation of these readme files? (it might be doable with a single `cat [01]*/README.md > labs.md` now that you have added a newline at the end of file, although with one fewer newlines between the files - not sure it that matters in the result).

samuong:
> Thanks Sam. The bad formatting was due to the format originally being gopresent format with an improper conversion to markdown.
> 
> Could you also run:
> 
> ```
> for i in [01]*/README.md; do cat "$i"; echo; echo > labs.md; done
> ```
> 
> to update the `labs.md` file which is just a concatenation of these readme files? (it might be doable with a single `cat [01]*/README.md > labs.md` now that you have added a newline at the end of file, although with one fewer newlines between the files - not sure it that matters in the result).

Ah I didn't notice `labs.md`, thanks for pointing it out. I've just regenerated this, but I used `for i in [01]*/README.md; do cat "$i"; echo; done > labs.md` instead.

camscale (approved):
yeah, forgot only one echo is needed now.

LGTM

PR #140, opened 2019-05-28
Fix/03 letters

eastrd:
There is a syntax error within the README.md file, where `)` within `fmt.Println(strings.Join(sortLetters(letters("aba"))), "\n")` is in the wrong place, which is causing `Join()` function to fail.

It should be `fmt.Println(strings.Join(sortLetters(letters("aba")), "\n"))` instead.


01_fib/eastrd/main.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func fib(n int) {
+	if n <= 0 {
+		fmt.Fprintln(out, "Fibonacci number has to be positive")
+		return
+	}
+	prev_prev, prev := 1, 1
```
golangcibot: don't use underscores in Go names; var `prev_prev` should be `prevPrev` (from `golint`)

PR #141, opened 2019-05-28
Lab1

### LINKED ISSUE #149: Lab 1 - fibonacci series

choohy:
Fixes #149 

Review of colleague's PR #<PR-NUMBER>

- Test Lab1


01_fib/choohy/main.go
```diff
@@ -0,0 +1,41 @@
...
+
+	fmt.Fprintln(out, "Call Fibonacci numbers!")
+
+	// fmt.Println(1)
+	// fmt.Println(1)
+	// fmt.Println(2)
+	// fmt.Println(3)
+	// fmt.Println(5)
+	// fmt.Println(8)
+	// fmt.Println(13)
```
juliaogris: Remove commented out code.

01_fib/choohy/main.go
```diff
@@ -0,0 +1,41 @@
...
+	fib(7)
+}
+
+var previous = 0
+var current = 1
+var next = current + previous
```
juliaogris: Global variables won't work if you call `fib` twice - please write a test for this then fix it. 

It's probably quite hard to do this recursively when printing numbers and not just computing them (you will double print). Instead implement a solution using a `for` loop.

01_fib/choohy/main.go
```diff
@@ -0,0 +1,41 @@
...
+
+func fib(n int) {
+
+	if n > 0 {
+		fib(n - 1)
+		//fmt.Println("n: ", n, "previous: ", previous, ", current: ", current,", next", next)
```
juliaogris: Remove commented out code.

01_fib/choohy/main_test.go
```diff
@@ -0,0 +1,22 @@
...
+	main()
+
+	expected := strconv.Quote("Call Fibonacci numbers!\n1\n1\n2\n3\n5\n8\n13\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Helper()
+		t.Errorf("Unexpected output in main()")
+	}
+}
```
juliaogris, line 78: Write more tests.
Use corner cases too (negative numbers, 0).

juliaogris (changes requested):
### Thank you for submitting your PR!

A couple high level comments:
- Fix PR description
- Fix commit messages
- Don't commit commented out code
- Don't use global variables that can be scoped local to functions
- Don't commit binaries

### PR description
Please make sure you use proper Markdown in your PR description and fill it in accordingly:
- You need to create a [new issue](https://github.com/anz-bank/go-course/issues/new) that you reference in your PR descrtiption:
`Fixes: #<ISSUE-NUMBER>` this means it will be automatically closed when the PR is merged.
- Please also fill in `Changes proposed in this PR:`

In your case that could look something like:
```
Fixes #<ISSUE-NUMBER>

Review of colleague's PR #<PR-NUMBER>

- Test Lab1
```

### Commit messages
Use `git rebase -i <COMMIT-HASH>` to rework your commits

- Separate subject from body with a blank line
- Limit the subject line to 60 characters
- Use the imperative mood in the subject line
- Do not end the subject line with a period
- Wrap the body at 80 characters
- Use the body to explain what and why vs. how


Just in terms of formatting and grammar your commits should be   
`Lab 1 completed.` => `Complete Lab 1`   
`Built fib code to binary` => `Build fib code to binary`

juliaogris:
Consider using raw strings for readability instead of `"1\n1\n2\n3\n5\n8\n13\n"`
```go
`1
1
2
3
5
8
13
`
```

01_fib/choohy/main.go
```diff
@@ -0,0 +1,45 @@
...
+	// var next = current + previous
+
+	fib(7)
+}
+
+func fib(n int) {
+	i := 0
+	previous := 0
+	current := 1
+	next := current + previous
```
golangcibot: ineffectual assignment to `next` (from `ineffassign`)

01_fib/choohy/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+	fmt.Fprintln(out, "Call Fibonacci numbers!")
+	fib(7)
+}
+
+func fib(n int) {
+	i := 0
+	previous := 0
+	current := 1
+	next := 1
```
golangcibot: ineffectual assignment to `next` (from `ineffassign`)

choohy: I don't understand.

choohy:
Fix PR description
Fix comment code
Fix corner case tests
Unable to rebase commit, help required

01_fib/choohy/main.go
juliaogris: `ineffectual` ~ has no effect.
you can declare next only inside the loop body:
```go
next := current + previous
```
and it will have the same effect.

juliaogris (changes requested):
Thanks for all the fixes.   
If you are free come just before the talk tomorrow morning around 8.45am or just afterwards and we will help with the rebasing.  
Please remove the binary.

Also, the fib function can be streamlined a bit with:
```go
	n1, n2 := 1, 1
	for i := 0; i < n; i++ {
		// print n1
		n1, n2 = n2, n1+n2
	}
```

01_fib/choohy/main.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+
+	fmt.Fprintln(out, "Call Fibonacci numbers!")
```
rokane, line 13: Little nit, but there is no need to include this `fmt.Println()` here. Please remove this and the additional line above.
Lab stated 
>Call fib(7) in main to print

So just need to call 
```go
func main() {
    fib(7)
}
```

01_fib/choohy/main.go
```diff
@@ -0,0 +1,24 @@
...
+	fib(7)
+}
+
+func fib(n int) {
+
```
rokane, line 18: Again here, little nit but I would remove the extra white space to be consistent across all functions.

01_fib/choohy/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	fib(1)
+
+	expected := strconv.Quote(`1
+`)
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Helper()
+		t.Errorf(expected)
```
rokane, line 51: You can combine these into a single call, there isn't a need to call this twice.
```go
t.Errorf("Expected: %v, Actual: %v", expected, actual)
```

01_fib/choohy/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+	var buf bytes.Buffer
+	out = &buf
+
+	fib(0)
+
+	expected := strconv.Quote(``)
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Helper()
```
rokane, line 34: I'm not really sure what this `Helper()` method does, or if it is required. From reading the documentation:
> Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped. Helper may be called simultaneously from multiple goroutines.

Does that mean that this test is skipped? If this isn't adding anything to your tests, it might be best to remove it.

01_fib/choohy/main_test.go
```diff
@@ -0,0 +1,78 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestFibNeg11(t *testing.T) {
```
rokane, line 9: Good work adding a few more test cases. One thing I would suggest though, is that these can be combined into a single Test function. You can see that the structure of each of the 3 tests below is the exact same, with the only difference being the input to the function, and the expected output. This is a good sign that some abstraction can be used to refactor these together and clean up your code. 

The way to do this would be to create a `map` of cases which you want to test against, and then range over those to achieve this.

Check out [Multiple Test Cases](https://goplay.space/#Zy2ta-z8vrO) for some inspiration and see if you can achieve something similar.

rokane (changes requested):
Looking good, but there are still a couple of things which I have pointed out throughout. If you can take a look at those that would be great. It would also be beneficial for you to include a negative test case, what is the expected behaviour / does it behave as you expect in this case. 

I can also see that the number of commits for this PR are getting longer and longer, this is a good sign that we could potentially clean these up a little bit. I am happy to sit with you and go through the whole rebasing process if that is required. Just let me know or reach out on slack and I can help :)

juliaogris (changes requested):
Please remove binary (2MB) `fib` from PR and commit history with `git rebase -i hash`

juliaogris:

PR #137, opened 2019-05-28
Lab1

brvishnu92:
Fixes #

Review of colleague's PR #

PR #145, opened 2019-05-29
Lab1  Fibonacci and bonus negafibonacci

### LINKED ISSUE #171: Lab1 Fibonacci and bonus negafibonacci
Build Fibonacci and bonus negafibonacci solution

pfannerj:
Fixes #171

Review of colleague's PR #137

- build fibonacci solution
- build bonus negafibonacci solution
- build test cases for fibonacci and negafibonacci

### DIFF
--- a/01_fib/pfannerj/main.go
+++ b/01_fib/pfannerj/main.go
@@ -11,25 +11,24 @@ var mainout io.Writer = os.Stdout
 
 func main() {
 	fmt.Fprintf(mainout, "Fibonacci series starting...\n")
-	f := 7
-	fib(f)
+	fib(7)
 	fmt.Fprintf(mainout, "Fibonacci series completed...\n")
 }
 
 func fib(n int) {
 	prev := 0
 	this := 1
-	nega := false
-	sign := false
+	nega := false //For negafibonacci - indicates whether the incoming integer value is negative
+	sign := false //Used to generate the sign output for the negafibonacci series on alternate iterations
 	if n < 0 {
 		nega = true
 		n *= -1
 	}
 	for i := 1; i <= n; i++ {
 		if nega && sign {
-			fmt.Fprintf(fibout, "-")
+			fmt.Fprintf(fibout, "-") //Print the sign for alternate iterations if nega
 		}
-		sign = !sign
+		sign = !sign //Sign only needs to be printed on alternate iterations for negafibonacci
 		fmt.Fprintln(fibout, this)
 		sum := this + prev
 		prev = this
--- a/01_fib/pfannerj/main.go
+++ b/01_fib/pfannerj/main.go
@@ -11,7 +11,7 @@ var mainout io.Writer = os.Stdout
 
 func main() {
 	fmt.Fprintf(mainout, "Fibonacci series starting...\n")
-	fib(100)
+	fib(7)
 	fmt.Fprintf(mainout, "Fibonacci series completed...\n")
 }
 
--- a/01_fib/pfannerj/main.go
+++ b/01_fib/pfannerj/main.go
@@ -11,27 +11,27 @@ var mainout io.Writer = os.Stdout
 
 func main() {
 	fmt.Fprintf(mainout, "Fibonacci series starting...\n")
-	fib(7)
+	fib(100)
 	fmt.Fprintf(mainout, "Fibonacci series completed...\n")
 }
 
-func fib(n int) {
+func fib(n int) { //NB: This caters for int values in the range -92 to 92
 	prev := 0
 	this := 1
 	nega := false //For negafibonacci - indicates whether the incoming integer value is negative
-	sign := false //Used to generate the sign output for the negafibonacci series on alternate iterations
 	if n < 0 {
 		nega = true
 		n *= -1
 	}
+	if n > 92 {
+		fmt.Fprintf(fibout, "Value outside allowable range (-92 to 92)\n")
+		return
+	}
 	for i := 1; i <= n; i++ {
-		if nega && sign {
-			fmt.Fprintf(fibout, "-") //Print the sign for alternate iterations if nega
+		if nega && (i%2 == 0) { //Negative sign needs to be printed on alternate iterations for negafibonacci
+			fmt.Fprintf(fibout, "-") //Print the negitive sign
 		}
-		sign = !sign //Sign only needs to be printed on alternate iterations for negafibonacci
 		fmt.Fprintln(fibout, this)
-		sum := this + prev
-		prev = this
-		this = sum
+		prev, this = this, this+prev
 	}
 }
--- a/01_fib/pfannerj/main.go
+++ b/01_fib/pfannerj/main.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


juliaogris:
Please add `lab1` and `Ready for Review` Labels when you are ready (you can find them in the right side bar).

01_fib/pfannerj/main.go
```diff
@@ -0,0 +1,37 @@
...
+	this := 1
+	nega := false //For negafibonacci - indicates whether the incoming integer value is negative
+	sign := false //Used to generate the sign output for the negafibonacci series on alternate iterations
+	if n < 0 {
+		nega = true
+		n *= -1
+	}
+	for i := 1; i <= n; i++ {
+		if nega && sign {
+			fmt.Fprintf(fibout, "-") //Print the sign for alternate iterations if nega
```
lexual: clever, my implementation did a bunch of math, but this is quite similar, and probably more efficient ;)

01_fib/pfannerj/main.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		sign = !sign //Sign only needs to be printed on alternate iterations for negafibonacci
```
lexual: You could consider doing modulus division by 2 as an alternate path.

Just a thought, your implementation is fine.

something like:


```
if nega && n % 2 == 0 {
}

File: 01_fib/pfannerj/main.go [hunk 3]
```diff
@@ -0,0 +1,37 @@
...
+		fmt.Fprintln(fibout, this)
+		sum := this + prev
+		prev = this
+		this = sum
+	}
```
lexual, line 36: Again, this is fine.

You can do this on one live if you wanted to:

```
prev, this = this, this + prev
```

01_fib/pfannerj/main_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+func TestFibOutputFibSeven(t *testing.T) {
+	var buf bytes.Buffer
+	fibout = &buf
+
+	fib(7)
+
+	expected := strconv.Quote("1\n1\n2\n3\n5\n8\n13\n")
```
lexual: Again this fine, but to help readability a little I did something like this:


```
        numbers := []string{
                "1",
                "1",
                "2",
                "3",
                "5",
                "8",
                "13",
        }
        expected := strings.Join(numbers, "\n") + "\n"
```

lexual:
I made some comments, but they are really just feedback.

You don't have to change any code for us to merge.

Nice way of implementing for negative n, with just printing the '-', I did the whole math with math.Pow thing ;(

01_fib/pfannerj/main.go
pfannerj: yep, definitely better... then sum isn't needed either. I think that was leftover from wip when I was printing sum (rather than this).  thanks again

01_fib/pfannerj/main.go
pfannerj: Yeah, I agree that's better then there's no need for the sign bool.  thanks...

01_fib/pfannerj/main_test.go
juliaogris: Alternatively you could use a raw string:
```go
expected := `1
1
2
3
5
8
13
`
```

01_fib/pfannerj/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+var fibout io.Writer = os.Stdout
+var mainout io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintf(mainout, "Fibonacci series starting...\n")
```
juliaogris, line 13: Nit: this output was not requested in lab descriptions.

juliaogris (changes requested):
Looks pretty close!

I've updated your PR description (see bulletins and heading for "Changes proposed in this PR"?). Please create a [new issue](https://github.com/anz-bank/go-course/issues/new) that you reference in your PR description:    
    `Fixes #<NEW-ISSUE-NUMBER>` means it will be automatically closed when the PR is merged.

Also please rebase your commit history and capitalise each commit message.

juliaogris (dismissed):
Please comment on every comment in code (either as "Done" or reason for not doing it e.g. [here](https://github.com/anz-bank/go-course/pull/145/files#r288846202)

Please update your commit message to start with capital letters with `git rebase`(we can help tomorrow morning if you have trouble doing this):
```
Build lab1 fibonacci solution and tests
Build lab1 negafibonacci solution and tests
```

01_fib/pfannerj/main.go
pfannerj: Noted... leaving it for now as a possible value-add??

01_fib/pfannerj/main.go
pfannerj: Simple, not very "mathematic" but meets the requirements.. leaving this as is

01_fib/pfannerj/main.go
pfannerj: Since I need to do a 'rebase' (as per Julia's comments) I will include this change

01_fib/pfannerj/main_test.go
pfannerj: Since I need to do a 'rebase' (as per Julia's comments) I will include this change...

01_fib/pfannerj/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	fib(0)
+
+	expected := ``
+	actual := buf.String()
+	if expected != actual {
+		t.Errorf("Unexpected output in fib(0)")
+	}
+}
+
```
camscale, line 35: It would be nice to see a test for `fib(1)`. Boundary conditions are often where bugs hide. I'd suggest that you repurpose the `fib(9)` test since that's not really testing any more than the `fib(7)` test, particularly since you have a `fib(-9)` test too.

camscale (changes requested):
This is very good. Just one suggestion in the comments for slightly different tests.

A minor nitpick, but I'll also note that your comment style is at odds with most other Go code in that comments start with slash-slash-space (`// `), rather than just slash-slash ('//`).

Also for later labs, we'll probably lean a bit harder on the fixup commits in the PR. The PR is a good place to rework your commits so when they go into master, they appear as a series of commits in final form only. For this PR, I'd be squashing the four commits into one (using `git rebase -i <commit>` and choosing `squash` or `fixup` for the last three commits). But you can leave that as-is for this PR.

camscale (approved):
Actually, I'm going to approve this now. I want to focus "changes requested" on changes related to Go, not software engineering in general. From a Go perspective this PR is great. Just take my previous comments under advisement for the next PRs.

01_fib/pfannerj/main_test.go
pfannerj: Yeah, I agree that fib(1) would be a better to include another edge case.  Also noted your observation above regarding the comment style and the commits.. I did squash some and will do further squashing for the other Labs now that I know how to do it :)

PR #148, opened 2019-05-29
Lab1

### LINKED ISSUE #257: lab01 complete for @lexual
Implement fibonacci printing for positive and negative integers.

lexual:
Implements fibonacci printing for both negative and positive n.

fix #257 

code-review #145

### DIFF
--- a/01_fib/lexual/main.go
+++ b/01_fib/lexual/main.go
@@ -3,15 +3,28 @@ package main
 import (
 	"fmt"
 	"io"
+	"math"
 	"os"
 )
 
 var out io.Writer = os.Stdout
 
+// print out n fibonacci numbers, use negafibonacci if n < 0
 func fib(n int) {
+	isPos := n > 0
+	if !isPos {
+		n *= -1
+	}
 	a, b := 0, 1
-	for i := 0; i < n; i++ {
-		fmt.Fprintln(out, b)
+	var f int
+	for i := 1; i <= n; i++ {
+		if isPos {
+			f = b
+		} else {
+			// negafibonnaci: fib(-n) = (-1)**(n+1) * fib(n)
+			f = int(math.Pow(-1, float64(i+1))) * b
+		}
+		fmt.Fprintln(out, f)
 		a, b = b, a+b
 	}
 }
--- a/01_fib/lexual/main.go
+++ b/01_fib/lexual/main.go
@@ -1,0 +1,21 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func fib(n int) {
+	a, b := 0, 1
+	for i := 0; i < n; i++ {
+		fmt.Fprintln(out, b)
+		a, b = b, a+b
+	}
+}
+
+func main() {
+	fib(7)
+}
--- a/01_fib/lexual/main.go
+++ b/01_fib/lexual/main.go
@@ -9,7 +9,7 @@ import (
 
 var out io.Writer = os.Stdout
 
-// print out n fibonacci numbers, use negafibonacci if n < 0
+// fib prints out n fibonacci numbers, use negafibonacci if n < 0
 func fib(n int) {
 	isPos := n > 0
 	if !isPos {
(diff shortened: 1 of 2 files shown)


juliaogris:
Please fix failing CI build.
Please go through the [common high level code review comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments) collected from other PRs - they might be useful to you as well.

lexual:
Here is code review I performed for someone else, for lab 1.

https://github.com/anz-bank/go-course/pull/145

01_fib/lexual/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+	numbers := []string{
+		"1",
+		"-1",
+		"2",
+		"-3",
+		"5",
+		"-8",
+		"13",
+	}
+	runTest(t, -7, numbers)
```
jessodri: Perhaps also add a test for edge cases ie. fib(0)?

01_fib/lexual/main.go
```diff
@@ -0,0 +1,34 @@
...
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// print out n fibonacci numbers, use negafibonacci if n < 0
```
rokane: I know it isn't normally required to include documentation for non exported function, but for future reference when you are including documentation for an exported method or function, it is general convention to start with the name of the function / method. Eg.
```go
// fib prints out n fibonacci numbers, use negafibonacci if n < 0
func fib(n int) {
  ...
}
```

01_fib/lexual/main_test.go
```diff
@@ -0,0 +1,72 @@
...
+	numbers := []string{
+		"1",
+		"1",
+		"2",
+		"3",
+		"5",
+		"8",
+		"13",
+	}
+	runTest(t, 7, numbers)
```
rokane: I would avoid having a helper function here to run the test. What is normal convention is to have the test comparison logic and `testing.T` method calls done inside the base test. 

I can see where you are coming from with this approach, thinking to abstract out common code into a function, but perhaps you should think in this case, maybe I could structure my test differently to avoid duplicating this code twice. This brings me to incorporating test cases. Perhaps you could combine the test cases into a data structure and iterate over that data structure in the one test.

It is idiomatic in go to have multiple cases for the one function in a single test, rather than multiple tests for each case on a function.

See [Testing Cases](https://goplay.space/#Zy2ta-z8vrO) for inspiration.

01_fib/lexual/main_test.go
rokane: Yeah can you please add a case for this as @jessodri suggested

rokane (changes requested):
Looking good so far. If you can restructure your tests and have a go at simplifying the fib logic as suggested that would be great!

01_fib/lexual/main_test.go
lexual: see latest push

lexual:
I have a new commit pushed, that responds to comments. Specifically:

* table-driven testing adopted
* better comment
* test fib(0) case.

01_fib/lexual/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+// fib prints out n fibonacci numbers, use negafibonacci if n < 0
+func fib(n int) {
+	isPos := n > 0
+	if !isPos {
+		n *= -1
+	}
+	a, b := 0, 1
+	var f int
+	for i := 1; i <= n; i++ {
```
rokane, line 20: I think I forgot to include this in my last review. But you can simplify this by assigning a default value and removing the else clause. If you do this with a rename to the `isPos` variable, it improves readability as well.

```go
for i := 1; i <= n; i++ {
  f = b
  if isNeg{
    f = int(math.Pow(-1, float64(i+1))) * b
  }
  fmt.Fprintln(out, f)
  a, b = b, a+b
}
```

rokane (approved):
Looks good to me. Can you please take a look at my comment attached for a suggestion on how to simplify. I won't request any changes, but will be handy to know for next time.

01_fib/lexual/main.go
lexual: noted

PR #161, opened 2019-05-30
Lab02

eastrd:
Fixes #

Review of colleague's PR #

### DIFF
--- a/02_bubble/eastrd/main.go
+++ b/02_bubble/eastrd/main.go
@@ -2,8 +2,8 @@ package main
 
 import (
 	"fmt"
-	"os"
 	"io"
+	"os"
 )
 
 func bubble(s []int) []int {
@@ -33,5 +33,5 @@ func bubble(s []int) []int {
 var out io.Writer = os.Stdout
 
 func main() {
-	fmt.Fprintln(out, bubble([]int{3,2,1,5}))
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
 }
--- a/02_bubble/eastrd/main.go
+++ b/02_bubble/eastrd/main.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"io"
+)
+
+func bubble(s []int) []int {
+	/*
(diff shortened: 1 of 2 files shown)


02_bubble/eastrd/main.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+func bubble(s []int) []int {
+	/*
+		Returns a sorted copy of slice s using bubble sort
```
juliaogris, line 11: Typically `go` function comments are written write above the function starting with the function name.
In you case that would be
```go
// bubble returns a sorted copy of slice s using bubble sort
func bubble(s []int) []int {
```

02_bubble/eastrd/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestBbl1(t *testing.T) {
+	expected := []int{1, 2, 3, 4, 5}
+	actual := bubble([]int{5, 4, 3, 2, 1})
+
+	if (expected == nil) != (actual == nil) || len(expected) != len(actual) {
```
juliaogris, line 13: `len(nil) == 0` so you could shorten the condition to:
`if len(expected) != len(actual)`

Use a different error message to the one below for easier debugging.

Error messages should start with lower case (see [official recommendations](https://github.com/golang/go/wiki/CodeReviewComments#error-strings)).

Last, at this stage it might be easier to use testify's [assert.Equal](https://godoc.org/github.com/stretchr/testify/assert#Equal) for slice comparison.

juliaogris (changes requested):
Please read the [collectd code review comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments):
- Fix your PR description
- Use table driven tests to avoid repetition ([slide sample](https://github.com/anz-bank/go-slides/blob/master/samples/testing/basic/cases_test.go), [Dave Cheney post](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests))

juliaogris:

PR #164, opened 2019-05-30
Implement function to print fibonacci series

### LINKED ISSUE #162: Lab 1 - fibonacci series
- [ ] Create an executable go program in directory 01_fib/CameronSchafer
```
func fib(n int)
```
```
1
1
2
3
5
8
13
```

### LINKED ISSUE #311: Print fibonacci series of any length with support for negative sequences
- [X] Simple application that prints a fibonacci series of length 7
- [X] Add exported function to print fibonacci series of any length with support for negative sequences
- [X] Add unit tests to cover various sequence lengths

nicholascross:
Fixes: #311

Review of colleague's PR #165

 - Simple application that prints a fibonacci series of length 7
 - Add exported function to print fibonacci series of any length with support for negative sequences
 - Add unit tests to cover various sequence lengths

### DIFF
--- a/01_fib/nicholascross/main.go
+++ b/01_fib/nicholascross/main.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+	if n == 0 {
+		return
+	}
+
+	prev, curr, sign := 0, 1, 1
+
+	if n < 0 {
+		sign = -1
+	}
+
+	fmt.Fprint(out, curr)
+
+	length := sign * n
+	for i := 1; i < length; i++ {
+		next := prev + curr*sign
+		prev = curr
+		curr = next
+		fmt.Fprint(out, "\n", curr)
+	}
+}
--- a/01_fib/nicholascross/main_test.go
+++ b/01_fib/nicholascross/main_test.go
@@ -1,0 +1,53 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote(format("1 1 2 3 5 8 13"))
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("%s does not match expected %s", actual, expected)
+	}
+}
+
+func TestFib(t *testing.T) {
+	cases := []struct {
+		length   int
+		sequence string
+	}{
+		{length: 7, sequence: "1 1 2 3 5 8 13"},
+		{length: -7, sequence: "1 -1 2 -3 5 -8 13"},
+		{length: 0, sequence: ""},
+		{length: 1, sequence: "1"},
+		{length: 10, sequence: "1 1 2 3 5 8 13 21 34 55"},
+	}
+
+	for _, test := range cases {
+		var buf bytes.Buffer
+		out = &buf
+		length := test.length
+		want := format(test.sequence)
+
+		t.Run(string(length), func(t *testing.T) {
+			fib(length)
+			assert.Equalf(t, buf.String(), want, "%v was incorrect", length)
+		})
+	}
+}
+
+func format(sequence string) string {
+	return strings.ReplaceAll(sequence, " ", "\n")
+}
--- a/01_fib/nicholascross/main_test.go
+++ b/01_fib/nicholascross/main_test.go
@@ -49,5 +49,5 @@ func TestFib(t *testing.T) {
 }
 
 func format(sequence string) string {
-	return strings.ReplaceAll(sequence, " ", "\n")
+	return strings.Replace(sequence, " ", "\n", -1)
 }


01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,41 @@
...
+
+	sign := sign(n)
+	sequenceLength := sign * n
+
+	lastNumber := 0
+	currentNumber := 1
+	fmt.Fprint(out, currentNumber)
+
+	for index := 0; index < sequenceLength-1; index++ {
+		nextNumber := lastNumber + currentNumber*sign
```
jonathanheaden: there is a convention within the go community to use very short variable names within functions. When you are reviewing the context of the function then a variablename like `ln` instead of `lastName` allows for faster reading whilst still allowing you to parse the whole function. I would suggest you refactor for smaller var names.

01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,41 @@
...
+		lastNumber = currentNumber
+		currentNumber = nextNumber
+		fmt.Fprint(out, " ", currentNumber)
+	}
+}
+
+func sign(n int) (sign int) {
+	if n < 0 {
```
jonathanheaden, line 22: Is a seperate function really necessary for this? Personally I don't think it is. It would be quicker to parse the `Fib` func in your head if the logic of determining sign were right there rather than off in a seperate func. I know this is a trivial implementation but sometimes it makes more sense to keep the logic for a function in that fiunction rather than splitting it out to a helper

jonathanheaden (changes requested):
Tests look good. I have suggested two changes in comments

01_fib/nicholascross/main.go
nicholascross: I am feeling compelled to dissent as this is completely counter to classic advice where variable name brevity should be avoided for the sake of clarity and readability.

If that is the convention, then I will update.

Are there any guiding principals for good variable names in go or is it just keep it short as possible?

nicholascross:
@jonathanheaden I have updated the PR with hopefully more idiomatic go style.

jonathanheaden (dismissed):
lgtm

01_fib/nicholascross/main.go
rokane: I generally follow (and I feel this is common convention within the community) a convention where variables (or consts) which are going to be long lived should have a descriptive name, as it may be referenced away from it's initial declaration in the code. Anything which is to be short lived should have a concise name, see this for further reference: [Why does go promote short names for variables?](https://www.quora.com/Why-does-Golang-promote-short-and-kind-of-meaningless-names-for-variables)

For variables which live in the context for a single loop like this, I would be updating 
`index -> i` 
`seqenceLength -> len` (len is a reserved word, so something similar... maybe `length`
`nextNumber -> next`
`lastNumber -> last` 
`currentNumber -> curr`

01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	Fib(7)
+}
+
+func Fib(n int) {
```
rokane: Since this function is only going to be used internally within the `main` package, this should be lowercased. Can you please update this.
It is also good practice to include documentation for any exported functions (capitalised functions), which will be coming up in future labs, so if you were to include some kind of exported function, you would be required to include some documentation to go with this. The way you to that is like below, where a comment starting with the function name is included directly above the function declaration:

```go
// Fib takes as input a positive or negative integer and returns 
// the fibonacci sequence up to that many digits. If the input is
// negative, the sequence returned will be negative.
func Fib(n int) {
  ...
}
```

01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
```
rokane, line 9: I would generally try to avoid declaring this as a package level variable. Can you please declare this as part of the `fib` function, since this is where the scope of this variable should live.

01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,35 @@
...
+	prev, curr, sign := 0, 1, 1
+
+	if n < 0 {
+		sign = -1
+	}
+
+	fmt.Fprint(out, curr)
+
+	length := sign * n
+	for i := 0; i < length-1; i++ {
```
rokane: I find this easier to read if the loop iterates from 1
```go
for i := 1; i < length; i ++ {
  ...
}
```

01_fib/nicholascross/main.go
```diff
@@ -0,0 +1,35 @@
...
+		next := prev + curr*sign
+		prev = curr
+		curr = next
+		fmt.Fprint(out, " ", curr)
```
rokane: Getting a little picky here, but I believe the lab had the output printing each number in the sequence on a new line, rather than a single line with spaces.
```bash
fib(7)

# Output
1
1
2
3
5
8
13
```

01_fib/nicholascross/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestFib(t *testing.T) {
+	expectFib(7, "1 1 2 3 5 8 13", t)
+	expectFib(-7, "1 -1 2 -3 5 -8 13", t)
+	expectFib(0, "", t)
+	expectFib(1, "1", t)
+	expectFib(10, "1 1 2 3 5 8 13 21 34 55", t)
+}
+
+func expectFib(sequenceLength int, sequence string, t *testing.T) {
```
rokane: Again here, I would prefer the use of `length`.

01_fib/nicholascross/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("1 1 2 3 5 8 13")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
```
rokane: Common practice to show the output vs expected, as you have done below. Please update this.

01_fib/nicholascross/main_test.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+}
+
+func TestFib(t *testing.T) {
+	expectFib(7, "1 1 2 3 5 8 13", t)
```
rokane: Since you are testing over a range of different cases, you can do something similar to what is shown here [Testing Range Cases](https://goplay.space/#Zy2ta-z8vrO)

```go
func TestIsEven(t *testing.T) {
	cases := []struct {
		length   int
		sequence string
	}{
		{length: 7, sequence: "1 1 2 3 5 8 13"},
		{length: -7, sequence: "1 -1 2 -3 5 -8 13"},
		{length: 0, sequence: ""},
		{length: 1, sequence: "1"},
		{length: 10, sequence: "1 1 2 3 5 8 13 21 34 55"},
	}

	for _, c := range cases {
		expectFib(c.length, c.sequence, t)
	}
}
```

rokane (dismissed):
Looking good. There are a couple of things throughout which need addressing. As well as that, the commit message could be tidied up as well, I like the title using an imperitive tone, however I think there is no need to add the lab 1 section. 
`[162] Lab1 - Implement function to print fibonacci series` update this to `Implement function to print fibonacci series`

There is also a line in the body of the commit message which is longer than 80 chars and doesn't wrap nicely. This should be shortened, you can see it when looking at the `Commits` tab in the Pull Request viewer.

01_fib/nicholascross/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+		t.Run(string(length), func(t *testing.T) {
+			fib(length)
+			assert.Equalf(t, buf.String(), want, "%v was incorrect", length)
+		})
+	}
+}
+
+func format(sequence string) string {
+	return strings.ReplaceAll(sequence, " ", "\n")
```
golangcibot: ReplaceAll not declared by package strings (from `typecheck`)

nicholascross: @rokane @jonathanheaden Do you guys understand why I am getting this CI failure, it works locally 😕

01_fib/nicholascross/main.go
nicholascross: I agree, it is potentially dangerous if accessed concurrently.  I copied this from lab0.  Would it be okay to pass it in as a function parameter?

01_fib/nicholascross/main_test.go
nicholascross: This might actually be an issue with the golangci-lint version used on CI.  Cam and Julia are investigating.

nicholascross: Will retry after https://github.com/anz-bank/go-course/pull/199 is merged into master

01_fib/nicholascross/main.go
nicholascross: @rokane I don't see how to do this without changing the function signature stipulated for this lab.  Is the alternative approach you are suggesting to add an extra parameter?

01_fib/nicholascross/main_test.go
nicholascross: @camh-anz based on your [comment](https://github.com/anz-bank/go-course/pull/199#issuecomment-498070838) I will ignore this lint issue.

01_fib/nicholascross/main.go
nicholascross: Sorry I just realised the above comment might not have all context. "do this" 😆 .

The reason this declared outside of the `fib` function is so it can be modified to allow testing of the function print output.  A bit unconventional but this seems to be the approach recommended for purposes of this course.

nicholascross: This is the accepted approach on several merged PRs.  I am going to resolve conversation for this issue as "Won't Fix".  So hopefully it can be merged as well 😄

juliaogris: @rokane - necessary for this kind of lab setup (executable printing to stdout & 100% coverage)

juliaogris (approved):
LGTM
personally i would have inlined `length` as in `for i := 1; i < n*sign; i++`
and not bothered with the `format` function in tests but maybe used a slice of int and concatenated it with `strings.Join(ints, "\n")` but this is really a matter of taste, the lab is solved. 
Also, you were meant to create an issue that you are auto-resolving with PR merging (see [Collected Code Review Comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments) ) - please remember for next time.

nicholascross:
@juliaogris Oh I will have to do this retrospectively I have 3 other PRs open without associated issues.

01_fib/nicholascross/main.go
rokane: @nicholascross apologies on the late reply. I hadn't checked the other PR's yet, yours was my first. All good :)

PR #133, opened 2019-05-21
Lab 3 - Letter frequency

kasunfdo:
Fixes #167

Review of colleague's PR #201

#### Changes proposed in this PR:
* Implement letter frequency

PR #165, opened 2019-05-30
Lab1

### LINKED ISSUE #162: Lab 1 - fibonacci series

CameronSchafer:
Fixes #162

Review of colleague's PR #133

- Implement Lab1 NegaFibonacci numbers
- Test Lab1

### DIFF
--- a/01_fib/CameronSchafer/fib.go
+++ b/01_fib/CameronSchafer/fib.go
@@ -15,8 +15,10 @@ func fib(n int) {
 	if n < 0 {
 		sign = -1
 	}
+
+	n *= sign
 	//loop until all numbers have been printed
-	for i := 0; i < n*sign; i++ {
+	for i := 0; i < n; i++ {
 		if i%2 == 0 || sign > 0 {
 			//will get the negafib number if the sign is -1
 			n1, n2 = n2, n1+n2*sign
--- a/01_fib/CameronSchafer/fib.go
+++ b/01_fib/CameronSchafer/fib.go
@@ -10,7 +10,22 @@ var out io.Writer = os.Stdout
 
 //base fibonacci function
 func fib(n int) {
-	fmt.Fprintln(out, n)
+	sign, n1, n2 := 1, 0, 1
+	//check if use nega or normal fib sequence
+	if n < 0 {
+		sign = -1
+	}
+	//loop until all numbers have been printed
+	for i := 0; i < n*sign; i++ {
+		if i%2 == 0 || sign > 0 {
+			//will get the negafib number if the sign is -1
+			n1, n2 = n2, n1+n2*sign
+		} else {
+			//will account for the negative number during the negafib sequence
+			n1, n2 = n2, n1-n2
+		}
+		fmt.Fprintln(out, n1)
+	}
 }
 
 func main() {
--- a/01_fib/CameronSchafer/fib.go
+++ b/01_fib/CameronSchafer/fib.go
@@ -1,0 +1,18 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+//base fibonacci function
+func fib(n int) {
+	fmt.Fprintln(out, n)
+}
+
+func main() {
+	fib(7)
+}
--- a/01_fib/CameronSchafer/fib_test.go
+++ b/01_fib/CameronSchafer/fib_test.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+//main() test
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("7\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()\nexpected: %q\nactual: %q", expected, actual)
+	}
+}
+
+func TestFibOutput(t *testing.T) {
+	//tempBuf is used to reset the output buffer for each test
+	var buf, tempBuf bytes.Buffer
+	out = &buf
+
+	//test cases with descriptions.
+	testCases := []struct {
+		description string
+		input       int
+		expected    string
+	}{
+		{description: "fib 7", input: 7,
+			expected: strconv.Quote("7\n"),
+		},
+		{description: "fib 1", input: 1,
+			expected: strconv.Quote("1\n"),
+		},
+		{description: "fib -7", input: -7,
+			expected: strconv.Quote("-7\n"),
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
+		// t.Run creates a sub test and runs it like a normal test
+		t.Run(test.description, func(t *testing.T) {
+			fib(test.input)
+			result := strconv.Quote(buf.String())
+			if result != test.expected {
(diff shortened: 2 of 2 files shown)


01_fib/CameronSchafer/fib.go
```diff
@@ -0,0 +1,33 @@
...
+
+//base fibonacci function
+func fib(n int) {
+	sign, n1, n2 := 1, 0, 1
+	//check if use nega or normal fib sequence
+	if n < 0 {
+		sign = -1
+	}
+	//loop until all numbers have been printed
+	for i := 0; i < n*sign; i++ {
```
nicholascross: We could potentially optimise this by removing the multiplication everytime the the condition is evaluated.  Calculating the sequence length once in advance given the value `n*sign` does not change during iteration.

01_fib/CameronSchafer/fib.go
```diff
@@ -0,0 +1,33 @@
...
+		if i%2 == 0 || sign > 0 {
+			//will get the negafib number if the sign is -1
+			n1, n2 = n2, n1+n2*sign
```
nicholascross, line 24: For my own benefit, can you tell me does this evaluate everything on the right of `=` prior to assignment?  I guess it must.

01_fib/CameronSchafer/fib_test.go
```diff
@@ -0,0 +1,79 @@
...
+-3
+5
+-8
+13
+`),
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
```
nicholascross, line 68: Why do we need to do this?

01_fib/CameronSchafer/fib.go
nicholascross: Thanks I found it does do that, this has helped me simplify my second lab 👍

nicholascross:
This looks good to me the tests are passing and seem to cover most interesting edge cases except maybe `0`

I had a suggestion and question for the author but no required changes.

01_fib/CameronSchafer/fib.go
CameronSchafer: Yes it does. This assignment is shorthand for
```
n1 = n2
n2 = n1 + (n2 * sign)
```

01_fib/CameronSchafer/fib_test.go
CameronSchafer: Without this assignment, test was not accessible within the for loop.

01_fib/CameronSchafer/fib.go
CameronSchafer: Thank you, I will try change this up now.

CameronSchafer: @nicholascross I updated the code as per your recommendation.

01_fib/CameronSchafer/fib.go
nicholascross: From my experiments it is more closely equivalent to

```
x1 := n1
x2 := n2

n1 = x2 
n2 = x1 + (x2 * sign)
```
So the values on the right hand side are always the original values before any assignment occurs.

So you can swap variable values like this

```
n1, n2 = n2, n1
```

You can see if it translated as seperate statements you would end up with the same value in `n1` and `n2`

01_fib/CameronSchafer/fib_test.go
nicholascross: Is it something to do with the anonymous function/closure?

CameronSchafer: To be honest, I'm not exactly sure. I think it has something to do with the way I was trying to access the struct array. I will look into it some more and get back to you.

CameronSchafer: It is a linting error that pops up if I don't do that assignment:
`Using the variable on range scope 'test' in function literal (scopelint)`

So I feel it is safe to just do the assignment to remove the error.

01_fib/CameronSchafer/fib.go
CameronSchafer: Yeah that makes sense, it is a really useful feature of Go.

01_fib/CameronSchafer/fib_test.go
nicholascross: @CameronSchafer I think I found the [answer](https://stackoverflow.com/questions/26692844/captured-closure-for-loop-variable-in-go)

When your function literal captures the `test` variable it is capturing a `reference to the value` not the `value itself`.  So if the functions are executed outside of the for loop then they will have the last value of the range not the value at the time of iteration.

So you have done the right thing here 👍

CameronSchafer: @nicholascross Oh Ok. That makes sense now. Thank you

01_fib/CameronSchafer/fib.go
```diff
@@ -0,0 +1,35 @@
...
+
+//base fibonacci function
+func fib(n int) {
+	sign, n1, n2 := 1, 0, 1
+	//check if use nega or normal fib sequence
+	if n < 0 {
+		sign = -1
+	}
+
+	n *= sign
```
camscale, line 19: minor suggestion: you could move this inside the `n < 0` block above since it only needs to be done for negative `n`.

01_fib/CameronSchafer/fib.go
```diff
@@ -0,0 +1,35 @@
...
+	//loop until all numbers have been printed
+	for i := 0; i < n; i++ {
+		if i%2 == 0 || sign > 0 {
+			//will get the negafib number if the sign is -1
+			n1, n2 = n2, n1+n2*sign
+		} else {
+			//will account for the negative number during the negafib sequence
+			n1, n2 = n2, n1-n2
```
camscale, line 27: Since this branch of the `if` statement is only taken when `sign == -1`, and `n1+n2*-1` is the same as `n1-n2`, the two branches of the `if` statement are functionally identical. You could remove the `if` statement entirely and just leave the first branch.

01_fib/CameronSchafer/fib.go
```diff
@@ -0,0 +1,35 @@
...
+func fib(n int) {
+	sign, n1, n2 := 1, 0, 1
+	//check if use nega or normal fib sequence
+	if n < 0 {
+		sign = -1
+	}
+
+	n *= sign
+	//loop until all numbers have been printed
+	for i := 0; i < n; i++ {
```
camscale, line 21: minor suggestion: count from 1 to n, instead of 0 to n-1 (i.e. `for i := 1; i <= n; i++`) because we are computing the nth fibonacci number, not the nth-1 fibonacci number.

01_fib/CameronSchafer/fib_test.go
```diff
@@ -0,0 +1,79 @@
...
+			expected: strconv.Quote(`1
+-1
+2
+-3
+5
+-8
+13
+`),
+		},
+	}
```
camscale, line 65: I would add a `fib(0)` test case too as it is on the boundary.

camscale (approved):
I'm approving this with suggestions as the suggestions all relate to software engineering rather than Go specifically. From the perspective of Go, this code is great and shows some good idiomatic use of Go.

A minor nitpick, but I'll also note that your comment style is at odds with most other Go code in that comments start with slash-slash-space (`// `), rather than just slash-slash (`//`).

PR #169, opened 2019-05-30
01 fib

### LINKED ISSUE #221: Lab01 Fibonacci
**Lab 1 - Fibonacci sequence**
- [ ] Create an executable go program in directory 01_fib

jonathanheaden:
Fixes #221

Implement Fib func
Add tests for fib(7) and fib(-7)

### DIFF
--- a/01_fib/jonathanheaden/main.go
+++ b/01_fib/jonathanheaden/main.go
@@ -8,30 +8,32 @@ import (
 
 var out io.Writer = os.Stdout
 
-func fib(n int) string {
-	f := [3]int{0, 1, 1}
-	var prefix, res string
-	var j int
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+	a, b := 0, 1
+	var prefix string
 	var neg = false
 	if n < 0 {
 		n *= -1
 		neg = true
 	}
-	for i := 0; i < n; i++ {
-		if j = i + 1; i > 1 {
-			f = [3]int{f[1], f[2], (f[1] + f[2])}
-			j = 2
-		}
-		if prefix = ""; (i%2 == 1) && neg {
+	for i := 0; i <= n; i++ {
+		if prefix = ""; neg && (i%2 == 0) {
 			prefix = "-"
 		}
-		res = fmt.Sprintf("%s\n%s%d", res, prefix, f[j])
+		switch i {
+		case 0:
+			if n == 0 {
+				fmt.Fprintln(out, "0")
+			}
+		case 1:
+			fmt.Fprintf(out, "%s%d\n", prefix, 1)
+		default:
+			a, b = b, a+b
+			fmt.Fprintf(out, "%s%d\n", prefix, b)
+		}
 	}
-	return res
-}
-
-func main() {
-
-	fmt.Fprintln(out, fib(7))
-
 }
--- a/01_fib/jonathanheaden/main.go
+++ b/01_fib/jonathanheaden/main.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func fib(n int) string {
+	f := [3]int{0, 1, 1}
+	var prefix, res string
+	var j int
+	var neg = false
+	if n < 0 {
+		n *= -1
+		neg = true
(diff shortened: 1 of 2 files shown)


01_fib/jonathanheaden/main.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+var fibs = map[int]int{}
```
camscale: a map where each index is a sequential number from 0 is really the same data structure as an array/slice, just less efficient.

Also do you really need to store *all* the fibs, or just the last two? Can you get away with out an indexed data structure at all?

01_fib/jonathanheaden/main.go
```diff
@@ -0,0 +1,34 @@
...
+	prefix := ""
+	var res string
+	if n < 0 {
+		n *= -1
+		prefix = "-"
+	}
+	fibs[0] = 1
+	fibs[1] = 1
+	for i := 0; i < n; i++ {
+		if i > 1 {
```
camh-anz: Rather than have this if statement, can you just start the loop at 2?

camh-anz: maybe not, since you need to print the earlier numbers...

01_fib/jonathanheaden/main_test.go
```diff
@@ -0,0 +1,24 @@
...
+	// Given
+	r := require.New(t)
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote("\n1\n1\n2\n3\n5\n8\n13\n\n-1\n-1\n-2\n-3\n-5\n-8\n-13\n")
```
camscale: The output of the negative fib series is not correct. Double-check the wikipedia page. The sign alternates.

01_fib/jonathanheaden/main.go
jonathanheaden: agree. I have replaced the map with a three element array

01_fib/jonathanheaden/main_test.go
jonathanheaden: I had to think through this a lot, I see what you mean. I have altered the test and code

juliaogris (dismissed):
Please have a read through the [collected lab 1 review comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments):
- Create issue and add link to PR description
- Fix up markdown in Pr
- Consider using table driven tests
- Fold your 8 commits into one or two with `git rebase`

Your solution could be simplified to something like:

```go
	sign := // 1 for positive n ; -1 for negative n
	n1, n2 := 0, 1
	for j := 0; j < n*sign; j++ {
		n1, n2 = n2, n1+sign*n2
		// print n1
	}
```

01_fib/jonathanheaden/main.go
```diff
@@ -0,0 +1,39 @@
...
+	var neg = false
+	if n < 0 {
+		n *= -1
+		neg = true
+	}
+	for i := 0; i <= n; i++ {
+		if prefix = ""; neg && (i%2 == 0) {
+			prefix = "-"
+		}
+		switch i {
```
juliaogris, line 27: No need to have a switch on loop index inside look - might as well do it before imo.

camscale (approved):
I happy with the Go code in this PR. You could implement the loop more simply along the lines of Julia's earlier suggestion, and a `fib(1)` test case would be nice to see as well, since you have a `case 1` in the code, but these are more software engineering issues and we want to focus on Go coding for these lab PRs.

juliaogris (approved):
The lab is solved!

You have done everything that's required.
I still think when you calculate fib you care about the right values as integer and not just them being printed correctly, so the prefix is a bid odd to me.
To me a solution like the listing below would have been a bit cleaner:
```go
func fib(n int) {
	if n == 0 {
		return
	}

	sign := 1
	if n < 0 {
		sign = -1
	}

	n1, n2 := 0, 1
	for j := 0; j < n*sign; j++ {
		n1, n2 = n2, n1+sign*n2
		fmt.Fprintln(out, n1)
	}
}
```

PR #179, opened 2019-05-30
Lab 1 - Implement Fibonaci and NegaFibinacci

marksomething:
Fixes #240 

Review of colleague's PR #183

#### Changes proposed in this PR:
* Implement recursion based function to return a slice of
  Fibonacci numbers starting at 0
* Implement functions to present fibonacci as a string / stdout

PR #155, opened 2019-05-29
create a fibonacci number generator function

hsy3418:
Fixes #392

Review of colleague's PR #246

PR #183, opened 2019-05-30
Complete Lab 1 with bonus Fibonacci and  Negafibonacci number generation

### LINKED ISSUE #245: Changes done for Lab1
#183

shampa077:
Fixes #245 

Review of colleague's PR #155, #179 

- Test Lab1


01_fib/shahriys/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+2
+-3
+5
+-8
+13
+`
+	actual := buffib.String()
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
```
marksomething: It can be helpful to print the actual and expected when a test fails. It makes it actionable.

marksomething (dismissed):
Nice work!

marksomething:
There are performance issues relating to iterating into a recursive function but not critical in this case as int64 will overflow before it is a real performance impact.

01_fib/shahriys/main_test.go
shampa077: yap good suggestion. I actually added while checking. Later removed it

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var outfib io.Writer = os.Stdout
+
+// Function to calculate nth Fibonacci or Negafibonacci number
+func nfib(n int) int64 {
```
Jake88: I like the separation of concerns and creating a recursive function to work out Fn. Nice work.

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,51 @@
...
+			return nfib(n+2) - nfib(n+1)
+		}
+	}
+}
+
+// Function to printh first n Fibonacci or Negafibonacci numbers
+func fib(n int) {
+	if n >= 0 {
+		for i := 0; i < n; i++ {
+			fmt.Fprintln(outfib, nfib(i))
```
Jake88: Looks like there's a performance hit here with the for loop into recursion. Potentially you could create a look up / map for Fn numbers that have already been calculated to reduce the number of iterations nfib would run, but that's at the cost of storing a map, so it's give and take. 👍

01_fib/shahriys/main_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	fib(7)
+
+	expected := `1
+1
+2
+3
+5
+8
+13
```
Jake88, line 46: I found when writing my expected result I was able to write 
```
expected := `1\n1\n2\n...`
```
rather than having my numbers hard against the left wall. Personal preference, but noting it here incase it influences your future tests. :)

Jake88 (dismissed):
Sorry for hi-jacking this PR after it's already been reviewed, there's not enough awaiting PRs at the moment!

Just a couple of opinionated comments, nothing major.

Overall nice work. Looks like you've got a good grasp on Go Lang already! :)

01_fib/shahriys/main.go
anzboi: Cost of creating/storing a map and assignment/lookup is `O(1)`. The current recursive implementation is `O(2^n)`, then on top of that we do the calculation for each value from `0` to `n`, meaning the overall complexity is `O(n2^n)`. The map would eliminate the extra factor of `2^n`, so definitely worth it.
See my other review comments though, since you don't need to do this to achieve `O(n)`.

01_fib/shahriys/main.go
anzboi: While the separation of concern is usually a good thing, here it causes the performance hit, especially with this recursive implementation. Naive recursion runs in `O(2^n)` time, whereas iterative would be in `O(n)`. I would recommend going with the iterative solution at least. You can do it recursively in `O(n)` with tail recursion.
You can also print the results as you go if you don't want to have to recalculate each fibonacci number, but I like the separation of concerns as well, since this is more likely how it would look in practice.

01_fib/shahriys/main_test.go
```diff
@@ -0,0 +1,73 @@
+package main
```
anzboi, line 1: Can we add a test for `fib(0)`.

01_fib/shahriys/main_test.go
anzboi: That's odd @Jake88, are you sure you don't mean to use `"` instead of backticks? ([see here](https://goplay.space/#drRsTZ-HVfK)). If you are testing with backticks, then the output of your function is wrong (each number is sparated by a literal `\n` instead of a newline).

As for whether to test agains quotes with `\n` or backticks with newlines, I think this comes down to preference, I personally prefer a string literal (backticks, what you currently have) as it actually looks like the output. I also don't find it too hard to see a bunch of `\n`s in a row in quotes, and know what the output should look like. I'll leave it up to you if you want to change it.

01_fib/shahriys/main_test.go
anzboi: Can you re-add it?

anzboi (changes requested):
Getting there,
I would like you to rethink the recursive implementation (see my comments). While it is not a requirement to achieve optimal complexity, I would like you to give it a shot.
I would also like one extra test for `fib(0)`.
Can you also add a little more detail to the PR message, briefly mention the problem (normally you would reference an issue, but a little hard in this repo), and describe your solution.

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,72 @@
...
+			return -1
+		default:
+			return nfib(n+2) - nfib(n+1)
+		}
+	}
+}
+
+// Function to printh first n Fibonacci or Negafibonacci numbers
+func fib(n int) {
+	if n > 0 {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

01_fib/shahriys/main_test.go
shampa077: added

01_fib/shahriys/main.go
shampa077: Recursion replaced with loop

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,51 @@
...
+func fib(n int) {
+	switch n {
+	case 0:
+		fmt.Fprintln(outfib, 1)
+	default:
+		if n > 0 {
+			n1, n2 := 1, 1
+			for i := 0; i < n; i++ {
+				switch i {
+				case 0, 1:
```
anzboi, line 21: You can actually remove this switch statement by changing the code in the `default` block slightly

01_fib/shahriys/main_test.go
anzboi: 👍

anzboi (changes requested):
Great work, this is good to go now.
However, your branch deletes @kasunfdo 's Lab1. Can you please fix your branch so that it is not deleted.
Ive added a comment for you to think about as well.

01_fib/shahriys/main.go
shampa077: branch Fixed @anzboi

juliaogris (changes requested):
Please read through the [Collected Code Review Comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments)

- [Fix your pr description](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#pr-description)
- [Limit the number of commits to one or two](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#limit-the-number-of-commits-to-one-or-two)
- Undo the removal of `01_fib/kasunfdo`

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,51 @@
...
+					fmt.Fprintln(outfib, -1)
+				default:
+					fmt.Fprintln(outfib, n1-n2)
+					n2, n1 = n1-n2, n2
+				}
+
+			}
+		}
+	}
+
```
juliaogris, line 45: Unnecessary blank line.

juliaogris (approved):
Your solution is functionally correct but it is a lot more verbose than needed.
Readability is the holy grail of writing code IMO and usually less code is easier to grok than more (though one can clearly go overboard). 
Here is our reference solution - feel free to use, adjust or ignore:
```go
func fib(n int) {
	sign := 1
	if n < 0 {
		sign = -1
	}

	n1, n2 := 0, 1
	for j := 0; j < n*sign; j++ {
		n1, n2 = n2, n1+sign*n2
		fmt.Fprintln(out, n1)
	}
}
```

anzdaddy:
Please close merge or close.

PR #187, opened 2019-05-31
Lab2 - BubbleSort

marksomething:
Fixes #

Review of colleague's PR # todo

- New function bubble([]int) []int returns sorted copy of passed-in slice.
- Test cases to cover positive values, negative values, empty and single element slices.


marksomething:
Will Implement insertion sort & qsort

PR #191, opened 2019-05-31
Lab3

### LINKED ISSUE #288: Undo the removal of 01_fib/kasunfdo and test cases split

shampa077:
Fixes #288 

Review of colleague's PR #

-- Complete Lab 3
-- Seperate Test cases for Letters & SortLetters


03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+// As rune type can't be sorted so creating new type
+type RuneSlice []rune
+
+func (p RuneSlice) Len() int           { return len(p) }
+func (p RuneSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p RuneSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func sortLetters(m map[rune]int) []string {
+
+	var keys []rune
```
golangcibot: Consider preallocating `keys` (from `prealloc`)

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+	for key := range m {
+		keys = append(keys, key)
+	}
+	var runes RuneSlice = keys
+	sort.Sort(runes)
+	var returnval []string
```
golangcibot: Consider preallocating `returnval` (from `prealloc`)

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	type test struct {
+		input    string
+		expected []string
+		actual   []string
+	}
+
+	tests := []test{
+		{input: "abac", expected: []string{"a:2", "b:1", "c:1"}, actual: sortLetters(letters("abac"))},
+		{input: "aAdbbDA", expected: []string{"A:2", "D:1", "a:1", "b:2", "d:1"}, actual: sortLetters(letters("aAdbbDA"))},
+		{input: "aAAADDDbacZZZ", expected: []string{"A:3", "D:3", "Z:3", "a:2", "b:1", "c:1"}, actual: sortLetters(letters("aAAADDDbacZZZ"))},
```
golangcibot: line is 136 characters (from `lll`)

03_letters/shahriys/main.go
shampa077: addrees in new commit @feb5723bdc427e41521295be96f75bd5c2a77c98

03_letters/shahriys/main.go
shampa077: addrees in new commit @feb5723bdc427e41521295be96f75bd5c2a77c98

03_letters/shahriys/main_test.go
shampa077: addrees in new commit @feb5723bdc427e41521295be96f75bd5c2a77c98

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+	var keys = make([]rune, 0, 256)
+	for key := range m {
+		keys = append(keys, key)
+	}
+	var runes RuneSlice = keys
+	sort.Sort(runes)
+	var returnval = make([]string, 0, 256)
+
+	for _, k := range runes {
+		s := string(k) + ":" + strconv.FormatInt(int64(m[k]), 10)
```
CameronSchafer: For concatenating strings together you could use something like:
```
s := fmt.Sprintf("%c:%d", k, m[k])
```
This will format your string for you. I think it is more succinct than doing type conversions and using pluses. :)

CameronSchafer (commented):
The code looks good! I just commented on this part as I believe it could be improved.

anzboi:
@CameronSchafer Have you finished your review? It looks like you finished reviewing and added comments but it is still labelled `under review`.

CameronSchafer:
> @CameronSchafer Have you finished your review? It looks like you finished reviewing and added comments but it is still labelled `under review`.

@anzboi Oh yes thank you for the reminder. I will change the label now.

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+			alphamap[rune(s[i])] = 1
+		}
+
+	}
+	//fmt.Println(alphamap)
+	return alphamap
+}
+
+// As rune type can't be sorted so creating new type
+type RuneSlice []rune
```
anzboi, line 24: 👍

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func (p RuneSlice) Len() int           { return len(p) }
+func (p RuneSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p RuneSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func sortLetters(m map[rune]int) []string {
+
+	var keys = make([]rune, 0, 256)
```
anzboi: You don't need to allocate capacity for each possible rune value. You know there are only `len(m)` different runes in your map.

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+var outletter io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+
+	alphamap := make(map[rune]int)
+
+	for i := 0; i < len(s); i++ {
+		if val, ok := alphamap[rune(s[i])]; ok {
```
anzboi: `alphamap[key] += 1` will work even if `key` does not yet exist in that map ([example](https://goplay.space/#bSn7U4A9ZMc)). You can use this to get rid of the `if` block entirely.
👍 on using assignment in the `if` block though

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+			alphamap[rune(s[i])] = val + 1
+		} else {
+			alphamap[rune(s[i])] = 1
+		}
+
+	}
+	//fmt.Println(alphamap)
```
anzboi: Remove commented code.

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func sortLetters(m map[rune]int) []string {
+
+	var keys = make([]rune, 0, 256)
+	for key := range m {
+		keys = append(keys, key)
+	}
+	var runes RuneSlice = keys
+	sort.Sort(runes)
+	var returnval = make([]string, 0, 256)
```
anzboi: Similar here, you don't need the full 256 capacity.

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+			actual: sortLetters(letters("aAAADDDbac"))},
+		{input: "", expected: []string{}, actual: sortLetters(letters(""))},
+	}
+
+	for _, tc := range tests {
+		if len(tc.expected) > 0 && len(tc.actual) > 0 {
+			if !reflect.DeepEqual(tc.expected, tc.actual) {
+				fmt.Printf("%v  %T", tc.expected, tc.expected)
+				fmt.Printf("%v  %T", tc.actual, tc.actual)
+				t.Fatalf(" input:%v, expected: %v, got: %v", tc.input, tc.expected, tc.actual)
```
anzboi: Use `Errorf` instead of `Fatalf`. This will tell you which cases pass and which fail. With `Fatalf`, you only get to see up to the first case that fails.
You can also use a `map[string]test` and `t.Run` to run each case as a separate named sub-test, see [here](https://goplay.space/#Zy2ta-z8vrO) for an example.

anzboi (changes requested):
Looking really good, two things
1. you don't need the full `256` capacity in your `sortLetters` function. You can get exactly how big the array needs to be from the map
2. Your test cases are good, but the test uses `Fatalf` instead of `Errorf`. This causes the test to exit on the first failure, rather than telling you all the failing tests.
I also provided a hint as to how you can remove the `if` block in `letters`.

03_letters/shahriys/main.go
shampa077: addressed

03_letters/shahriys/main.go
shampa077: addressed

03_letters/shahriys/main.go
shampa077: incorporated

03_letters/shahriys/main.go
shampa077: addressed

03_letters/shahriys/main.go
shampa077: addressed

03_letters/shahriys/main_test.go
shampa077: addressed

anzboi (dismissed):
This is just about good to go. However this PR deletes @kasunfdo 's lab1. Can you fix your branch so that their work is not modified.
It also looks like one comment was not addressed, not a big one but may as well do while fixing the branch

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,49 @@
...
+	var runes RuneSlice = keys
+	sort.Sort(runes)
+	var returnval = make([]string, 0, len(m))
+
+	for _, k := range runes {
+		s := fmt.Sprintf("%c:%d", k, m[k])
+		returnval = append(returnval, s)
+	}
+	return returnval
+}
```
rokane, line 45: Little nit pick. Insert a new line here between the two functions

03_letters/shahriys/main.go
```diff
@@ -0,0 +1,49 @@
...
+func (p RuneSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p RuneSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func sortLetters(m map[rune]int) []string {
+
+	var keys = make([]rune, 0, len(m))
+	for key := range m {
+		keys = append(keys, key)
+	}
+	var runes RuneSlice = keys
```
rokane, line 36: You can avoid having to create a custom type which implements the Sortable interface if you avoid creating a `[]rune` from the map. Is there a way in which you could directly make a `[]string`? You are doing it below... Are you able to generate that from the map directly, without the step inbetween?

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+	main()
+	expected := `a:2
+b:1`
+	actual := bufletter.String()
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()" + expected + " " + actual)
+	}
+
+}
```
rokane, line 21: Insert a line between these functions

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	main()
+	expected := `a:2
+b:1`
+	actual := bufletter.String()
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()" + expected + " " + actual)
+	}
+
```
rokane: Remove the extra new line here

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+}
+func TestSortLetters(t *testing.T) {
```
rokane: I think it is best to test each of these two functions in isolation, rather than testing the output of concatenating the two function calls together. Can you please break this down to:

```go
func TestLetters(t *testing.T) {
  ....
}

func TestSortLetters(t *testing.T) {
  ....
}
```

03_letters/shahriys/main_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+	tests := []test{
+		{input: "abac", expected: []string{"a:2", "b:1", "c:1"}, actual: sortLetters(letters("abac"))},
+		{input: "aAdbbDA", expected: []string{"A:2", "D:1", "a:1", "b:2", "d:1"}, actual: sortLetters(letters("aAdbbDA"))},
+		{input: "aAAADDDbac", expected: []string{"A:3", "D:3", "a:2", "b:1", "c:1"},
+			actual: sortLetters(letters("aAAADDDbac"))},
+		{input: "", expected: []string{}, actual: sortLetters(letters(""))},
+	}
+
+	for _, tc := range tests {
```
rokane: I think you are on the right track here with creating a range of tests, but this can probably be cleaned up a little bit. Take a look at [Multiple Cases](https://goplay.space/#Zy2ta-z8vrO) for inspiration. A couple of points here:
 - I would generally avoid storing the `actual` value as a part of the test struct, there is no need to store it, rather generate it as the first step of the loop over the testCases.
 - I don't see the need to do the two `fmt.Printf()` calls when you are printing the expected and output as a part of the `t.Errorf()` call, please remove these.
 - If you use a `map[string]struct` you can use the key as a test name and easily structure your tests according to different cases, as seen in the example.

```go
cases := map[string]struct{input string, expected []string}{ "Test Name": {"input", {"expected"}} }

    if actual := sortLetters(test.input); actual != test.expected {
      t.Errorf("error msg here")
    } 
  })
}

rokane (changes requested):
Good work!! I have commented on the test file requesting this to be cleaned up and broken down into seperate Test functions for the two functions in isolation. As well as this I have left you a hint in the `sortLetters` function as to how you can simply the code there. See how you go and if you can spot it.

I'll also ask for you to address the other comments mentioned by @anzboi and update the Pull Request Description, and include a link to a review which you have completed.

03_letters/shahriys/main.go
anzboi: Looks like it is still here

03_letters/shahriys/main_test.go
shampa077: done

03_letters/shahriys/main_test.go
shampa077: added

pentaphobe (changes requested):
Please include a reference to your review of a colleague's PR in your description

juliaogris:

PR #195, opened 2019-05-31
Lab1 - Fibonacci function

### LINKED ISSUE #193: Lab 1 Fibonacci sequence
**Lab 1 - Fibonacci sequence**
- [ ] Create an executable go program in directory 01_fib
```
func fib(n int)
```
```
1
1
2
3
5
8
13
```

Jake88:
Note: I understand storing a map of all the calculated fibonacci numbers is inefficient given the criteria is only to _print_ them, so I have over-engineered here.

Fixes #193

Review of colleague's PR #183

- Solution to lab1 - Fibonacci sequence
- Tests for the fib function

### DIFF
--- a/01_fib/Jake88/main.go
+++ b/01_fib/Jake88/main.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+	store := map[int]int{
+		0: 0,
+	}
+
+	i := 0
+	for i != n {
+		current := 0
+
+		if n < 0 {
+			current = store[i] - store[i-1]
+			if current == 0 {
+				current = 1
+			}
+			store[i-2] = current
+			i--
+		} else {
+			current = store[i] + store[i-1]
+			if current == 0 {
+				current = 1
+			}
+			store[i+1] = current
+			i++
+		}
+		fmt.Fprintln(out, current)
+	}
+}
--- a/01_fib/Jake88/main.go
+++ b/01_fib/Jake88/main.go
@@ -13,29 +13,34 @@ func main() {
 }
 
 func fib(n int) {
-	store := map[int]int{
-		0: 0,
+	// Our intial map, with the default starting values needed to begin the sequence.
+	fibonacciMap := map[int]int{
+		-1: 1,
+		0:  0,
+	}
+	negative := false
+
+	// Check if we are performaing negafibonacci or not, and if so remove the sign from n
+	if n < 0 {
+		n *= -1
+		negative = true
 	}
 
-	i := 0
-	for i != n {
-		current := 0
+	// Form our map of fibonacci numbers, ensuring i matches Fn
+	for i := 1; i < n+1; i++ {
+		// Store the next positive fibonacci iteration
+		current := fibonacciMap[i-1] + fibonacciMap[i-2]
+		fibonacciMap[i] = current
 
-		if n < 0 {
-			current = store[i] - store[i-1]
-			if current == 0 {
-				current = 1
+		// If negafibonacci, store the next negative iteration
+		if negative {
+			if i%2 == 0 {
+				current = -current
 			}
-			store[i-2] = current
-			i--
-		} else {
-			current = store[i] + store[i-1]
-			if current == 0 {
-				current = 1
-			}
-			store[i+1] = current
-			i++
+			fibonacciMap[-i] = current
 		}
+
+		// Print out the current iteration value
 		fmt.Fprintln(out, current)
 	}
 }
--- a/01_fib/Jake88/main.go
+++ b/01_fib/Jake88/main.go
@@ -22,7 +22,7 @@ func absolute(n int) (int, bool) {
 }
 
 func fib(n int) {
-	// Cator for fib(0)
+	// Cater for fib(0).
 	if n == 0 {
 		fmt.Fprintln(out, 0)
 		return
@@ -43,7 +43,7 @@ func fib(n int) {
 
 		prev2, prev1 = prev1, current
 
-		// print based on whether we negafibonacci was requested or not
+		// print based on whether negafibonacci was requested or not.
 		if nega && i%2 == 1 {
 			fmt.Fprintln(out, -current)
 		} else {
--- a/01_fib/Jake88/main.go
+++ b/01_fib/Jake88/main.go
@@ -35,11 +35,15 @@ func fib(n int) {
 	for i := 0; i < n; i++ {
 		current := prev1 + prev2
 
+		// Protect against integer overflow / out of bounds error.
+		if current < prev1 {
+			fmt.Fprintln(out, "Error: Number overflow")
+			return
 		}
 
 		prev2, prev1 = prev1, current
 
-		// print based on whether we are in negafibonacci or not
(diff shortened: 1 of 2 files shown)


01_fib/Jake88/main.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+	// Our intial map, with the default starting values needed to begin the sequence.
```
golangcibot: `intial` is a misspelling of `initial` (from `misspell`)

willshen8:
Hi Jake, can I do the review your PR once it's passed CI? For now, can you please remove 'Ready for review' label?

Jake88:
> Hi Jake, can I do the review your PR once it's passed CI? For now, can you please remove 'Ready for review' label?

More than happy for you to review my changes! :) I've fixed up my typo now, so go for gold.

I do think I need to rebase (or something) my typo commit, but I honestly don't know how to go about doing that so I might need to ask an adult for help!

willshen8:
Hi Jake,
All looks good. Few things to consider:

1. Map would occupy more computing resources compare to array or slice, so computing a larger fib number might have efficiency differences?
2. If you compute large fib number, you might get integer overflow from int, so change it to a larger native type. 
3. A good programming practices is seperation of concerns. You could have a function that generate the map and another function does the print. So that these two functions can be resued if the program gets larger.

Jake88:
> Hi Jake,
> All looks good. Few things to consider:
> 
> 1. Map would occupy more computing resources compare to array or slice, so computing a larger fib number might have efficiency differences?
> 2. If you compute large fib number, you might get integer overflow from int, so change it to a larger native type.
> 3. A good programming practices is seperation of concerns. You could have a function that generate the map and another function does the print. So that these two functions can be resued if the program gets larger.

Thanks!
1. You're right here. In hindsight I should have stuck to the criteria and just stored the last two numbers in the sequence, since we don't need to retain or return the full list!
2. I didn't really think about error handling or scaling too high! I should have tried to break it first and then I would have seen these gaps. ;) Good spot.
3. Sure. I think my fib function is definitely more complex than it needs to be. I guess if the program needed to scale I would look at taking reusable pieces and putting them into their own function.

anzboi:
Please refrain from merging master into your branch until you have the `merge me` label added by an instructor. This way you will have less master merges in your PR.

01_fib/Jake88/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+func fib(n int) {
+	// Our initial map, with the default starting values needed to begin the sequence.
+	fibonacciMap := map[int]int{
+		-1: 1,
+		0:  0,
+	}
+	negative := false
+
+	// Check if we are performaing negafibonacci or not, and if so remove the sign from n
```
anzboi: s/performaing/performing

01_fib/Jake88/main.go
```diff
@@ -0,0 +1,46 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func fib(n int) {
+	// Our initial map, with the default starting values needed to begin the sequence.
+	fibonacciMap := map[int]int{
```
anzboi: This map is overkill. You only use two of its value at any given time, and it is never used once the function returns.
I would suggest you store two variables (perhaps `current` and `previous`) initialized to `1` and `0`, then update them in the loop accordingly.
Alternatively, you can store the map as an unexported `var` outside this function to turn it into a global variable,
```
var fibonacciMap = map[int]int{
  -1: 1,
  0: 0,
}
```
If you go down this path, you will find it harder to write the fibonacci function (and may find it easier to write a recursive function that first checks the map, then populates it if it doesn't exist before returning).
I wouldn't recommend this though.

01_fib/Jake88/main.go
```diff
@@ -0,0 +1,46 @@
...
+		// Store the next positive fibonacci iteration
+		current := fibonacciMap[i-1] + fibonacciMap[i-2]
+		fibonacciMap[i] = current
+
+		// If negafibonacci, store the next negative iteration
+		if negative {
+			if i%2 == 0 {
+				current = -current
+			}
+			fibonacciMap[-i] = current
```
anzboi: The map values are never accessed for negative integers (except `-1` on the first iteration). This is storing numbers that will never be read.

01_fib/Jake88/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	type test struct {
+		input    int
+		expected string
+	}
+
+	tests := []test{
+		{-7, `"1\n-1\n2\n-3\n5\n-8\n13\n"`},
+		{7, `"1\n1\n2\n3\n5\n8\n13\n"`},
+		{-3, `"1\n-1\n2\n"`},
+		{4, `"1\n1\n2\n3\n"`},
```
anzboi, line 37: Can we add a test for `fib(0)`.
I also don't think you need the backticks here (correct me if I am wrong).

anzboi (changes requested):
Great attempt so far, love that you are playing around with various objects in `go` 👍 .
For the purposes of the lab though, I would like you to rethink the use of the map inside the `fib` function. See my comments for more details.
Also please test any edge cases in your labs (`fib(0)` in this case).

01_fib/Jake88/main.go
Jake88: Actioned. Converted to two variables instead.

01_fib/Jake88/main_test.go
Jake88: Actioned. Added both a test and functionality to handle fib(0).

01_fib/Jake88/main.go
Jake88: Actioned. Proof read my new comments and made fixes accordingly.

01_fib/Jake88/main.go
```diff
@@ -0,0 +1,53 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	fib(7)
+}
+
+func absolute(n int) (int, bool) {
+	var negative bool
+	if n < 0 {
+		n *= -1
```
anzboi, line 18: `n=-n` works just as well.

01_fib/Jake88/main.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	prev2, prev1 := 1, 0
+
+	n, nega := absolute(n)
+
+	for i := 0; i < n; i++ {
+		current := prev1 + prev2
+
+		// Protect against integer overflow / out of bounds error.
```
anzboi, line 38: 👍

01_fib/Jake88/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+		if test.expected != actual {
+			t.Errorf("Unexpected output printed by fib.\n(Expected) : %v\n(Actual)   : %v", test.expected, actual)
+		}
+
+		buf.Reset()
+	}
+
+	// Test out of bounds
+	fib(100)
```
anzboi, line 53: 👍 for testing implemented functionality even if not required.
Normally we would write a separate test for this though.

anzboi (approved):
LGTM. Just one thing to think about for future labs.
When testing different behaviour (out of bounds vs normal), it should go in a different test.

PR #197, opened 2019-05-31
Complete Lab 4 Implement function that returns a slice of numeronyms for its input strings

### LINKED ISSUE #289: Undo the removal of 01_fib/kasunfdo and more test cases added

shampa077:
Fixes #289

Review of colleague's PR #

- Complete Lab 4
- Implement function that returns a slice of numeronyms for its input strings


01_fib/kasunfdo/main_test.go
```diff
@@ -1,87 +0,0 @@
-package main
```
n0npax: If these changes are needed, they should be in separate PR.

I guess you don't want to remove these files. You should be able to rebase to upstream master and get rid of this modification from git history.

n0npax: lab04 code LGTM

n0npax (changes requested):

01_fib/kasunfdo/main_test.go
shampa077: @n0npax  those files came from someone committed to real master. You still want me to rebase?

04_numeronym/shahriys/main.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+	"strconv"
+)
+
+var outnum io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	var returnval = make([]string, 0, len(vals))
+	for _, val := range vals {
+		switch len(val) {
```
anzboi, line 15: Interesting use of a `switch` as opposed to an `if` block. Perfectly fine though
You can assign a variable in the `switch` statement and reuse it without having to recompute it
```
switch length := len(vals) {
  case 0, 1, 2, 3:
    // code
  default: 
    l := length - 2
}
```

04_numeronym/shahriys/main.go
```diff
@@ -0,0 +1,31 @@
...
+		case 0, 1, 2, 3:
+			returnval = append(returnval, val)
+		default:
+			l := len(val) - 2
+			s := string(val[0]) + strconv.FormatInt(int64(l), 10) + string(val[l+1])
+			returnval = append(returnval, s)
+		}
+
```
anzboi: Typically extra lines like this is not encouraged in go.

04_numeronym/shahriys/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+	if expected != actual {
+		t.Errorf("Unexpected output in main()" + expected + " " + actual)
+	}
+
+}
+func TestNumeronym(t *testing.T) {
+
+	type test struct {
+		expected []string
+		actual   []string
```
anzboi: Normally you would only store the input and expected, not actual, rather than evaluating the function while creating the test data.
This is because often times you test functions that can return multiple values, you can't capture multiple return values in an inline struct initialization.
I suggest using an input slice, and evaluating the result in the test before testing it. Remember to call with
`numeronyms(tc.input...)`

04_numeronym/shahriys/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	tests := []test{
+		{expected: []string{"a2c"}, actual: numeronyms("abac")},
+		{expected: []string{}, actual: numeronyms("")},
+		{expected: []string{"a"}, actual: numeronyms("a")},
+		{expected: []string{"aa", "s3s", "s10g", ""}, actual: numeronyms("aa", "sssss", "sdhjhjhjhjfg", "")},
```
anzboi: Good coverage, though I can think of one more test case,
`numeronyms()`

04_numeronym/shahriys/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestNumeronym(t *testing.T) {
+
+	type test struct {
+		expected []string
+		actual   []string
+	}
+
+	tests := []test{
+		{expected: []string{"a2c"}, actual: numeronyms("abac")},
+		{expected: []string{}, actual: numeronyms("")},
```
anzboi: I'm surprised this test passes. An empty string `""` should have numeronym `""`, not nothing. I would expect this to return `[]string{""}`

anzboi (dismissed):
Looking good.
You numeronym function looks fine, though there is a test case I would expect to fail. Can you please comment on that.
Also, your branch currently deletes @kasunfdo 's completed lab1. Can you fix you branch so that it doesn't modify their lab ([see here](https://github.com/anz-bank/go-course/wiki/Don't-delete-completed-labs)).

n0npax (approved):
lgtm

04_numeronym/shahriys/main.go
```diff
@@ -0,0 +1,29 @@
...
+func numeronyms(vals ...string) []string {
+	var returnval = make([]string, 0, len(vals))
+	for _, val := range vals {
+		switch len(val) {
+		case 0, 1, 2, 3:
+			returnval = append(returnval, val)
+		default:
+			l := len(val) - 2
+			s := string(val[0]) + strconv.FormatInt(int64(l), 10) + string(val[l+1])
```
rokane, line 20: An alternate approach which I find a little neater is to use the `fmt.Sprintf()` function.

```go
fmt.Sprintf("%c%d%c", val[0], l, val[l+1])
```

Your's is fine though

04_numeronym/shahriys/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	outnum = &bufnum
+
+	main()
+	expected := `[a11y K8s abc]`
+	actual := bufnum.String()
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()" + expected + " " + actual)
+	}
+
```
rokane, line 20: Little nit, but you don't need the extra white space at the end of each function. Keep this in mind in future.

04_numeronym/shahriys/main_test.go
```diff
@@ -0,0 +1,56 @@
...
+	for name, test := range cases {
+		actual := numeronyms(test.input)
+		if !reflect.DeepEqual(actual, test.expected) {
+			t.Errorf(" test:%v input:%v, expected: %v, got: %v", name, test.input, test.expected, actual)
+		}
+	}
+
+}
+func TestLettersMulti(t *testing.T) {
```
rokane, line 39: There is no need to have this split across multiple tests. The underlying test code is the same, it is only the cases which differ. You should add these cases into the above test and rename accordingly -`TestNumeronyms()` Can you please update this?

If you want each output to appear as an individual test then you can run them as sub tests within the one function. Like below:

```go

    input := test.input
    expected := test.expected
        assert.Equal(t, expected, numeronyms(input), "Expected %v, Got %v", expected, numeronyms(input))
    })
}

```

rokane (changes requested):
Just a minor request on the structure of your tests, if you can address that this will be good to go :)

juliaogris (changes requested):
Please add link to review of colleagues PR

anzdaddy:
Please close merge or close.

PR #201, opened 2019-05-31
Implement sorted letter counting

### LINKED ISSUE #230: Sorted letter counts
- [X] Add Simple app to demonstrate sorted counts
- [X] Implement letters function to count runes in a string
- [X] Implement sortLetters function to sort rune counts
- [X] 100% coverage

nicholascross:
Fixes: #230

Review of colleague's PR #219

 - Add Simple app to demonstrate sorted counts
 - Implement letters function to count runes in a string
 - Implement sortLetters function to sort rune counts

### DIFF
--- a/03_letters/nicholascross/main.go
+++ b/03_letters/nicholascross/main.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]string, 0, len(m))
+	for k, v := range m {
+		keys = append(keys, fmt.Sprintf("%c:%d", k, v))
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func letters(s string) map[rune]int {
+	count := map[rune]int{}
+
+	for _, char := range s {
+		count[char]++
+	}
+	return count
+}
--- a/03_letters/nicholascross/main_test.go
+++ b/03_letters/nicholascross/main_test.go
@@ -1,0 +1,79 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote(`a:2
+b:1
+`)
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main() %v", actual)
+	}
+}
+
+func TestSortLetters(t *testing.T) {
+	var testCases = map[string]struct {
+		input map[rune]int
+		want  []string
+	}{
+		"correct counts": {
+			input: map[rune]int{'a': 1, 'b': 2, 'c': 3, 'd': 4, 'e': 5},
+			want:  []string{"a:1", "b:2", "c:3", "d:4", "e:5"},
+		},
+		"sorted counts": {
+			input: map[rune]int{'c': 3, 'd': 4, 'e': 5, 'a': 1, 'b': 2},
+			want:  []string{"a:1", "b:2", "c:3", "d:4", "e:5"},
+		},
+		"empty": {input: map[rune]int{}, want: []string{}},
+	}
+
+	for name, test := range testCases {
+		input := test.input
+		want := test.want
+		t.Run(name, func(t *testing.T) {
+			assert.Equalf(t, sortLetters(input), want, "%v was not sorted", input)
+		})
+	}
+}
+
+func TestLetters(t *testing.T) {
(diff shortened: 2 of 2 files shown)


03_letters/nicholascross/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+	}{
+		"correct counts": {
+			input: "abbcccddddeeeee",
+			want:  map[rune]int{'a': 1, 'b': 2, 'c': 3, 'd': 4, 'e': 5},
+		},
+		"unordered counts": {
+			input: "ebdbcaddecdecee",
+			want:  map[rune]int{'a': 1, 'b': 2, 'c': 3, 'd': 4, 'e': 5},
+		},
+		"special characters" :{
```
golangcibot: File is not `goimports`-ed (from `goimports`)

03_letters/nicholascross/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]string, 0, len(m))
+	for k := range m {
```
kasunfdo: Improvement: Use range with key value pair. Then no need to access map again using `m[k]`
```
for k, v := range m {
  ...
}
```

kasunfdo (changes requested):
Looks great! One improvement is suggested.

03_letters/nicholascross/main.go
nicholascross: Thanks - that is better.  I will update 👍

kasunfdo (dismissed):
Good job!

03_letters/nicholascross/main.go
```diff
@@ -0,0 +1,33 @@
...
+	for k, v := range m {
+		keys = append(keys, fmt.Sprintf("%c:%d", k, v))
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func letters(s string) map[rune]int {
+	count := map[rune]int{}
+
```
rokane, line 28: Little nit, I would allocate the space here and remove the additional line.

```go
count := make(map[rune]int)
for _, char := range s {
  count[char]++
}
return count
```

03_letters/nicholascross/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+		"special characters": {
+			input: "无马e😊e马👍👍a👌😒",
+			want:  map[rune]int{'a': 1, '马': 2, '👍': 2, '👌': 1, 'e': 2, '无': 1, '😊': 1, '😒': 1},
+		},
+		"empty": {input: "", want: map[rune]int{}},
+	}
+
+	for name, test := range testCases {
+		input := test.input
+		want := test.want
```
rokane, line 74: I prefer `expected` as a name over `want`.

rokane (changes requested):
Well done, great tests and succinct code. Just a couple of real minor things, then this will look good to me 👍

03_letters/nicholascross/main_test.go
nicholascross: Sure, I received some early feedback that my variable names were too long, I actually prefer that too, I am still trying to find the right balance between brevity and clarity in `go`.

I will update 👍

03_letters/nicholascross/main.go
nicholascross: @rokane I am confused, what is the difference between these statements?

```
map[rune]int{}
```

```
make(map[rune]int)
```

I thought they both initialise an empty map?

nicholascross: Do we prefer one style of initialisation over the other, is that it?

03_letters/nicholascross/main_test.go
rokane: Yeah that makes sense, within tests I would generally always follow the `input / expected` convention. But for the same short lived variable within a normal function, I would be more concise with my naming. It is a bit of personal preference as well I guess.

03_letters/nicholascross/main.go
rokane: I've just been reading up on this, the two result in the same. I am happy for you to leave it either way, I usually always declare it with the `make` keyword, but it seems totally fine to do use the `{}` as well. Alternatively you could declare the maximum size of the map upfront which might give a slight performance bump, but in this context not required.
```go
make(map[rune]int, len(s))
```
[What is the difference between make and {} to initialise a map](https://groups.google.com/forum/#!topic/golang-nuts/YnqSi2yURSY)

03_letters/nicholascross/main_test.go
rokane: So I have just taken a look at the example we provide in the course notes, and it uses `want`. Hahaha maybe I am just as confused. Feel free to keep it as is as we are endorsing this in the notes 👍 Sorry if I am just causing further confusion.

rokane (dismissed):
Really good mate 👍

nicholascross:
@rokane Thanks for the review.  I made the mistake of rebasing to get my branch up to date, this has dismissed your approval.

rokane:
@nicholascross no worries.

rokane (approved):

PR #132, opened 2019-05-20
Lab 2 - Bubble sort

kasunfdo:
Fixes #166

Review of colleague's PR #155

#### Changes proposed in this PR:
* Implement bubble sort
* Implement insertion sort
* Implement merge sort

PR #206, opened 2019-06-01
Lab02

### LINKED ISSUE #223: Lab02 implement bubble sort
- [ ] Create an executable go program in directory 02_bubble/jonathanheaden
- [ ] Write a function that returns a sorted copy of int slice s using bubble sort:

jonathanheaden:
Fixes #223 

Review of colleague's PR #132


02_bubble/jonathanheaden/main.go
```diff
@@ -0,0 +1,31 @@
...
+		sw := false
+		for i := 0; i < len(s)-1; i++ {
+			if s[i] > s[i+1] {
+				sw = true
+				s[i], s[i+1] = s[i+1], s[i]
+			}
+			done = !sw
+		}
+	}
+	return s
```
nicholascross: Does this mutate the original slice? If so is that what you intended?

02_bubble/jonathanheaden/main.go
```diff
@@ -0,0 +1,31 @@
...
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	done := false
+	if len(s) == 0 {
+		return s
+	}
+	for !done {
+		sw := false
```
nicholascross, line 19: The additional variable might not necessary, you might be able to track sort completion with a single variable.

02_bubble/jonathanheaden/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	fmt.Fprintln(out, bubble([]int{3, -2, 1, -5, 7}))
+
+	// Then
+	expected := strconv.Quote("[-5 -2 1 3 7]\n")
+	actual := strconv.Quote(buf.String())
+	r.Equalf(expected, actual, "Unexpected output in main()")
```
nicholascross, line 39: When I was solving this lab I used `github.com/stretchr/testify/assert`, I noticed you have used `github.com/stretchr/testify/require` I just [read](https://godoc.org/github.com/stretchr/testify/require) up on the difference and it seems require will stop test execution on failure, is that what you want?

02_bubble/jonathanheaden/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func TestAllElemetsEqual(t *testing.T) {
```
nicholascross, line 42: Nice, I didn't think about testing this case 👍

02_bubble/jonathanheaden/main_test.go
jonathanheaden: For the purposes of this exercise I do want it to stop on fail, but thanks for pointing this out. Assert is what I would usually use

02_bubble/jonathanheaden/main.go
jonathanheaden: Yes it does mutate the original slice, and for the purpose of this exercise it is what I want. If I were implementing without already having the func signature defined in the readme I would not have it return anything but instead mutate the slice in place as written

02_bubble/jonathanheaden/main.go
jonathanheaden: I use the two vars for tracking overall completion between the two loops, done only gets set true when all iterations of the the `sw` swap flag in the inner loop return false (no more sorting required).

02_bubble/jonathanheaden/main_test.go
jonathanheaden: thanks

02_bubble/jonathanheaden/main.go
jonathanheaden: It was how I intended originally, but I agree it should not mutate the original slice. if the signature of the function didn't return anything then mutation would be expected. As the signature is written it should not mutate the passed in slice but work on a copy instead. I will fix this up

02_bubble/jonathanheaden/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	done := false
+	if len(s) == 0 {
+		return s
+	}
+	r := make([]int, len(s))
+	copy(r,s)
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

jonathanheaden: I ran `go fmt` and pushed changes

nicholascross (approved):

02_bubble/jonathanheaden/main.go
rokane: I think you might be able to achieve the same thing with a single variable as @nicholascross mentioned.

```go
sorted := false
for !sorted {
    sorted = true
    for i := 0; i < len(r)-1; i++ {
        if r[i] > r[i+1] {
            sorted = false
            r[i], r[i+1] = r[i+1], r[i]
        }
    }
}
```

I haven't tested it out properly but I think that might do the job.

02_bubble/jonathanheaden/main_test.go
rokane, line 42: Nice work, well structured tests. You should (and probably have) look to structure your tests so you are iterating over different cases in the future labs. That way you avoid the repetitive nature of the test skeleton code.

rokane (approved):
LGTM 👍 I have mentioned a couple of minor things which could be applied, but I won't hold you from merging as they are very minor. Keep up the good work :)

02_bubble/jonathanheaden/main_test.go
jonathanheaden: will do
thanks

PR #200, opened 2019-05-31
Lab1

shuutech:
Fixes #301 

Review of colleague's PR #196 196

PR #209, opened 2019-06-01
Lab03 - Letter frequency

### LINKED ISSUE #207: Lab03 - letter frequency

n0npax:
Fixes #207 

Review of colleague's PR #200

Implement solution for lab03

* letterFrequency
* tests

### DIFF
--- a/03_letters/n0npax/main.go
+++ b/03_letters/n0npax/main.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+func letters(s string) (freqMap map[rune]int) {
+	freqMap = map[rune]int{}
+	ss := []rune(s)
+	for i := range ss {
+		freqMap[ss[i]]++
+	}
+	return
+}
+
+type letterFreq struct {
+	letter    rune
+	occurency int
+}
+
+func (lf letterFreq) String() string {
+	return fmt.Sprintf("%c:%d", lf.letter, lf.occurency)
+}
+
+func sortLetters(m map[rune]int) (sortedLetters []string) {
+
+	freq, i := make([]letterFreq, len(m)), 0
+	for l, occurency := range m {
+		lf := letterFreq{letter: l, occurency: occurency}
+		freq[i] = lf
+		i++
+	}
+	sort.Slice(freq, func(i, j int) bool {
+		return freq[i].occurency > freq[j].occurency
+	})
+	sortedLetters = make([]string, len(freq))
+	for i, lf := range freq {
+		sortedLetters[i] = lf.String()
+	}
+	return
+}
+
+var out io.Writer = os.Stdout
+
+func main() {
+	text := "aba"
+	fmt.Fprintln(out, strings.Join(sortLetters(letters(text)), "\n"))
+}
--- a/03_letters/n0npax/main.go
+++ b/03_letters/n0npax/main.go
@@ -8,40 +8,39 @@ import (
 	"strings"
 )
 
-func letters(s string) (freqMap map[rune]int) {
-	freqMap = map[rune]int{}
+func letters(s string) map[rune]int {
+	freqMap := map[rune]int{}
 	ss := []rune(s)
 	for i := range ss {
 		freqMap[ss[i]]++
 	}
-	return
+	return freqMap
 }
 
 type letterFreq struct {
-	letter    rune
-	occurency int
+	letter rune
+	count  int
 }
 
 func (lf letterFreq) String() string {
-	return fmt.Sprintf("%c:%d", lf.letter, lf.occurency)
+	return fmt.Sprintf("%c:%d", lf.letter, lf.count)
 }
 
-func sortLetters(m map[rune]int) (sortedLetters []string) {
-
+func sortLetters(m map[rune]int) []string {
 	freq, i := make([]letterFreq, len(m)), 0
-	for l, occurency := range m {
-		lf := letterFreq{letter: l, occurency: occurency}
+	for l, count := range m {
+		lf := letterFreq{letter: l, count: count}
 		freq[i] = lf
 		i++
 	}
 	sort.Slice(freq, func(i, j int) bool {
-		return freq[i].occurency > freq[j].occurency
+		return freq[i].count > freq[j].count
 	})
-	sortedLetters = make([]string, len(freq))
+	sortedLetters := make([]string, len(freq))
 	for i, lf := range freq {
 		sortedLetters[i] = lf.String()
 	}
-	return
+	return sortedLetters
 }
 
 var out io.Writer = os.Stdout
--- a/03_letters/n0npax/main_test.go
+++ b/03_letters/n0npax/main_test.go
@@ -1,0 +1,109 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestLetters(t *testing.T) {
+	testCases := map[string]struct {
+		input    string
+		expected map[rune]int
+	}{
+		"polish": {input: "dzień dobry",
+			expected: map[rune]int{'y': 1, ' ': 1, 'd': 2, 'z': 1, 'i': 1, 'e': 1, 'ń': 1, 'o': 1, 'b': 1, 'r': 1}},
+		"russian": {input: "доброе утро",
+			expected: map[rune]int{'о': 3, 'р': 2, 'т': 1, 'у': 1, ' ': 1, 'е': 1, 'б': 1, 'д': 1}},
+		"english": {input: "good morning",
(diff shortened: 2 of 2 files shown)


03_letters/n0npax/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"strings"
+)
+
+func letters(s string) (freqMap map[rune]int) {
+	freqMap = map[rune]int{}
+	ss := []rune(s)
+	for i := range ss {
+		freqMap[ss[i]]++
+	}
+	return
```
rokane: Interesting, I haven't seen this too often. I would normally go down the path of explicitly returning the `freqMap` variable as I feel it helps with readability, and is clear what you are returning. But I don't hate, it is growing on me.

03_letters/n0npax/main.go
```diff
@@ -0,0 +1,52 @@
...
+	"os"
+	"sort"
+	"strings"
+)
+
+func letters(s string) (freqMap map[rune]int) {
+	freqMap = map[rune]int{}
+	ss := []rune(s)
+	for i := range ss {
+		freqMap[ss[i]]++
```
rokane, line 15: Nice work 👍

03_letters/n0npax/main.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+	return
+}
+
+type letterFreq struct {
+	letter    rune
+	occurency int
```
rokane: Can you please rename this to `freq` or `count`

```go
type letterFreq struct {
  letter  rune
  count int
}
```

03_letters/n0npax/main_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+}
+
+func helperTestLetter(name string, input string, expected map[rune]int) func(*testing.T) {
+	return func(t *testing.T) {
+		result := letters(input)
+		for k, v := range result {
+			if v != expected[k] {
+				for k := range result {
+					if expected[k] != result[k] {
```
rokane: I think that this is starting to get quite confusing and difficult to read. I don't think it is necessary to do the final iteration over the result. Can you please remove the double nested loop like below?

```go
for k, count := range result {
  if _, ok := expected[k]; !ok {
    t.Errorf("Found unexpected key: %c", k)
  }
  if count != expected[k] {
    t.Errorf("Incorrect count for key %c. Found: %d, Expected %d", k, count, expected[k])
  }
}
```

rokane (changes requested):
Looking really good. I like the way you are breaking things down and abstracting your test suites out. If you can please update according to the two comments I have provided, then this will be right for merging 👍

03_letters/n0npax/main_test.go
```diff
@@ -0,0 +1,109 @@
...
+		"chineese": {input: "早上好",
+			expected: map[rune]int{'早': 1, '上': 1, '好': 1}},
+	}
+
+	for name, test := range testCases {
+		t.Run(name, helperTestLetter(name, test.input, test.expected))
+	}
+}
+
+func helperTestLetter(name string, input string, expected map[rune]int) func(*testing.T) {
```
golangcibot: `helperTestLetter` - `name` is unused (from `unparam`)

03_letters/n0npax/main_test.go
n0npax: Thanks for that. Fixed :)

03_letters/n0npax/main.go
n0npax: done. Thanks for that

03_letters/n0npax/main.go
juliaogris: Avoid named result parameters and naked returns unless you have got good reason  (source [golang wiki](https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters)) - which I believe you don't here.

03_letters/n0npax/main.go
```diff
@@ -0,0 +1,52 @@
...
+	letter rune
+	count  int
+}
+
+func (lf letterFreq) String() string {
+	return fmt.Sprintf("%c:%d", lf.letter, lf.count)
+}
+
+func sortLetters(m map[rune]int) (sortedLetters []string) {
+
```
juliaogris: unnecessary blank line

03_letters/n0npax/main.go
```diff
@@ -0,0 +1,52 @@
...
+		i++
+	}
+	sort.Slice(freq, func(i, j int) bool {
+		return freq[i].count > freq[j].count
+	})
+	sortedLetters = make([]string, len(freq))
+	for i, lf := range freq {
+		sortedLetters[i] = lf.String()
+	}
+	return
```
juliaogris: avoid naked returns and named result parameters, as above.

03_letters/n0npax/main_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+func TestLetterFreqString(t *testing.T) {
+	testCases := map[string]struct {
+		input    letterFreq
+		expected string
+		positive bool
+	}{
+		"latinum":  {input: letterFreq{letter: 'a', count: 2}, expected: "a:2", positive: true},
+		"cyrillic": {input: letterFreq{letter: 'а', count: 2}, expected: "а:2", positive: true},
+		// cyrillic 'а' is different char than latin 'a'
```
juliaogris, line 50: neat!

03_letters/n0npax/main_test.go
```diff
@@ -0,0 +1,109 @@
...
+		"mixed": {input: letterFreq{letter: 'а', count: 2}, expected: "a:2", positive: false},
+	}
+	for name, test := range testCases {
+		t.Run(name, helperTestLetterFreqString(name, test.input, test.expected, test.positive))
```
juliaogris: Nit: this is a matter of style, but imo it is better to keep TestFuncs purely self contained module some test fixtures to generate input data. Helpers should not be used to create assertions.

I your case I'd just pull out a bunch of `NotEqual` and `Equal` test cases in two separate test functions and test tables I also recommend using `testify/assert` at this stage.

then your asserts could just by:
```go
for name, tc := range testCases {
        got :=  tc.input.String()
        assert.Equal(t, tc.expected, got)
    } )
}
```
and in the not equal case:
```go
for name, tc := range testCases {
        got :=  tc.input.String()
        assert.NotEqual(t, tc.notExpected, got)
    } )
}
```

juliaogris (dismissed):
A few more little nitpicks - overall looks really great!

Please fix up your last commit messages: `CR fixes` should be `Fix miscellaneous according to CR` or similar. Better yet: squash all commits into a single commit  with `git rebase -i ` or `git commit --amend` to update only the last commit and then `git push -f`.
If you do want separate commits, I recommend: `git commit --fixup <HASH>` where <HASH> is the commit SHA1 of the commit which is being fixed up; if you run `git rebase -i --autosquash master` it should do as you want.

03_letters/n0npax/main.go
n0npax: fixed. Will avoid this in the future. Thanks!

03_letters/n0npax/main.go
n0npax: fixed. Thanks

03_letters/n0npax/main.go
n0npax: Fixed. Thanks!

03_letters/n0npax/main_test.go
n0npax: fixed. Thanks for that!

03_letters/n0npax/main_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	for name, test := range testCases {
+		t.Run(name, helperTestLetter(test.input, test.expected))
+	}
+}
+
+func helperTestLetter(input string, expected map[rune]int) func(*testing.T) {
+	return func(t *testing.T) {
+		result := letters(input)
+		for k, count := range result {
```
rokane, line 34: I just realised there is another case which you should guard against in this test, you should check the length of the result against the length of the expected. 
If there were a case where `result := { 'a': 1, 'b': 2 }` and `expected := { 'a': 1, 'b': 2, 'c': 3 }` the below wouldn't be able to catch it.

```go
result := letters(input)
if len(result) != len(expected) {
  t.Errorf("some error msg here ... ")
}
for k, count := range result {
  ...
}

rokane (changes requested):
Apologies, I should have picked up on that case the first time I reviewed this. Are you able to please include that one and then I will take another look first thing in the morning.

03_letters/n0npax/main_test.go
n0npax: Fixed. Thanks

rokane (approved):

PR #211, opened 2019-06-01
lab04 - numeronyms

### LINKED ISSUE #210: Lab 04

n0npax:
Implement numeronyms
Implement unit tests

Fixes #210 

Review of colleague's PR #197

## - Solve lab 04
implement main logic
implement unit tests

### DIFF
--- a/04_numeronym/n0npax/main.go
+++ b/04_numeronym/n0npax/main.go
@@ -9,11 +9,12 @@ import (
 var out io.Writer = os.Stdout
 
 func shorten(word string) string {
-	if len(word) < 4 {
+	wordLen := len(word)
+	if wordLen < 4 {
 		return word
 	}
-	num := len(word) - 2
-	return fmt.Sprintf("%c%d%c", word[0], num, word[len(word)-1])
+	num := wordLen - 2
+	return fmt.Sprintf("%c%d%c", word[0], num, word[wordLen-1])
 }
 
 func numeronyms(vals ...string) (shortened []string) {
--- a/04_numeronym/n0npax/main.go
+++ b/04_numeronym/n0npax/main.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func shorten(word string) string {
+	if len(word) < 4 {
+		return word
+	}
+	num := len(word) - 2
+	return fmt.Sprintf("%c%d%c", word[0], num, word[len(word)-1])
+}
+
+func numeronyms(vals ...string) (shortened []string) {
+	shortened = make([]string, len(vals))
+	for i, val := range vals {
+		shortened[i] = shorten(val)
+	}
+	return
+}
+
+func main() {
+	fmt.Fprintln(out, numeronyms("Kubernetes", "anz", "internationalization"))
+}
--- a/04_numeronym/n0npax/main_test.go
+++ b/04_numeronym/n0npax/main_test.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("[K8s anz i18n]\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output. Expected: %q - Actual: %q", expected, actual)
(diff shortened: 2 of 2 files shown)


04_numeronym/n0npax/main.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func shorten(word string) string {
+	if len(word) < 4 {
```
nicholascross: Could potentially use a local variable for length I notice it is derived several times.

04_numeronym/n0npax/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+	testCases := map[string]struct {
+		input    string
+		expected string
+	}{
+		"anz":                  {input: "anz", expected: "anz"},
+		"Kubernetes":           {input: "Kubernetes", expected: "K8s"},
+		"internationalization": {input: "internationalization", expected: "i18n"},
+	}
+	for name, test := range testCases {
+		t.Run(name, helperTestShorten(test.input, test.expected))
```
nicholascross: Could we just use the anonymous function inline?  I don't believe we would lose any readability, it might make it a bit more succinct.

n0npax: It will cause referring to variables out of the scope and `golangci-lint` will complain

04_numeronym/n0npax/main.go
n0npax: Thanks for this. Fixed

04_numeronym/n0npax/main_test.go
nicholascross: I believe it is probably complaining because the function is capture the test value which will possibly be incorrect by the time the function is executed.  You can resolve this by shadowing the variable.

```go
		input := test.input
		want := test.want
			assert.Equalf(t, want, numeronyms(input...), "%v was not sorted", input)
		})
	}
```

~If you really want to use a separate function do we still need the internal anonymous function inside `helperTestShorten`?  Is it possible to use it directly?~

**EDIT**

I think if you tried to use it directly you would probably end up with the same problem again, so maybe that is not a good suggestion.  Separating out into a separate function achieves the same thing as creating local variables it is just more verbose.

n0npax: The suggested solution works. Thanks!
I'm surprised the shadowing solves this. `want` and `input` is still outside of the anonymous function.

04_numeronym/n0npax/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func shorten(word string) string {
```
anzboi, line 11: Don't feel the need to change this, since this is already fine as is.
Just wanted to point out you can also assign `wordlen` in the `if` block...
```
if wordlen := len(word); wordlen < 4 {
  return word
} else {
  return fmt.Sprintf("%c%d%c", word[0], wordlen-2, word[wordLen-1])
}
```

04_numeronym/n0npax/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	}{
+		"anz":                  {input: "anz", expected: "anz"},
+		"Kubernetes":           {input: "Kubernetes", expected: "K8s"},
+		"internationalization": {input: "internationalization", expected: "i18n"},
+	}
+	for name, test := range testCases {
+		input, want := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			actual := shorten(input)
+			assert.Equalf(t, want, actual, "For input: %v expected: %v got %v", input, want, actual)
```
anzboi, line 48: Again, don't feel the need to change this.
You can avoid the need to pass `t` into these `testify` functions by creating an assertions object at the start of the test
```
func TestName(t *testing.T) {
  a := assert.New(t)
  // Test Code
  a.Equalf(want, actual, "message") // Don't need to pass t
}
```

This is really a matter of preference, you trade having to pass `t` for having to create `a` at the start.

anzboi (approved):
LGTM
Some comments on how you can do things a little differently, but these are mostly up to preference.

04_numeronym/n0npax/main_test.go
nicholascross: Thanks for pointing this out 👍

04_numeronym/n0npax/main.go
n0npax: thanks for hint

PR #213, opened 2019-06-02
Implement Fibonacci and NegaFibonacci sequences

undrewb:
Fixes #255 
Review of colleague's PR #246

Changes proposed in this PR:

Implementation of fibonacci and negafibonacci sequences as specified in lab1 of the anz go-course.
Updated to use iteration used and I've put in a limit of +-60 entries to stop overflow errors occuring.

PR #215, opened 2019-06-02
Complete lab05 - Stringer

### LINKED ISSUE #214: Lab 05

n0npax:
Implement IPAddr Stringer method
Add tests

Fixes #214 

Review of colleague's PR #213

Add `IPAddr` type backed by an array of 4 bytes
Add `String` method for `IPAddr`
Add unit tests

### DIFF
--- a/05_stringer/n0npax/main.go
+++ b/05_stringer/n0npax/main.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+// IPAddr is a type to encapsulate IPv4 (4 bytes)
+type IPAddr [4]byte
+
+// Stringer for IPAddr
+func (addr IPAddr) String() string {
+	values := make([]string, len(addr))
+	for i, v := range addr {
+		values[i] = fmt.Sprint(v)
+	}
+	return strings.Join(values, ".")
+}
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, IPAddr{127, 0, 0, 1})
+}
--- a/05_stringer/n0npax/main.go
+++ b/05_stringer/n0npax/main.go
@@ -4,19 +4,12 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 )
 
-// IPAddr is a type to encapsulate IPv4 (4 bytes)
 type IPAddr [4]byte
 
-// Stringer for IPAddr
 func (addr IPAddr) String() string {
-	values := make([]string, len(addr))
-	for i, v := range addr {
-		values[i] = fmt.Sprint(v)
-	}
-	return strings.Join(values, ".")
+	return fmt.Sprintf("%d.%d.%d.%d", addr[0], addr[1], addr[2], addr[3])
 }
 
 var out io.Writer = os.Stdout
--- a/05_stringer/n0npax/main_test.go
+++ b/05_stringer/n0npax/main_test.go
@@ -4,6 +4,8 @@ import (
 	"bytes"
 	"strconv"
 	"testing"
+
+	assertion "github.com/stretchr/testify/assert"
 )
 
 func TestMainOutput(t *testing.T) {
@@ -33,17 +35,15 @@ func TestStringify(t *testing.T) {
 		"mask 32":              {input: IPAddr{255, 255, 255, 255}, expected: "255.255.255.255"},
 		"mask 0":               {input: IPAddr{0, 0, 0, 0}, expected: "0.0.0.0"},
 		"anz":                  {input: IPAddr{202, 2, 59, 40}, expected: "202.2.59.40"},
+		"empty":                {input: IPAddr{}, expected: "0.0.0.0"},
+		"half-empty":           {input: IPAddr{10, 10}, expected: "10.10.0.0"},
 	}
 	for name, test := range testCases {
-		t.Run(name, helperTestStringify(test.input, test.expected))
-	}
-}
-
-func helperTestStringify(input IPAddr, expected string) func(*testing.T) {
-	return func(t *testing.T) {
-		actual := input.String()
-		if actual != expected {
-			t.Errorf("Expected: %q - Actual: %q", expected, actual)
-		}
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			assert := assertion.New(t)
+			actual := input.String()
+			assert.Equalf(expected, actual, "Testcase %s failed")
+		})
 	}
 }
--- a/05_stringer/n0npax/main_test.go
+++ b/05_stringer/n0npax/main_test.go
@@ -1,0 +1,49 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("127.0.0.1\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output. Expected: %q - Actual: %q", expected, actual)
+	}
+}
+
+func TestStringify(t *testing.T) {
+	testCases := map[string]struct {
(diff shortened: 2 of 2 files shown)


05_stringer/n0npax/main.go
```diff
@@ -0,0 +1,26 @@
...
+// IPAddr is a type to encapsulate IPv4 (4 bytes)
+type IPAddr [4]byte
+
+// Stringer for IPAddr
+func (addr IPAddr) String() string {
+	values := make([]string, len(addr))
+	for i, v := range addr {
+		values[i] = fmt.Sprint(v)
+	}
+	return strings.Join(values, ".")
```
nicholascross: This all looks good to me.  

I just wanted to point out that since you know the exact size of the array you don't need to loop it.  You could have done a one liner instead.

```go
func (addr IPAddr) String() string {
	return fmt.Sprintf("%d.%d.%d.%d", addr[0], addr[1], addr[2], addr[3])
}
```

nicholascross: 😆 I just realised you reviewed my PR so you already know this

nicholascross (dismissed):

juliaogris (dismissed):
Please [fix your pr description](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#pr-description) and update the `String` function as suggested by @nicholascross .

The rest looks great!

n0npax:
@juliaogris fixed. I thought changes proposed by Nicholas are optional, sorry.
The PR description and `String` implementation updated.

05_stringer/n0npax/main.go
n0npax: refactored. Cheers :)

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"mask 32":              {input: IPAddr{255, 255, 255, 255}, expected: "255.255.255.255"},
+		"mask 0":               {input: IPAddr{0, 0, 0, 0}, expected: "0.0.0.0"},
+		"anz":                  {input: IPAddr{202, 2, 59, 40}, expected: "202.2.59.40"},
+		"empty":                {input: IPAddr{}, expected: "0.0.0.0"},
+		"half-empty":           {input: IPAddr{10, 10}, expected: "10.10.0.0"},
+	}
+	for name, test := range testCases {
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			mt := assert.New(t)
```
rokane: I'm not really sure what `mt` is supposed to stand for here. It is super minor so I won't block you from merging, but I would suggest more descriptive. When I read something like 

```go
assert := assert.New()
assert.Equalf(expected, actual, "Testcase %s failed")
```
It is a lot clearer to me. Although to do this you would have to provide an alias for the `testify/assert` package, something like below:
```go
import assertion "github.com/stretchr/testify/assert"

assert := assertion.New()
assert.Equalf(expected, actual, "Testcase %s failed")
```

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"localhost":            {input: IPAddr{127, 0, 0, 1}, expected: "127.0.0.1"},
+		"google primary dns":   {input: IPAddr{8, 8, 8, 8}, expected: "8.8.8.8"},
+		"google secondary dns": {input: IPAddr{8, 8, 4, 4}, expected: "8.8.4.4"},
+		"CloudFare":            {input: IPAddr{1, 1, 1, 1}, expected: "1.1.1.1"},
+		"mask 24":              {input: IPAddr{255, 255, 255, 0}, expected: "255.255.255.0"},
+		"mask 32":              {input: IPAddr{255, 255, 255, 255}, expected: "255.255.255.255"},
+		"mask 0":               {input: IPAddr{0, 0, 0, 0}, expected: "0.0.0.0"},
+		"anz":                  {input: IPAddr{202, 2, 59, 40}, expected: "202.2.59.40"},
+		"empty":                {input: IPAddr{}, expected: "0.0.0.0"},
+		"half-empty":           {input: IPAddr{10, 10}, expected: "10.10.0.0"},
```
rokane, line 39: Good work identifying these cases 👍

rokane (dismissed):
Looking really good here :) Great use of the testify package as well. I really love this package and basically use it for all my tests now, makes things a lot easier!

05_stringer/n0npax/main_test.go
n0npax: In other PR I was suggested to provide a similar construct. The aliased `assert` is much more readable. Thanks for pointing that. Anyway Fixed :) Could you approve once again, please?

rokane (approved):

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	assertion "github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("127.0.0.1\n")
```
alextmz, line 17: I'd move expected right to the top to increase legibility, but it's really a nit as the code is very short here anyway.

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output. Expected: %q - Actual: %q", expected, actual)
```
alextmz, line 21: And since you're importing assert anyway, I'd use assert.Equal to make it one line one and a bit clearer (IMHO). But it's really nitpicking again, running out of options here :)

05_stringer/n0npax/main_test.go
alextmz: Nitpick-o-tron to 11 :japanese_ogre:, not sure if there is value in, for example, testing both google dns IPs, plus cloudflare and ANZ, as they don't look like running into any different error scenario (off-by-1, zeros etc). And for consistency, doing a Camh just because you were the first one to finish (and I'm the oh-very-last!)  :sweat_smile: if you're testing masks 0/24/32, you're missing mask 16 and ordering them - 24,32,0? C'mon, you can do better! :joy_cat:

Jokes apart, excellent choice of testcases too - I haven't myself realized to test some of them. :+1:

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"anz":                  {input: IPAddr{202, 2, 59, 40}, expected: "202.2.59.40"},
+		"empty":                {input: IPAddr{}, expected: "0.0.0.0"},
+		"half-empty":           {input: IPAddr{10, 10}, expected: "10.10.0.0"},
+	}
+	for name, test := range testCases {
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			assert := assertion.New(t)
+			actual := input.String()
+			assert.Equalf(expected, actual, "Testcase %s failed")
```
alextmz, line 46: Thou shalt not uppercase error messages

05_stringer/n0npax/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		"mask 32":              {input: IPAddr{255, 255, 255, 255}, expected: "255.255.255.255"},
+		"mask 0":               {input: IPAddr{0, 0, 0, 0}, expected: "0.0.0.0"},
+		"anz":                  {input: IPAddr{202, 2, 59, 40}, expected: "202.2.59.40"},
+		"empty":                {input: IPAddr{}, expected: "0.0.0.0"},
+		"half-empty":           {input: IPAddr{10, 10}, expected: "10.10.0.0"},
+	}
+	for name, test := range testCases {
+		input, expected := test.input, test.expected
+		t.Run(name, func(t *testing.T) {
+			assert := assertion.New(t)
```
alextmz, line 44: I don't get why using the package as `assertion` when there's no other conflict, but I'm guessing leftovers from old code. Thus line 44 and package rename on import can be disposed of, also pending adding a `t` parameter to `assert.Equalf` below.

PR #192, opened 2019-05-31
Lab2 Bubble & Insertion sort

CameronSchafer:
Fixes #188

Review of colleague's PR #191

#### Changes proposed in this PR:

- Add bubble sort, Insertion sort
- Test bubble sort, Insertion sort

PR #180, opened 2019-05-30
Bubble and insertion sort

nicholascross:
Fixes: #229

Review of colleague's PR #206

#### Changes proposed in this PR:

Using the bubble and insertion sort algorithm to sort an integer slice into ascending order
 - Add simple application to demonstrate sorting algorithms
 - Add unit tests

PR #219, opened 2019-06-02
Implement Lab2 - Bubble and Insertion

### LINKED ISSUE #218: Lab 2 - Bubble and Insertion Sort
Create an executable go program in directory 02_bubble/brvishnu92.
Write function that returns a sorted copy of int slice s using bubble sort.
Write test cases covering all function runs.

brvishnu92:
Fixes #218

Review of colleague's PR #192

Create an executable go program in directory 02_bubble/brvishnu92.
Write function that returns a sorted copy of int slice s using bubble sort.
Write function that returns a sorted copy of int slice s using insertion sort.
Write test cases covering all function runs.

### DIFF
--- a/02_bubble/brvishnu92/main.go
+++ b/02_bubble/brvishnu92/main.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	sorted := make([]int, len(s))
+	copy(sorted, s)
+	swapped := true
+	for swapped {
+		swapped = false
+		for i := 1; i < len(sorted); i++ {
+			if sorted[i] < sorted[i-1] {
+				swapped = true
+				sorted[i-1], sorted[i] = sorted[i], sorted[i-1]
+			}
+		}
+	}
+	return sorted
+}
+
+func insertion(s []int) []int {
+	sorted := make([]int, len(s))
+	copy(sorted, s)
+	for j := 1; j < len(sorted); j++ {
+		i := j - 1
+		inserted := sorted[j]
+		for i >= 0 && sorted[i] > inserted {
+			sorted[i+1] = sorted[i]
+			i--
+		}
+		sorted[i+1] = inserted
+	}
+	return sorted
+}
+
+func main() {
+	fmt.Fprintln(out, insertion([]int{3, 2, 1, 5}))
+}
--- a/02_bubble/brvishnu92/main_test.go
+++ b/02_bubble/brvishnu92/main_test.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBubble(t *testing.T) {
+	testCases := map[string]struct {
+		input    []int
+		expected []int
+	}{
+		"test1":       {input: []int{5, 2, 3, 1}, expected: []int{1, 2, 3, 5}},
+		"test2":       {input: []int{-9, 2, -12, 12}, expected: []int{-12, -9, 2, 12}},
+		"empty":       {input: []int{}, expected: []int{}},
+		"sameval":     {input: []int{1, 1, 1, 1}, expected: []int{1, 1, 1, 1}},
+		"combination": {input: []int{3, 4, 2, -1, 3, 5, 3}, expected: []int{-1, 2, 3, 3, 3, 4, 5}},
+		"oneval":      {input: []int{11}, expected: []int{11}},
+	}
+	for name, test := range testCases {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			result := bubble(test.input)
+			assert.Equalf(t, result, test.expected, "got %v want %v", result, test.expected)
+		})
+	}
+}
+
+func TestInsertion(t *testing.T) {
+	testCases := map[string]struct {
+		input    []int
+		expected []int
+	}{
+		"empty":       {input: []int{}, expected: []int{}},
+		"sameval":     {input: []int{1, 1, 1, 1}, expected: []int{1, 1, 1, 1}},
+		"combination": {input: []int{3, 4, 2, -1, 3, 5, 3}, expected: []int{-1, 2, 3, 3, 3, 4, 5}},
+		"oneval":      {input: []int{11}, expected: []int{11}},
+		"test1":       {input: []int{5, 2, 3, 1}, expected: []int{1, 2, 3, 5}},
+		"test2":       {input: []int{-9, 2, -12, 12}, expected: []int{-12, -9, 2, 12}},
+	}
+	for name, test := range testCases {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			result := insertion(test.input)
+			assert.Equalf(t, result, test.expected, "got %v want %v", result, test.expected)
+		})
+	}
+}
+
+func TestMain(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	expected := "[1 2 3 5]\n"
+	main()
+	actual := buf.String()
+	assert.Equalf(t, actual, expected, "got %v want %v", actual, expected)
+}


02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,45 @@
...
+
+	swap := true
+	for swap {
+		z := s[0]
+		swap = false
+		for i, x := range s {
+			if x < z && i != 0 {
+				swap = true
+				temp := s[i-1]
+				s[i-1], s[i] = s[i], temp
```
nicholascross: You probably don't need the temp variable here.

You can write `s[i-1], s[i] = s[i], s[i-1]` and it will swap the values, the right hand side is evaluated prior to the assignment of those values.

brvishnu92: Oh yeah, thanks for that @nicholascross 👍

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,44 @@
...
+		swap = false
+		for i, x := range s {
+			if x < z && i != 0 {
+				swap = true
+				s[i-1], s[i] = s[i], s[i-1]
+			}
+			z = x
+		}
+	}
+	return s
```
nicholascross: Does this produce a sorted copy or mutate the original array?

02_bubble/brvishnu92/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+		t.Errorf("Unexpected output in main()\nexpected: %q\nactual: %q", expected, actual)
+	}
+}
+
+func TestInsertion(t *testing.T) {
+
+	expected := []int{1, 2, 3, 5}
+	actual := insertion([]int{3, 2, 1, 5})
+
+	if !reflect.DeepEqual(actual, expected) {
```
nicholascross: I actually did a similar thing originally, Julia [suggested]( https://github.com/anz-bank/go-course/pull/180#discussion_r289164436) we should use testify instead.

nicholascross (changes requested):

02_bubble/brvishnu92/main_test.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import (
+	"bytes"
+	"github.com/stretchr/testify/assert"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

02_bubble/brvishnu92/main.go
brvishnu92: Yeah i was mutating it.. My bad... Thanks :-)

02_bubble/brvishnu92/main_test.go
```diff
@@ -0,0 +1,39 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBubble(t *testing.T) {
```
CameronSchafer, line 10: You could do some table tests for both bubble and insertion sorting.

brvishnu92: Sure. added it :-) thanks

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+	}
+	return sort
+}
+
+func insertion(s []int) []int {
+	sort := make([]int, len(s))
+	copy(sort, s)
+	for i := 1; i < len(sort); i++ {
+		j := i
```
nicholascross: I believe you should be able to write the inner loop like a standard `for` loop as you have done above.

brvishnu92: I have assumed u meant the for outer for loop and updated the code. As the inner for loop just checks the index. Hope its what you meant.

nicholascross: Sorry I wasn't not very clear in my description.  What I meant was that you probably do not need to initialise and increment the variables separately.  You can do it as part of the for loop.

```go
for j := i; j > 0; j-- {
	if sort[j-1] > sort[j] {
		sort[j-1], sort[j] = sort[j], sort[j-1]
	}
}
```

I think that will achieve the same thing

nicholascross (changes requested):
Looks good but I think you can still simplify your loop logic.

02_bubble/brvishnu92/main.go
brvishnu92: Oh got it :-)  Thanks

nicholascross (dismissed):

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	sort := make([]int, len(s))
```
juliaogris: keep verbs for function names. should be `sorted`

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,46 @@
...
+	copy(sort, s)
+	swap := true
```
juliaogris: `swapped`

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,46 @@
...
+	for swap {
+		z := sort[0]
+		swap = false
+		for i, x := range sort {
```
juliaogris: This seems to always resort already sorted parts of the result slice.
Seeing as there is a bit of magic going on with indices, it is easier to just stick with them rather than index and assign values IMO:

```go
func bubble(s []int) []int {
        s2 := // make copy of s
	for i := len(s2); i > 0; i-- {
		swapped := false
		for j := 1; j < i; j++ {
                         // swap if s2[j-1] is bigger than s2[j] and set swapped to true
		}
		if !swapped {
			return s2
		}
	}
	return s2
}
```

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func insertion(s []int) []int {
+	sort := make([]int, len(s))
+	copy(sort, s)
+	for i := range sort {
+		for j := i; j > 0; j-- {
+			{
+				if sort[j-1] > sort[j] {
+					sort[j-1], sort[j] = sort[j], sort[j-1]
```
juliaogris: this is bubble sort i believe.
insertion sort looks at its core something like this (you find the right spot for an element by shifting the rest):
```go
	for j := 1; j < len(s2); j++ {
		i := j - 1
		inserted := s2[j]
		for i >= 0 && s2[i] > inserted {
			s2[i+1] = s2[i]
			i--
		}
		s2[i+1] = inserted
	}
```

02_bubble/brvishnu92/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	testCases := map[string]struct {
+		input    []int
+		expected []int
+	}{
+		"test1": {input: []int{5, 2, 3, 1}, expected: []int{1, 2, 3, 5}},
```
juliaogris: add empty slice, single element and slice with multiple elements of same value please.

juliaogris (changes requested):
i think insertion sort needs an update.

02_bubble/brvishnu92/main.go
brvishnu92: Sure, will keep it in mind in future :-)

02_bubble/brvishnu92/main_test.go
brvishnu92: Added these corner test cases 👍

02_bubble/brvishnu92/main.go
brvishnu92: Oh yeah, somehow got messed up :-D

brvishnu92:
@juliaogris I have updated the insertion sort :-)

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,49 @@
...
+	copy(sorted, s)
+	swapped := true
+	z := 0
+	if len(s) > 0 {
+		z = sorted[0]
+	}
+	for swapped {
+		swapped = false
+		for i, x := range sorted {
+			if x < z && i != 0 {
```
juliaogris: Don't compare loop index inside loop, but in your case start iterating from `i := 1`, this way you can also get rid of `if len(s) > 0 {` ...

```go
	for swapped {
		swapped = false
		for i := 1; i < len(sorted); i++ {
			if sorted[i] < sorted[i-1] {
				swapped = true
				sorted[i-1], sorted[i] = sorted[i], sorted[i-1]
			}
		}
	}
```

I see that wikipedia suggests exactly your approach, so I can't argue that this is incorrect. 
For completeness, IMO the most basic bubble sort algorithm looks like this:
```go
	for i := len(s); i > 0; i-- {
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
			}
		}
	}
```
see how the inner loop only iterates until `i`? this is because after the first iteration the largest element will have "bubbled up" to the right most spot in the array - there is no more need to compare anything against this biggest element again.

This can then be optimised to break if no swap happens at all (ideal for already sorted input):
```go
	for i := len(s); i > 0; i-- {
		swapped := false
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
				swapped = true
			}
		}
		if !swapped {
			return
		}
	}
```
IMO this is not what you have implemented ;)

Either case, please remove the loop index comparison inside the loop.

02_bubble/brvishnu92/main.go
```diff
@@ -0,0 +1,49 @@
...
+				swapped = true
+				sorted[i-1], sorted[i] = sorted[i], sorted[i-1]
+			}
+			z = x
+		}
+	}
+	return sorted
+}
+
+func insertion(s []int) []int {
```
juliaogris, line 27: 👍

02_bubble/brvishnu92/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+func TestMain(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	expected := "[1 2 3 5]\n"
+	main()
+	actual := buf.String()
+
+	if actual != expected {
```
juliaogris: use `assert.Equalf` here as well.

juliaogris (dismissed):
I think you should slightly update your bubble sort algorithm and modify one test.

Also, please fold your fixup commits into your initial commits with rebase.
If you have difficulties doing this, please reach out.

02_bubble/brvishnu92/main.go
brvishnu92: Done. I was fixated on not having a loop with the index like (for i:=1) . Hence all the workaround on it :-) Fixed now.

02_bubble/brvishnu92/main_test.go
brvishnu92: Done.

CameronSchafer:
All changes that have been request have been acted upon and the changes look good. So I am changing the label of this PR to approved.

anzboi (approved):
LGTM

PR #224, opened 2019-06-02
lab 3 completed with ut

### LINKED ISSUE #269: Lab3
- [ ] Create an executable go program in directory `03_letters/luliangLi`
- [ ] Write a function that returns a mapping of each letter to its frequency:
```
```

- [ ] Write a function that returns a sorted slice of strings with elements `"{key}:{val}"`. Use package [sort](https://golang.org/pkg/sort/):

```
```
- [ ] Call `fmt.Println(strings.Join(sortLetters(letters("aba")), "\n"))` in `main` to print:
```
a:2
b:1
```
- [ ] Bonus points: comprehensive tests

luliangLi:
Fixes #269

Review of colleague's PR #209 

- Implement Lab3 Letter frequency
- Test Lab3

### DIFF
--- a/03_letters/luliangLi/main.go
+++ b/03_letters/luliangLi/main.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+type Pair struct {
+	key   rune
+	value int
+}
+
+func (p Pair) String() string {
+	return fmt.Sprintf("%s:%d", string(p.key), p.value)
+}
+
+func letters(s string) map[rune]int {
+	rlt := make(map[rune]int)
+
+	for _, char := range s {
+		val := rlt[char]
+		rlt[char] = val + 1
+	}
+
+	return rlt
+}
+
+func sortLetters(m map[rune]int) []string {
+	pairs := make([]Pair, len(m))
+	rlt := make([]string, len(m))
+
+	i := 0
+	for k, v := range m {
+		pairs[i] = Pair{k, v}
+		i++
+	}
+
+	sort.Slice(pairs, func(i, j int) bool {
+		return pairs[i].value > pairs[j].value
+	})
+
+	j := 0
+	for _, pair := range pairs {
+		rlt[j] = pair.String()
+		j++
+	}
+
+	return rlt
+}
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
--- a/03_letters/luliangLi/main_test.go
+++ b/03_letters/luliangLi/main_test.go
@@ -1,0 +1,48 @@
+package main
+
+import (
+	"bytes"
+	"reflect"
+	"strconv"
+	"testing"
+)
+
+func TestSortLetters(t *testing.T) {
+
+	testCases := map[string]struct {
+		input string
+		want  []string
+	}{
+		"normal": {input: "aba", want: []string{"a:2", "b:1"}},
+		"blank":  {input: "", want: []string{}},
+		"long":   {input: "aabbcccdddd", want: []string{"d:4", "c:3", "a:2", "b:2"}},
+		"single": {input: "aaaa", want: []string{"a:4"}},
+	}
+
+	for name, test := range testCases {
+		// t.Run creates a sub test and runs it like a normal test
+		test := test
+
+		t.Run(name, func(t *testing.T) {
+			actual := sortLetters(letters(test.input))
+
+			if !reflect.DeepEqual(test.want, actual) {
+				t.Errorf("running : %v, expected %v, got %v", test.input, test.want, actual)
+			}
+		})
+	}
+}
+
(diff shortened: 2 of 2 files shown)


03_letters/luliangLi/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"bytes"
+	"reflect"
+	"strconv"
+	"testing"
+)
+
+func TestSortLetters(t *testing.T) {
+
+	testCases := map[string]struct {
```
CameronSchafer, line 12: You could write a test for letters in a sentence. That uses spaces and punctuation.

03_letters/luliangLi/main.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, char := range s {
+		val := rlt[char]
+		rlt[char] = val + 1
+	}
+
+	return rlt
+}
+
+func sortLetters(m map[rune]int) []string {
+	pairs := make([]Pair, len(m))
```
CameronSchafer, line 34: What is the main purpose of using the `Pair` struct that you have created? 
And could you just use the `map[rune]int` for the same purpose?

03_letters/luliangLi/main.go
```diff
@@ -0,0 +1,58 @@
...
+	rlt := make([]string, len(m))
+
+	i := 0
+	for k, v := range m {
+		pairs[i] = Pair{k, v}
+		i++
+	}
+
+	sort.Slice(pairs, func(i, j int) bool {
```
CameronSchafer, line 43: Here you could use the `sort.Ints()` method for the same purpose.

juliaogris (changes requested):
Please read through the [Collected Code Review Comments](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments)

- [Fix your PR description](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#pr-description)

03_letters/luliangLi/main.go
```diff
@@ -0,0 +1,58 @@
...
+func (p Pair) String() string {
+	return fmt.Sprintf("%s:%d", string(p.key), p.value)
+}
+
+func letters(s string) map[rune]int {
+	rlt := make(map[rune]int)
+
+	for _, char := range s {
+		val := rlt[char]
+		rlt[char] = val + 1
```
rokane, line 27: Nice 👍  Can achieve this without having to store another variable on the stack:
```go
rlt[char]++
```

03_letters/luliangLi/main.go
rokane: Agree with @CameronSchafer here, using the `Pair` struct means you duplicate the looping logic. You can achieve the same outcome by directly creating a `[]string` from the `map[rune]int`.
```go
for k, v := range m {
  output = append(output, fmt.Sprintf("%c:%d", k, v))
}
```

03_letters/luliangLi/main.go
```diff
@@ -0,0 +1,58 @@
...
+		pairs[i] = Pair{k, v}
+		i++
+	}
+
+	sort.Slice(pairs, func(i, j int) bool {
+		return pairs[i].value > pairs[j].value
+	})
+
+	j := 0
+	for _, pair := range pairs {
```
rokane, line 48: This loop shouldn't be required when you remove the `Pair` struct, but with this, you should be able to use the `index` value from the `range` result, rather than initialising a new `j` variable.
```go
for i, pair := range pairs {
  rlt[i], pair.String()
}
```

03_letters/luliangLi/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("a:2\nb:1\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("running main, expected %v, got %v", expected, actual)
```
rokane, line 46: Good work!!

03_letters/luliangLi/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+		"long":   {input: "aabbcccdddd", want: []string{"d:4", "c:3", "a:2", "b:2"}},
+		"single": {input: "aaaa", want: []string{"a:4"}},
+	}
+
+	for name, test := range testCases {
+		// t.Run creates a sub test and runs it like a normal test
+		test := test
+
+		t.Run(name, func(t *testing.T) {
+			actual := sortLetters(letters(test.input))
```
rokane, line 27: I think it is best to test your functions in isolation, that way if you get a failure, you can be sure what function is causing the underlying issue. Testing the concatenation of the two functions is sort of achieved in the main anyway, which acts as your integration test.

Please break this down into 2 tests.
```go
func TestSortLetters(t *testing.T) {
    ...
}

func TestLetters(t *testing.T) {
    ...
}
```

rokane (changes requested):
Looking good so far. A couple of things which need addressing and this should be good!

juliaogris:

PR #227, opened 2019-06-03
lab1 - Created Fibonacci sequencer

### LINKED ISSUE #324: Lab1 - Fibonacci
Implementation of Fibonacci and Negafibonacci starting from 0.

joel00wood:
Built fib(n int) as 3 functions. The base function `fibCalc`
calculates the Nth fibonacci value. The wrapper function `fibN` enables
support for Negafibonacci calculation. `fib` itself just prints 0 to
(+|-)n, on each line.

Borrowed output capture function from elsewhere to
enable testing of `fib` and `main` that do not return values.

Fixes #324

Review of colleague's PR #321

- Introduce `fib(n int)` function to print fibonacci numbers from 0 to (+/-)n

### DIFF
--- a/01_fib/joel00wood/main.go
+++ b/01_fib/joel00wood/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"math"
+)
+
+func fibCalc(n float64) float64 {
+	switch n {
+	case 0, 1:
+		return n
+	default:
+		return fibCalc(n-1) + fibCalc(n-2)
+	}
+}
+
+func fibN(n float64) float64 {
+	if n < 0 {
+		var i float64
+		n := math.Abs(n)
+		i = math.Pow(-1, n+1)
+		return fibCalc(n) * i
+	}
+	return fibCalc(n)
+}
+
+func fib(n int) {
+	for i := 0.0; i <= math.Abs(float64(n)); i++ {
+		if n < 0 && i > 0 {
+			fmt.Println(fibN(i * -1))
+		} else {
+			fmt.Println(fibN(i))
+		}
+	}
+}
+
+func main() {
+	fib(7)
+}
--- a/01_fib/joel00wood/main_test.go
+++ b/01_fib/joel00wood/main_test.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"testing"
+
+	output "github.com/joel00wood/test-helpers/capture"
+)
+
+func TestFibCalc(t *testing.T) {
+	testCases := map[string]struct {
+		input, expected float64
+	}{
+		"input 20": {20, 6765},
+		"input 3":  {3, 2},
+		"input 2":  {2, 1},
+		"input 1":  {1, 1},
+		"input 0":  {0, 0},
+	}
+
+	for name, test := range testCases {
+		actual := fibCalc(test.input)
+		if test.expected != actual {
+			t.Errorf("tests[%s] fibCalc(%f) wrong. expected=%f, got=%f",
+				name, test.input, test.expected, actual)
+		}
+	}
+}
+
+func TestFibN(t *testing.T) {
+	testCases := map[string]struct {
+		input, expected float64
+	}{
+		"input 20":  {20, 6765},
+		"input 3":   {3, 2},
+		"input 2":   {2, 1},
+		"input 1":   {1, 1},
+		"input 0":   {0, 0},
+		"input -1":  {-1, 1},
+		"input -2":  {-2, -1},
+		"input -3":  {-3, 2},
+		"input -20": {-20, -6765},
+	}
+
+	for name, test := range testCases {
+		actual := fibN(test.input)
+		if test.expected != actual {
+			t.Errorf("tests[%s] fibN(%f) wrong. expected=%f, got=%f",
+				name, test.input, test.expected, actual)
+		}
+	}
+}
+
+func TestFib(t *testing.T) {
+	testCases := map[string]struct {
+		input    int
+		expected string
+	}{
+		"input 7":  {7, "0\n1\n1\n2\n3\n5\n8\n13\n"},
+		"input 3":  {3, "0\n1\n1\n2\n"},
+		"input 2":  {2, "0\n1\n1\n"},
+		"input 1":  {1, "0\n1\n"},
+		"input 0":  {0, "0\n"},
+		"input -1": {-1, "0\n1\n"},
+		"input -2": {-2, "0\n1\n-1\n"},
+		"input -3": {-3, "0\n1\n-1\n2\n"},
+		"input -7": {-7, "0\n1\n-1\n2\n-3\n5\n-8\n13\n"},
+	}
+	for name, test := range testCases {
+		input := test.input
+		actual := output.CaptureOutput(func() { fib(input) })
+		if test.expected != actual {
+			t.Errorf("tests[%s] fib(%d) wrong. expected=%q, got=%q",
+				name, test.input, test.expected, actual)
+		}
+	}
+}
+
+// test main() which is essentially fib(7)
+func TestMain(t *testing.T) {
+	expected := "0\n1\n1\n2\n3\n5\n8\n13\n"
+	actual := output.CaptureOutput(func() { main() })
+	if expected != actual {
+		t.Errorf("Unexpected response for input main(){fib(7)}\nExpected: %q\nActual: %q",
+			expected, actual)
+	}
+}


01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,50 @@
...
+	"fmt"
+	"math"
+)
+
+func isWholeNumber(n float64) bool {
+	return n == math.Trunc(n)
+}
+
+func fibCalc(n float64) float64 {
+	if n == 0 {
```
golangcibot: ifElseChain: rewrite if-else to switch statement (from `gocritic`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+		"input 1": {1, "0\n1\n"},
+		"input 0": {0, "0\n"},
+		"input -1": {-1, "0\n1\n"},
+		"input -2": {-2, "0\n1\n-1\n"},
+		"input -3": {-3, "0\n1\n-1\n2\n"},
+		"input -7": {-7, "0\n1\n-1\n2\n-3\n5\n-8\n13\n"},
+		"input -1.5": {-1.5, "Invalid input, expected whole number: -1.5\n"},
+	}
+	for name, test := range testCases {
+		actual := captureOutput(func() {fib(test.input)})
```
golangcibot: Using the variable on range scope `test` in function literal (from `scopelint`)

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,50 @@
...
+	} else {
+		return fibCalc(n-1) + fibCalc(n-2)
+	}
+}
+
+func fibN(n float64) float64 {
+	if n < 0 {
+		var i float64
+		n := math.Abs(n)
+		i = math.Pow(-1, n + 1)
```
golangcibot: File is not `goimports`-ed (from `goimports`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+	go func() {
+		var buf bytes.Buffer
+		wg.Done()
+		io.Copy(&buf, reader)
+		out <- buf.String()
+	}()
+	wg.Wait()
+	f()
+	writer.Close()
+	return <- out
```
golangcibot: File is not `goimports`-ed (from `goimports`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+	os.Stdout = writer
+	os.Stderr = writer
+	log.SetOutput(writer)
+	out := make(chan string)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		var buf bytes.Buffer
+		wg.Done()
+		io.Copy(&buf, reader)
```
golangcibot: Error return value of `io.Copy` is not checked (from `errcheck`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"bytes"
+	"io"
+	"log"
+	"os"
+	"sync"
+	"testing"
+)
+
+// enter borrowed code
+// courtesy of: https://gist.github.com/hauxe/e935a7f9012bf2649710cf75af323dbf (typed manually so should probably validate)
```
golangcibot: line is 123 characters (from `lll`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+		"input 1": {1, true},
+		"input 1.5": {1.5, false},
+		"input 0.5": {0.5, false},
+		"input 1000000000": {1000000000, true},
+	}
+
+	for name, test := range testCases {
+		actual := isWholeNumber(test.input)
+		if test.expected != actual {
+			t.Errorf("Unexpected response for %s iwWholeNumber(%f)\nExpected: %t\nActual: %t", name, test.input, test.expected, actual)
```
golangcibot: line is 126 characters (from `lll`)

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,50 @@
...
+		return fibCalc(n) * i
+	} else {
```
golangcibot: `if` block ends with a `return` statement, so drop this `else` and outdent its block (from `golint`)

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"fmt"
+	"math"
+)
+
+func isWholeNumber(n float64) bool {
```
runnerdave: I would suggest to use integers, thus you won't need this function, and it the ending program will be simpler as well

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,147 @@
...
+		if test.expected != actual {
+			t.Errorf("Unexpected response for %q fib(%f)\nExpected: %q\nActual: %q", name, test.input, test.expected, actual)
+		}
+	}
+}
+
+
+// test main() which is essentially fib(7)
+func TestMain(t *testing.T) {
+	expected := "0\n1\n1\n2\n3\n5\n8\n13\n"
```
runnerdave, line 80: You can use:
```
expected := `0
1
1
.
`
```
And then filter out the \n with strings.Replace in the actual:
`actual := captureOutput(strings.Replace(func() {main()}, "\n", " ", -1))`

runnerdave (commented):
Good start, but it can be simplified, left some suggestions, especially consider just using integers, would help you also running:` golangci-lint run` before you push your repo

joel00wood:
@runnerdave haha yeah, I had yet to properly install `golangci-lint` locally due to some issues with running the `go get` for it on my dev machine. 

Primary reason for raising this as a draft was to mediate that until I had time to properly fix it and thus have it actually prepared for review and labelled accordingly. Will look into your feedback points all the same though!

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+}
+
+func TestFib(t *testing.T) {
+	testCases := map[string]struct {
+		input    float64
+		expected string
+	}{
+		"input 7":    {7, "0\n1\n1\n2\n3\n5\n8\n13\n"},
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+		"input 1":    {1, "0\n1\n"},
+		"input 0":    {0, "0\n"},
+		"input -1":   {-1, "0\n1\n"},
+		"input -2":   {-2, "0\n1\n-1\n"},
+		"input -3":   {-3, "0\n1\n-1\n2\n"},
+		"input -7":   {-7, "0\n1\n-1\n2\n-3\n5\n-8\n13\n"},
+	}
+	for name, test := range testCases {
+		input := test.input
+		actual := captureOutput(func() { fib(input) })
```
golangcibot: cannot use input (variable of type float64) as int value in argument to fib (from `typecheck`)

01_fib/joel00wood/main.go
joel00wood: Switched for integers as discussed and removed no longer necessary function

01_fib/joel00wood/main_test.go
joel00wood: In this case I'm happy to use the `\n` characters as it reads better and is easier to manage in the test framework

01_fib/joel00wood/main_test.go
joel00wood: Fixed for testCase struct switching `float64` for `int`

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"math"
+)
+
+func fibCalc(n float64) float64 {
+	switch n {
+	case 0:
```
pentaphobe: nitpick: stylistically I'd lean more towards either having an early out conditional:

```go
if n <= 1 {
  return n
}
```

or combining your first two cases:

```go
switch n {
    case 0, 1:
        return n
/* ... etc ... */
```

pentaphobe (approved):
Looks good to me, though I agree it'd be worth refactoring to just use `int` - this'd simplify your code without losing anything in the process

(Nice tests)

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,41 @@
...
+		return n
+	case 1:
+		return n
+	default:
+		return fibCalc(n-1) + fibCalc(n-2)
```
pentaphobe, line 13: I had the same compulsion to do it recursively, however as a thought experiment it's worth considering what the "nicest" way to do it non-recursively is

Partially because in this problem space, the desire is to store all fib numbers up to `n` so you're throwing performance out the window for recursive niceness - but also because it's just good practice :D

01_fib/joel00wood/main.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+}
+
+func fibN(n float64) float64 {
+	if n < 0 {
+		var i float64
+		n := math.Abs(n)
+		i = math.Pow(-1, n+1)
```
pentaphobe, line 21: this is definitely the most creative way I've seen to achieve this.

can you think of a less opaque way?  say, perhaps using bitwise operations?

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,125 @@
...
+	"log"
+	"os"
+	"sync"
+	"testing"
+)
+
+// enter borrowed code
+// https://gist.github.com/hauxe/e935a7f9012bf2649710cf75af323dbf
+// stdOut capture helper
+func captureOutput(f func()) string {
```
pentaphobe: I didn't mention this before, but it makes me really happy to see someone avoiding the global `out := os.Stdout` line and doing this "properly"

Given you're using it in all your labs, perhaps this is a good opportunity to explore local packages?

pentaphobe (dismissed):
LGTM, just a few comments

Tempted to "request changes" given the use of floats here is arguably bad practice (ie. unnecessary complexity, inappropriate type, extra memory/CPU strain) but that feels too nitpicky

01_fib/joel00wood/main.go
joel00wood: Seems reasonable to me, I think I did attempt option two initially but for whatever reason it didn't work so I stuck with the expanded approach

joel00wood: Updated

joel00wood:
Made the stylistic change regarding the switch statement, other pieces may take me a bit longer to address

juliaogris (dismissed):
I'll approve of this now, the lab is done and you _really_ have gone the extra mile (PR description, capturing streams, tests) 👍 .

You might find value in looking at our collaborate, canonical solution:
```go
func fib(n int) {
	sign := 1
	if n < 0 {
		sign = -1
	}

	n1, n2 := 0, 1
	for j := 0; j < n*sign; j++ {
		n1, n2 = n2, n1+sign*n2
		fmt.Fprintln(out, n1)
	}
}
```

juliaogris (approved):
Lgtm

juliaogris (changes requested):
Please fix linter warnings :)

joel00wood:
Seems it's upset about my re-use of `captureOutput()` so will have to find a different approach

01_fib/joel00wood/main_test.go
```diff
@@ -0,0 +1,85 @@
+package main
+
+import (
+	"github.com/joel00wood/test-helpers/capture"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

juliaogris (approved):
lgtm

PR #232, opened 2019-06-03
Lab1

ranjanasj:
Fixes #

Review of colleague's PR #


01_fib/anzranjana/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+var out io.Writer = os.Stdout
+
+func Fib(number int) {
+	firstNum, nextNum := int64(0), int64(1)
+	switch number {
+	case 0:
+		fmt.Fprintln(out, number)
+	default:
+		if number < 0 { // Negative Fibonacci series
```
brnbke: The logic here looks pretty similar for both the positive and negative Fibonacci numbers, can you abstract this out somehow to avoid duplication?

01_fib/shahriys/main.go
```diff
@@ -0,0 +1,51 @@
+package main
```
brnbke, line 1: Should shahriy's two files be in you PR?

01_fib/anzranjana/main_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	expected := "1\n1\n2\n3\n5\n8\n13\n"
+	actual := buf.String()
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestNegativeFibo(t *testing.T) {
```
brnbke, line 22: Did you look at https://goplay.space/#Zy2ta-z8vrO - good example of testing multiple cases rather than having to release the setup.

brnbke (changes requested):

01_fib/shahriys/main.go
ranjanasj: No, i dont understand how it got added

raghunadhpokkalath (changes requested):
Can you please do a git rebase.Pull request shows 4 files as changed

juliaogris:

PR #235, opened 2019-06-03
Lab3

### LINKED ISSUE #234: Lab3
- [ ] Create an executable go program in directory **03_letters**
- [ ] Write a function that returns a mapping of each letter to its frequency:
```go
```
- [ ] Write a function that returns a sorted slice of strings with elements "{key}:{val}". Use package [sort:](https://golang.org/pkg/sort/)
```go
```
Call **print(letters("aba"))** in **main** to print:
```go
a:2
b:1
```
- [ ] Bonus points: comprehensive tests

CameronSchafer:
Fixes #234

Review of colleague's PR #224

- Add letter sort functionality
- Add unit tests


01_fib/kasunfdo/main.go
```diff
@@ -1,36 +0,0 @@
-package main
```
lexual: I believe (I hope I'm not wrong), the idea is that once each persons lab implementation is approved, it gets merged into master branch, and so it will show up in your future branches.

If we merge your branch now, this persons work will be deleted.

I suggest you rebase and drop this commit, that drops this other contributors code.

03_letters/CameronSchafer/letters.go
```diff
@@ -0,0 +1,43 @@
...
+
+var out io.Writer = os.Stdout
+
+// string to rune with counts
+func letters(s string) map[rune]int {
+	tMap := make(map[rune]int)
+
+	for i := range s {
+		tMap[rune(s[i])]++
+	}
```
lexual: You could consider??:

```
for _, r := range s {
    tMap[r]++
}
```

03_letters/CameronSchafer/letters.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+// string to rune with counts
+func letters(s string) map[rune]int {
+	tMap := make(map[rune]int)
```
lexual: You could consider just having 't' as variable name, instead of including type in the variable. Same with tSlice/etc below.

03_letters/CameronSchafer/letters.go
```diff
@@ -0,0 +1,43 @@
...
+
+	for i := range s {
+		tMap[rune(s[i])]++
+	}
+
+	return tMap
+}
+
+func sortLetters(m map[rune]int) []string {
+	tSlice := []int{}
```
lexual: Would keys be a better variable name for tSlice?

i in the for-loop could be k, making it clearer we are looking over keys, and not array/slice indices.

03_letters/CameronSchafer/letters.go
```diff
@@ -0,0 +1,43 @@
...
+	for i := range m {
+		tSlice = append(tSlice, int(i))
+	}
+
+	// sort the ints
+	sort.Ints(tSlice)
+
+	sSlice := []string{}
+	for _, val := range tSlice {
+		sSlice = append(sSlice, fmt.Sprintf("%c:%d", val, m[rune(val)]))
```
lexual: Nice use of "%c", I couldn't find how to print out my rune correctly like this, and my solution was a bit less elegant ;)

03_letters/CameronSchafer/letters_test.go
```diff
@@ -0,0 +1,88 @@
...
+		{description: "sortLetters [97:7, 98:7]", input: map[rune]int{97: 7, 98: 7},
+			expectedArr: []string{"a:7", "b:7"},
+		},
+		{description: "sortLetters [32:2, 97:1]", input: map[rune]int{32: 2, 97: 1},
+			expectedArr: []string{" :2", "a:1"},
+		},
+	}
+
+	for _, test := range testCases {
+		test := test
```
lexual: I don't understand why this line is necessary, or what it is doing ;|

03_letters/CameronSchafer/letters_test.go
```diff
@@ -0,0 +1,88 @@
...
+		// t.Run creates a sub test and runs it like a normal test
+		t.Run(test.description, func(t *testing.T) {
```
lexual: Nice use of t.Run, I hadn't come across. I'm doing table-driven tests, but not using Run. I'll have to look at if for future labs ;)

03_letters/CameronSchafer/letters_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+// sortLetters()
+func TestSortLetters(t *testing.T) {
+	// test cases with descriptions.
+	testCases := []struct {
+		description string
+		input       map[rune]int
+		expectedArr []string
+	}{
+		{description: "sortLetters [97:2, 98:1]", input: map[rune]int{97: 2, 98: 1},
```
lexual: Isn't going to be easier to understand the tests if you use a rune literal, instead of the integer representation.

e.g. replace, 97 with 'a', etc, etc.

lexual:
Nice work @CameronSchafer .

I've made some comments, but they all pretty minor points, not all of my advice needs to be followed ;)

03_letters/CameronSchafer/letters_test.go
lexual: Nevermind, I've subsequently had this explained to me, its to do with scoping.

01_fib/kasunfdo/main.go
CameronSchafer: I think i fixed this by rebasing my branch :)

03_letters/CameronSchafer/letters_test.go
CameronSchafer: It is super useful if you need to run multiple tests through the same function

03_letters/CameronSchafer/letters_test.go
CameronSchafer: I updated the test cases to the rune literals. I didn't even know I could do that! 👍

03_letters/CameronSchafer/letters.go
CameronSchafer: I updated the name for this variable. It definitely looks neater with a shorter name.

03_letters/CameronSchafer/letters.go
CameronSchafer: I changed it to this and it looks much better! thank you 👍

03_letters/CameronSchafer/letters_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"testing"
+)
+
+// main() test
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+	expected := "a:2\nb:1\n"
```
brvishnu92: Could make it neater by using `string.replace` in your output or raw strings.

03_letters/CameronSchafer/letters_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()\nexpected: %q\nactual: %q", expected, actual)
+	}
+}
+
+// sortLetters()
+func TestSortLetters(t *testing.T) {
+	// test cases with descriptions.
+	testCases := []struct {
```
brvishnu92: Consider switching to map instead of struct and having key as testname.

brvishnu92 (changes requested):

03_letters/CameronSchafer/letters_test.go
CameronSchafer: Done

03_letters/CameronSchafer/letters_test.go
CameronSchafer: Done

brvishnu92 (dismissed):

03_letters/CameronSchafer/main.go
```diff
@@ -0,0 +1,43 @@
...
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+// string to rune with counts
```
rokane: General go convention when adding comments to functions is that they start with the name of the function and then an explanation. This was the comment is picked up in the godoc.

```go
// letters returns a count of each letter in the given string
func letters(s string) []string {
    ...
}
```

Can you update this?

03_letters/CameronSchafer/main.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func sortLetters(m map[rune]int) []string {
+	var keys = []int{}
+	for val := range m {
+		keys = append(keys, int(val))
+	}
+
+	// sort the ints
+	sort.Ints(keys)
```
rokane: A couple of things here. I don't think this comment adds any value, it is self explanatory in the method name what the call is doing.

And secondly, you can eliminate the first iteration over the map, if you directly create a `[]string` as you have below. Then you should be able to directly sort it using `sort.Strings` to achieve the same result with a single iteration.

Can you please update this?

03_letters/CameronSchafer/main_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// main() test
+func TestMainOutput(t *testing.T) {
```
rokane, line 10: Again here, I don't see these comments adding any value. The name of the test is documentation in itself. Please remove for all these

03_letters/CameronSchafer/main_test.go
```diff
@@ -0,0 +1,101 @@
...
+		"letters with spaces": {input: "a a ",
+			expectedArr: map[rune]int{' ': 2, 'a': 2},
+		},
+		"AaAa}{}{}}{}{}{": {input: "AaAa}{}{}}{}{}{",
+			expectedArr: map[rune]int{'A': 2, 'a': 2, '{': 5, '}': 6},
+		},
+		"empty string": {input: "",
+			expectedArr: map[rune]int{},
+		},
+		"only spaces": {input: "    ",
```
rokane, line 76: You can also add a case for special characters.
```
"special characters": {
			input: "无😊👍👍",
			expectedArr:  map[rune]int{ '👍': 2, '无': 1, '😊': 1},
		}
```

rokane (dismissed):
Looking good so far, there are a couple of things which need addressing throughout. Also it would be nice to have a little mode description in your commit messages. The title looks good, but you can add further detail in the body of the message to help someone reading who might not have context of the change.

03_letters/CameronSchafer/main.go
CameronSchafer: Done

03_letters/CameronSchafer/main.go
CameronSchafer: Done.
Stumped me for a bit on this one.

03_letters/CameronSchafer/main_test.go
CameronSchafer: Done

03_letters/CameronSchafer/main_test.go
CameronSchafer: Done

camscale (approved):
LGTM. Sorry for the delay with re-reviewing. Thanks for sticking with us.

PR #212, opened 2019-06-02
Go Lab 1 - calculate Fibonacci and Negafibonacci

brnbke:
Fixes #397

Fibonacci and Negafibonacci numbers, should handle numbers from -92 to 92 inclusive, and report when out of this range.

Review of colleague's PR #232

PR #239, opened 2019-06-03
02_bubble Lab2 commit for review

### LINKED ISSUE #237: Lab2 - 02_bubble : Bubble and Insertion Sort
Build lab 2 assignment code for Bubble and Insertion Sort for members of the slice.

saurabhvyas9:
Fixes #237

Review of colleague's PR #212

02_bubble code committed for Bubble and Insertion sort as part of second assignment for go-course.

### DIFF
--- a/02_bubble/saurabhvyas9/main.go
+++ b/02_bubble/saurabhvyas9/main.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	for j := len(r); j > 0; j-- {
+		swapped := false
+		for i := 0; i < j-1; i++ {
+			if r[i] > r[i+1] {
+				r[i], r[i+1] = r[i+1], r[i]
+				swapped = true
+			}
+		}
+		if !swapped {
+			break
+		}
+	}
+	return r
+}
+
+func insertion(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	for i := 1; i < len(r); i++ {
+		temp := r[:i]
+		var j int
+		for j = 0; j <= len(temp)-1; j++ {
+			if temp[j] >= r[i] {
+				break
+			}
+		}
+		tmp := r[i]
+		copy(r[j+1:i+1], r[j:i])
+		r[j] = tmp
+	}
+	return r
+}
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+	fmt.Fprintln(out, insertion([]int{3, 2, 1, 5}))
+}
--- a/02_bubble/saurabhvyas9/main_test.go
+++ b/02_bubble/saurabhvyas9/main_test.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	// Given
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote("[1 2 3 5]\n[1 2 3 5]\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equalf(t, expected, actual, "Sorting failed.")
+}
+
+var testCases = map[string]struct {
+	input []int
+	want  []int
+}{
+	"main":                {input: []int{3, 1, 2, 5}, want: []int{1, 2, 3, 5}},
+	"ascending sequence":  {input: []int{1, 2, 4, 3, 5}, want: []int{1, 2, 3, 4, 5}},
+	"descending sequence": {input: []int{5, 4, 3, 2, 1}, want: []int{1, 2, 3, 4, 5}},
+	"duplicates":          {input: []int{4, 4, 1, 100, 1}, want: []int{1, 1, 4, 4, 100}},
+	"negatives":           {input: []int{-5, -1, -2, 4, 3, 19, 0, 1, 4}, want: []int{-5, -2, -1, 0, 1, 3, 4, 4, 19}},
+	"empty":               {input: []int{}, want: []int{}},
+	"single":              {input: []int{1}, want: []int{1}},
+}
+
+func TestBubbleSort(t *testing.T) {
+	for _, test := range testCases {
+		input := test.input
+		expected := test.want
+		sorted := bubble(input)
+		assert.Equalf(t, expected, sorted, "Bubble sort failed")
+	}
+}
+
+func TestInsertionSort(t *testing.T) {
+	for _, test := range testCases {
+		input := test.input
+		expected := test.want
+		sorted := insertion(input)
+		assert.Equalf(t, expected, sorted, "Insertion sort failed")
+	}
+}


pfannerj (dismissed):
The insertion sort has a bug which is shown if you input {3, 2, 1, 5, 4} and it returns [1 2 3 5 4]...

Suggest including more test cases to test all the edge cases for both the bubble and insertion sorts.

02_bubble/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Then
+	expected := strconv.Quote("[1 2 3 5][1 2 3 5]")
+	actual := strconv.Quote(buf.String())
+	if expected != actual {
+		t.Errorf("Unexpected output in main()\nexpected: %s\nactual: %s", expected, actual)
+	}
+}
+
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

pfannerj: Tests look good now!

pfannerj:
You probably just need to use rebase to reduce your commits... just refer to this checklist for Lab2 and you'll be all good.  If u need help with rebase for your commits I should be able to help because I'm becoming an expert :o - I had to do lots of squashing and fixups myself for Lab1 & Lab2 :)

pfannerj:
Forgot to paste the link with the checklist - https://github.com/anz-bank/go-course/wiki/Lab-2:-Collected-Code-Review-Comments

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	r := s
+	chckcount := 1
```
juliaogris: It seems like you look `chkcount` as a bool so you should probably use a more expressive variable name for a bool, something like:
`swapped := true` 

Aside, `chkcount` probably means `checkCount` which to me is a function name not variable name. The variable name would be just `count` or `hasCount` or `countChecked` or so...

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	r := s
```
juliaogris: This assignment doesn't make a copy of `s` please write a test for it and then fix it with `copy`.
The test should have an `input` array say `[2,3,1]` 
after running `outpu := bubble(input)` input should still be `[2,3,1]`, output should be `[1,2,3]`

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,42 @@
...
+				r[i], r[i+1] = r[i+1], r[i]
+				chckcount++
+			}
+		}
+	}
+	return r
+}
+
+func insertion(s []int) []int {
+	r := s
```
juliaogris: not a copy

02_bubble/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,68 @@
...
+	if expected != actual {
+		t.Errorf("Unexpected output in main()\nexpected: %s\nactual: %s", expected, actual)
+	}
+}
+
+var testCases = map[string]struct {
+	input []int
+	want  []int
+}{
+	"TC1": {input: []int{3, 1, 2, 5}, want: []int{1, 2, 3, 5}},
```
juliaogris: nice! love the table driven test.
Maybe choose more meaningful names: e.g. `Simple`, `Sorted`, `Duplicates`...
Also add single element and empty slices please.

02_bubble/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,68 @@
...
+		sorted := insertion(input)
+		if !Equal(sorted, test.want) {
+			t.Errorf("Sorting failed. Input: %v, Sorted: %v.", input, sorted)
+		}
+	}
+}
+
+// Equal tells whether a and b contain the same elements.
+// A nil argument is equivalent to an empty slice.
+func Equal(a, b []int) bool {
```
juliaogris: Nicely done - however, `testify/assert` has got and `assert.Equal` that does it for you out of the box with great diff-printing if the test fails.
No need to update here, for the next lab.

juliaogris (changes requested):
A few comments - please also fix up your [commit messages](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages).

02_bubble/saurabhvyas9/main_test.go
saurabhvyas9: Done

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Updated. Not using bool.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Updated the code. Now making a copy of it. 
Done.

02_bubble/saurabhvyas9/main_test.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	countChecked := 1
```
juliaogris: Which count are you checking?
Typically this variable is called `swapped` of type bool: you flag if you have swapped any elements in bubble sort and if you haven't, the whole slice is already sorted so you can break.

I see that wikipedia suggests exactly your approach, so I can't argue that this is incorrect. 
For completeness, IMO the most basic bubble sort algorithm looks like this:
```go
	for i := len(s); i > 0; i-- {
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
			}
		}
	}
```
see how the inner loop only iterates until `i`? this is because after the first iteration the largest element will have "bubbled up" to the right most spot in the array - there is no need to compare anything against this biggest element again. In the next iteration the two rightmost elements will have found their final position on so forth.

This can then be optimised to break if no swap happens at all (ideal for already sorted input):
```go
	for i := len(s); i > 0; i-- {
		swapped := false
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
				swapped = true
			}
		}
		if !swapped {
			return
		}
	}
```

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,44 @@
...
+			}
+		}
+	}
+	return r
+}
+
+func insertion(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	for i := 1; i < len(r); i++ {
```
juliaogris, line 32: It took me a while to get my head around this - but it does seem to do insertion sort:
build up a sorted array from the left to the right.
However, it does more comparisons than necessary (the second iteration should go from i down and break when the insertion point is found). Also, the `temp` slice is not needed and does not add to readability IMO.

So a first approach to simplify this would be as:
```go
	for i := 1; i < len(r); i++ {
		for j := 0; j < i; j++ {
			if r[j] > r[i] {
				r[j], r[i] = r[i], r[j]
			}
		}
	}
```

Secondly, start the inner loop from `i` and decrement and also add an early bail once the current element to be inserted is bigger than the remaining elements to the left (what ever is left of the element to be inserted _is already sorted_)
```go
	for i := 1; i < len(s); i++ {
		for j := i; j > 0 && s[j-1] > s[j]; {
			s[j-1], s[j] = s[j], s[j-1]
			j--
		}
	}
	return s
``` 
this is short and sweet, but it does it does twice as many writes as necessary and it is also not straight obvious how the `s[i]` is the current element to be inserted.
The following listing might make this clearer:

```go
	for i := 1; i < len(s); i++ {
		var j int
		for j = i; j > 0; j-- {
			if s[j-1] <= s[i] {
				break // We've found the insertion point
			}
		}
		// insert s[i] at s[j] moving the rest up
		tmp := s[i]
		copy(s[j+1:i+1], s[j:i])
		s[j] = tmp
	}
```

It's up to you if you wan to update it.

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,44 @@
...
+			if temp[j] > r[i] {
+				temp[j], r[i] = r[i], temp[j]
+			}
+		}
+	}
+	return r
+}
+
+func main() {
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
```
juliaogris: Totally a matter of taste, but IMO this should go on it's own line with:
`fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))`

02_bubble/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote("[1 2 3 5][1 2 3 5]")
+	actual := strconv.Quote(buf.String())
+	assert.Equalf(t, expected, actual, "Sorting failed.")
+}
+
+var testCases = map[string]struct {
```
juliaogris, line 25: great tests!

juliaogris (changes requested):
Looks pretty good, just a couple of comments around the actual sorting.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: I see your point and updated the logic.

02_bubble/saurabhvyas9/main_test.go
saurabhvyas9: Thanks.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Thanks Julia. Really Appreciate your feedback. 
I am using `temp` slice because it helped me visualise the slice am working on.
But I got your point about the logic and that the usage of loop was more than required. With the recent commit, I have reduced it much more. (Now, loop usage is even less than suggested logic)
Thank you helping me improve. :)

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,79 @@
...
+	r := make([]int, len(s))
+	copy(r, s)
+	for j := len(r); j > 0; j-- {
+		swapped := false
+		for i := 0; i < j-1; i++ {
+			if r[i] > r[i+1] {
+				r[i], r[i+1] = r[i+1], r[i]
+				swapped = true
+			}
+			// fmt.Println(swapped, i, j, r)
```
juliaogris: remove commented out code please.

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,79 @@
...
+		}
+		// fmt.Println(swapped, j, r)
```
juliaogris: ditto.

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,79 @@
...
+func insertion(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	// var loopCount int
+	for i := 1; i < len(r); i++ {
+		temp := r[:i]
+		var j int
+		for j = 0; j <= len(temp)-1; j++ {
+			// fmt.Println(i,j,temp,r)
+			// loopCount++
```
juliaogris: ditto

02_bubble/saurabhvyas9/main.go
```diff
@@ -0,0 +1,79 @@
...
+			break
+		}
+	}
+	return r
+}
+
+func insertion(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+	// var loopCount int
```
juliaogris: ditto.

juliaogris (changes requested):
Please [fix commit messages](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages) and remove commented code.
Comments should be reserved for explanation of unusual/unexpected behaviour and/or issue/todo links.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

02_bubble/saurabhvyas9/main.go
saurabhvyas9: Done.

juliaogris:
Still needs commit message update, see [wiki](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages).

juliaogris (approved):
Lgtm

Issue #42, opened 2019-02-12
Lab 2 - Bubble sort


eobjects:
**Lab 2 - Bubble sort**
- [ ] Create an executable go program in directory 02_bubble
- [ ] Write a function that returns a sorted copy of int slice s using Bubble sort:
```
func bubble(s []int) []int
```
- [ ] Call fmt.Println(bubble([]int{3, 2, 1, 5})) in main to print: [1 2 3 5]
- [ ] Bonus points: implement Insertion sort
- [ ] Extra bonus points: implement an O(n log(n)) sorting algorithm

eobjects:
Lab 2 - Bubble sort
Create an executable go program in directory 02_bubble
Write a function that returns a sorted copy of int slice s using Bubble sort:
func bubble(s []int) []int
Call fmt.Println(bubble([]int{3, 2, 1, 5})) in main to print:
[1 2 3 5]
Bonus points: implement Insertion sort
Extra bonus points: implement an O(n log(n)) sorting algorithm

PR #243, opened 2019-06-03
Implement Bubble sort function plus tests

runnerdave:
Fixes #[42](https://github.com/anz-bank/go-course/issues/42) 

Review of colleague's PR #[227](https://github.com/anz-bank/go-course/pull/227) 

 - Implement bubble sort algorithm
 - Cover with 100% unit tests using tables

### DIFF
--- a/02_bubble/runnerdave/main.go
+++ b/02_bubble/runnerdave/main.go
@@ -13,15 +13,15 @@ func main() {
 }
 
 func bubble(s []int) []int {
-	len := len(s)
+	sorted := make([]int, len(s))
+	copy(sorted, s)
+	len := len(sorted)
 	for i := 0; i < len-1; i++ {
 		for j := 0; j < len-i-1; j++ {
-			if s[j] > s[j+1] {
-				temp := s[j]
-				s[j] = s[j+1]
-				s[j+1] = temp
+			if sorted[j] > sorted[j+1] {
+				sorted[j], sorted[j+1] = sorted[j+1], sorted[j]
 			}
 		}
 	}
-	return s
+	return sorted
 }
--- a/02_bubble/runnerdave/main.go
+++ b/02_bubble/runnerdave/main.go
@@ -1,0 +1,27 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+	len := len(s)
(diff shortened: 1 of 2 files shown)


02_bubble/runnerdave/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestSort(t *testing.T) {
+	tables := []struct {
+		x []int
+		n []int
+	}{
+		{[]int{1, 3, 2, 5}, []int{1, 2, 3, 5}},
```
anzboi, line 30: Can you add tests for edge cases: empty and 1 element slices.
You should also test that the sort does not modify the original slice

02_bubble/runnerdave/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
```
anzboi, line 15: This function modifies `s`, it shouldn't. I realise you are passing in a copy in your tests, which is probably just as good, but we have to go with the lab instructions and require the function does the copy.

02_bubble/runnerdave/main.go
```diff
@@ -0,0 +1,27 @@
...
+	len := len(s)
+	for i := 0; i < len-1; i++ {
+		for j := 0; j < len-i-1; j++ {
+			if s[j] > s[j+1] {
+				temp := s[j]
```
anzboi: You can do this in one line using multiple assignment...
`a, b = b, a`
This doesn't lose one of the values as you might expect it to in other languages.

anzboi (changes requested):
Looking good. Aside from the input modification issue this is just about done. Would also like to test a few edge cases.

02_bubble/runnerdave/main_test.go
runnerdave: done

02_bubble/runnerdave/main.go
runnerdave: good catch 😉

02_bubble/runnerdave/main.go
runnerdave: this is very nice, thanks for the detailed review

runnerdave:
@anzboi Hi Mate, any chance to finish this off? I implemented all the changes, kind regards

anzboi (approved):
LGTM

PR #252, opened 2019-06-04
Lab2 Bubble sort and bonus Insertion sort

pfannerj:
Fixes #233

Review of colleague's PR #239

**Changes proposed in this PR:**
* build bubble sort solution
* build insertion sort solution
* build test cases for bubble and insertion sorts

PR #254, opened 2019-06-04
Lab02 - Bubble and insertion sort with unit tests

### LINKED ISSUE #253: Lab02 - Bubble and Insertion Sort
- Complete bubble sort algorithm 
- Complete insertion sort algorithm
- Complete test cases for both sort algortihm

willshen8:
Fixes #253 

Review of colleague's PR #252 

- Bubble sort
- Insertion sort
- Test cases for both sort algorithm with 100% coverage

### DIFF
--- a/02_bubble/willshen8/bubble.go
+++ b/02_bubble/willshen8/bubble.go
@@ -1,0 +1,66 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	sliceLength := len(s)
+	sorted := false
+
+	for !sorted {
+		swapOccurred := false
+		for i := 0; i < sliceLength-1; i++ {
+			if s[i+1] < s[i] {
+				swap(s, i, i+1)
+				swapOccurred = true
+			}
+		}
+		if !swapOccurred {
+			sorted = true
+		}
+	}
+	return s
+}
+
+func insertionSort(s []int) []int {
+	for i := 1; i < len(s); i++ {
+		currValue := s[i]
+		emptySlot := i
+
+		for emptySlot > 0 && s[emptySlot-1] > currValue {
+			s[emptySlot] = s[emptySlot-1]
+			emptySlot--
+		}
+		s[emptySlot] = currValue
+
+	}
+	return s
+}
+
+func swap(slice []int, pos1 int, pos2 int) []int {
+	temp := slice[pos2]
+	slice[pos2] = slice[pos1]
+	slice[pos1] = temp
+	return slice
+}
+
+func equal(a, b []int) bool {
+	if len(a) != len(b) {
+		return false
+	}
+	for i, v := range a {
+		if v != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func main() {
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
+}
--- a/02_bubble/willshen8/bubble.go
+++ b/02_bubble/willshen8/bubble.go
@@ -9,14 +9,15 @@ import (
 var out io.Writer = os.Stdout
 
 func bubble(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
 	sliceLength := len(s)
 	sorted := false
 
 	for !sorted {
 		swapOccurred := false
 		for i := 0; i < sliceLength-1; i++ {
-			if s[i+1] < s[i] {
-				swap(s, i, i+1)
+			if sliceCopy[i+1] < sliceCopy[i] {
+				sliceCopy[i], sliceCopy[i+1] = sliceCopy[i+1], sliceCopy[i]
 				swapOccurred = true
 			}
 		}
@@ -24,29 +25,21 @@ func bubble(s []int) []int {
 			sorted = true
 		}
 	}
-	return s
+	return sliceCopy
 }
 
 func insertionSort(s []int) []int {
-	for i := 1; i < len(s); i++ {
-		currValue := s[i]
-		emptySlot := i
+	sliceCopy := append(s[:0:0], s...)
+	for key, value := range s {
 
-		for emptySlot > 0 && s[emptySlot-1] > currValue {
-			s[emptySlot] = s[emptySlot-1]
-			emptySlot--
+		for key > 0 && sliceCopy[key-1] > value {
+			sliceCopy[key] = sliceCopy[key-1]
+			key--
 		}
-		s[emptySlot] = currValue
+		sliceCopy[key] = value
 
 	}
-	return s
-}
-
-func swap(slice []int, pos1 int, pos2 int) []int {
-	temp := slice[pos2]
-	slice[pos2] = slice[pos1]
-	slice[pos1] = temp
-	return slice
+	return sliceCopy
 }
 
 func equal(a, b []int) bool {
--- a/02_bubble/willshen8/bubble_test.go
+++ b/02_bubble/willshen8/bubble_test.go
@@ -19,7 +19,7 @@ func TestMainOutput(t *testing.T) {
 	}
 }
 
-func TestBubbleSort(t *testing.T) {
+func TestBubbleAndInsertionSort(t *testing.T) {
 	tests := []struct {
 		name     string
 		input    []int
@@ -30,6 +30,8 @@ func TestBubbleSort(t *testing.T) {
 			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
 		{name: "Already sorted slice", input: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
 			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{name: "Slice with one number different to rest", input: []int{1, 1, 1, 2, 1},
+			expected: []int{1, 1, 1, 1, 2}},
 		{name: "Slice with negative number", input: []int{1, -1}, expected: []int{-1, 1}},
 	}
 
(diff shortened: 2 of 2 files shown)


02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,93 @@
...
+		{name: "Already sorted slice", input: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{name: "Slice with one number different to rest", input: []int{1, 1, 1, 2, 1},
+			expected: []int{1, 1, 1, 1, 2}},
+		{name: "Slice with negative number", input: []int{1, -1}, expected: []int{-1, 1}},
+	}
+
+	for _, test := range tests {
+		actual := insertionSort(test.input)
+		testResult := Equal(actual, test.expected)
```
golangcibot: undeclared name: `Equal` (from `typecheck`)

02_bubble/willshen8/bubble.go
```diff
@@ -0,0 +1,68 @@
...
+			sliceCopy[emptySlot] = sliceCopy[emptySlot-1]
+			emptySlot--
+		}
+		sliceCopy[emptySlot] = currValue
+
+	}
+	return sliceCopy
+}
+
+func swap(slice []int, pos1 int, pos2 int) []int {
```
marksomething: Go allows you to do in-place swaps `slice[pos1], slice[pos2] = slice[pos2], slice[pos1]`
You can replace this function with that.

02_bubble/willshen8/bubble.go
```diff
@@ -0,0 +1,68 @@
...
+	return sliceCopy
+}
+
+func insertionSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
+	for i := 1; i < len(s); i++ {
+		currValue := sliceCopy[i]
+		emptySlot := i
+
+		for emptySlot > 0 && sliceCopy[emptySlot-1] > currValue {
```
marksomething: an alternative here is to use copy() to move the existing entries rather than shuffling them along one at a time.

02_bubble/willshen8/bubble.go
```diff
@@ -0,0 +1,68 @@
...
+	for !sorted {
+		swapOccurred := false
+		for i := 0; i < sliceLength-1; i++ {
+			if sliceCopy[i+1] < sliceCopy[i] {
+				swap(sliceCopy, i, i+1)
+				swapOccurred = true
+			}
+		}
+		if !swapOccurred {
+			sorted = true
```
marksomething, line 25: Had a review comment on my code suggest moving the `return` here. It's kinda neat and removes the need for the `sorted` variable.

02_bubble/willshen8/bubble.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	return sliceCopy
+}
+
+func insertionSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
+	for i := 1; i < len(s); i++ {
```
marksomething: Looping an entire slice is more idiomatically done using `range`. You have used that technique elsewhere.

02_bubble/willshen8/bubble.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func swap(slice []int, pos1 int, pos2 int) []int {
+	temp := slice[pos2]
+	slice[pos2] = slice[pos1]
+	slice[pos1] = temp
+	return slice
+}
+
+func equal(a, b []int) bool {
```
marksomething, line 45: Have a look at testify/assert, it was recommended to me by @juliaogris and spares you from needing to implement your own slice equality (and diff printing) code.

02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,93 @@
...
+	for _, test := range tests {
+		t.Logf("Running test: %v", test.name)
+		actual := equal(test.sliceA, test.sliceB)
+		if actual != test.expected {
+			t.Fatalf("Test failed! Expected: %t, Got: %t", test.expected, actual)
+		}
+	}
+}
+
+func TestInsertionSort(t *testing.T) {
```
marksomething: Consider using the same cases between your two sorting implementations. These are all really good choices for test cases 👍

02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,93 @@
...
+	tests := []struct {
```
marksomething: Consider using a `map[string]struct{[]int, []int}`

02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,93 @@
...
+	}{
+		{name: "Empty slice", input: []int{}, expected: []int{}},
+		{name: "Reverse sorted slice", input: []int{10, 9, 8, 7, 6, 5, 4, 3, 2, 1},
+			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{name: "Already sorted slice", input: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10},
+			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{name: "Slice with negative number", input: []int{1, -1}, expected: []int{-1, 1}},
+	}
+
+	for _, test := range tests {
```
marksomething, line 38: Have a look at "Introducing sub tests" in [this blog](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests)

marksomething (changes requested):
Nice work! You code is really good in general, most of the ideas here I only learnt from other peoples code reviews :)

02_bubble/willshen8/bubble.go
willshen8: Thanks, done.

02_bubble/willshen8/bubble.go
willshen8: Thanks, just very used to this kind of loop and forgot to use the new style.

02_bubble/willshen8/bubble_test.go
willshen8: Sorry, not sure what you meant here.

02_bubble/willshen8/bubble.go
willshen8: Great point, that's for pointing that out for me. Will use testify next time.

02_bubble/willshen8/bubble_test.go
willshen8: Map actually is slower as it iterate through both the key and value. So in this case the original struct is better in performance.

02_bubble/willshen8/bubble.go
anzboi: One more note, you are modifying the slice here, no real need to return it. The swap can be done in one line
`slice[pos1], slice[pos2] = slice[pos2], slice[pos1]`
so it is debateable whether you need a separate function to do so, though it does make things slightly more readable in the sort functions.

02_bubble/willshen8/bubble_test.go
marksomething: You used different tests scenarios for bubble and insertionSort, it's not clear why they would not be the same.

02_bubble/willshen8/bubble_test.go
marksomething: It's a bit of a style choice I guess. The performance impact is trivial for this usage. I'm happy to accept this.

02_bubble/willshen8/bubble.go
marksomething: Don't forget to push your changes.

02_bubble/willshen8/bubble_test.go
willshen8: ok, thanks, I thought you meant upper or lower cases.

02_bubble/willshen8/bubble.go
willshen8: I couldn't get it to work without making my code confusing. I will definitely take on this advise in the future.

02_bubble/willshen8/bubble.go
willshen8: Thanks Dan, I copied the slice into a copySlice and only returning the copied one. So not mutating it.

02_bubble/willshen8/bubble_test.go
willshen8: Done, I included all test cases for both sort and merged the two test functions into one.

02_bubble/willshen8/bubble_test.go
anzboi: I find subtests are sometimes warranted and other times not.
Good: Different test cases (edge cases, different input types etc)
Bad: Randomized tests
In this case it would be a good idea to use them, but I will leave this in the 'do next time' bucket. since you do go through the effort to report the name in the test logs.
To run a subtest
```
  // sub test, written just like any other test, have access to test data
})
```

02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,78 @@
...
+		{name: "Slice with one number different to rest", input: []int{1, 1, 1, 2, 1},
+			expected: []int{1, 1, 1, 1, 2}},
+		{name: "Slice with negative number", input: []int{1, -1}, expected: []int{-1, 1}},
+	}
+
+	for _, test := range tests {
+		actual := bubble(test.input)
+		testResult := equal(actual, test.expected)
+		if testResult == false {
+			t.Fatalf("Bubble Sort Test Failed! Expected: %v, Got: %v", test.expected, actual)
```
anzboi, line 42: I recommend use `Errorf` instead of `Fatalf`. Using `Fatalf` only causes you to see the first case that failed, not all the failing cases.

02_bubble/willshen8/bubble_test.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+}
+
+func TestBubbleAndInsertionSort(t *testing.T) {
+	tests := []struct {
+		name     string
+		input    []int
+		expected []int
+	}{
+		{name: "Empty slice", input: []int{}, expected: []int{}},
```
anzboi, line 28: Great test coverage and incorporating both sorts in the same test. We will do suite testing when we get into the rest api and you will see how you can run the same test on multiple implementations of a particular function/interface.

anzboi (dismissed):
LGTM.
I've commented on how you can run sub-tests, but don't feel the need to change for this PR, keep in mind for next time.

marksomething (dismissed):
👍

anzboi (approved):
LGTM

PR #259, opened 2019-06-05
Lab1 - Fibonacci nega & positive function incl testing

Sarahdavy1983:
Fixes ##258

Review of colleague's PR #148 

Create an executable go program in directory 01_fib/USERNAME
Write a function that prints the first n Fibonacci numbers

### DIFF
--- a/01_fib/sarahdavy1983/main.go
+++ b/01_fib/sarahdavy1983/main.go
@@ -11,6 +11,7 @@ var out io.Writer = os.Stdout
 
 // fib returns a slice of integer fibonacci numbers
 func fib(n int) []int64 {
+
 	switch {
 	case n == 0:
 		return []int64{0, 0}
@@ -22,13 +23,12 @@ func fib(n int) []int64 {
 		x := fib(n + 1)
 		nextVal := x[len(x)-2] - x[len(x)-1]
 		return append(x, nextVal)
-
-	default:
+	case n > 1:
 		x := fib(n - 1)
 		nextVal := x[len(x)-2] + x[len(x)-1]
 		return append(x, nextVal)
 	}
-
+	return nil
 }
 
 //printEachindex iterates over the slice and prints each index
--- a/01_fib/sarahdavy1983/main.go
+++ b/01_fib/sarahdavy1983/main.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// Write to this rather that Stdout directly to help with testing
+var out io.Writer = os.Stdout
+
+// fib returns a slice of integer fibonacci numbers
+func fib(n int) []int64 {
+	switch {
+	case n == 0:
+		return []int64{0, 0}
+	case n == 1:
+		return []int64{0, 1}
+	case n == -1:
+		return []int64{0, 1}
+	case n < -1:
+		x := fib(n + 1)
+		nextVal := x[len(x)-2] - x[len(x)-1]
+		return append(x, nextVal)
+
+	default:
+		x := fib(n - 1)
+		nextVal := x[len(x)-2] + x[len(x)-1]
+		return append(x, nextVal)
+	}
+
+}
+
+//printEachindex iterates over the slice and prints each index
+func printEachIndex(fibSlice []int64) {
+	for i := 1; i < len(fibSlice); i++ {
+		fmt.Fprintln(out, fibSlice[i])
+	}
+}
+func main() {
+	printEachIndex(fib(7))
+}
--- a/01_fib/sarahdavy1983/main.go
+++ b/01_fib/sarahdavy1983/main.go
@@ -23,12 +23,12 @@ func fib(n int) []int64 {
 		x := fib(n + 1)
 		nextVal := x[len(x)-2] - x[len(x)-1]
 		return append(x, nextVal)
-	case n > 1:
+	default:
 		x := fib(n - 1)
 		nextVal := x[len(x)-2] + x[len(x)-1]
 		return append(x, nextVal)
 	}
-	return nil
+
 }
 
 //printEachindex iterates over the slice and prints each index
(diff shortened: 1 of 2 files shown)


01_fib/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,55 @@
...
+		{input: 0, want: []int64{0, 0}},
+		{input: 1, want: []int64{0, 1}},
+		{input: 7, want: []int64{0, 1, 1, 2, 3, 5, 8, 13}},
+		{input: -1, want: []int64{0, 1}},
+	}
+
+	for _, tc := range tests {
+
+		got := (fib(tc.input))
+		//strconv.Quote(buf.String())
```
cuminandpaprika: We should probably remove commented out code even in test files, as suggested in the wiki https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments

01_fib/sarahdavy1983/main.go
```diff
@@ -0,0 +1,42 @@
...
+	case n == 1:
+		return []int64{0, 1}
+	case n == -1:
+		return []int64{0, 1}
+	case n < -1:
+		x := fib(n + 1)
+		nextVal := x[len(x)-2] - x[len(x)-1]
+		return append(x, nextVal)
+
+	default:
```
cuminandpaprika, line 26: Sorry, just a nitpick here. Thoughts on using `case n > 1:` here instead of `default:` to make the last case handled clearer to the reader?

cuminandpaprika (changes requested):
Hey there, awesome work overall, just two tiny nitpicks which I've commented on above

01_fib/sarahdavy1983/main.go
Sarahdavy1983: I understand your concern but after removing the default and replacing with a case the program did not pass the build, so I have reverted back to default case within the switch

cuminandpaprika: Thanks for trying that out, I tested that locally and it seems the Go compiler doesn't believe we've handled every case when we use `case n > 1:` and says that not all code paths return a value. 
image


Could theoretically fix it by adding a `return 0` outside of the switch statement but that's wrong on too many levels! Great work :)

cuminandpaprika:
Thanks for the fixes, good to merge

cuminandpaprika (commented):
Good to merge

cuminandpaprika (approved):
Okay finally found the approve button, sorry!

01_fib/sarahdavy1983/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"io"
+	"os"
+)
+
+// Write to this rather that Stdout directly to help with testing
+var out io.Writer = os.Stdout
+
+// fib returns a slice of integer fibonacci numbers
+func fib(n int) []int64 {
+
```
juliaogris, line 14: unnecessary blank line

01_fib/sarahdavy1983/main.go
```diff
@@ -0,0 +1,42 @@
...
+	switch {
+	case n == 0:
+		return []int64{0, 0}
+	case n == 1:
+		return []int64{0, 1}
+	case n == -1:
+		return []int64{0, 1}
```
juliaogris, line 21: pull together as `case n==1 || n == -1`

01_fib/sarahdavy1983/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+// Write to this rather that Stdout directly to help with testing
+var out io.Writer = os.Stdout
+
+// fib returns a slice of integer fibonacci numbers
+func fib(n int) []int64 {
+
+	switch {
+	case n == 0:
+		return []int64{0, 0}
```
juliaogris, line 17: Bit of hair splitting here but imo this should be `[]int64{0}`

01_fib/sarahdavy1983/main.go
```diff
@@ -0,0 +1,42 @@
...
+	default:
+		x := fib(n - 1)
+		nextVal := x[len(x)-2] + x[len(x)-1]
+		return append(x, nextVal)
+	}
+
+}
+
+//printEachindex iterates over the slice and prints each index
+func printEachIndex(fibSlice []int64) {
```
juliaogris, line 35: I'd name this `printIntSlice` or similar (nothing to do with fib imo) and have it print from index 0.
Pass it the appropriate slice with fib(n)[1:].
You could also rename your `fib` function to something like `fibSlice` because the required `fib` function was meant to have signature `fib(int n)`.
Then have a wrapper function `fib(int n)` that calls `printIntSlice(fibSlice(n)[1:])` or so

It's a bit a question of taste, so feel free to push back ;)

01_fib/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+	tests := []test{
+		{input: -7, want: []int64{0, 1, -1, 2, -3, 5, -8, 13}},
+		{input: 0, want: []int64{0, 0}},
+		{input: 1, want: []int64{0, 1}},
+		{input: 7, want: []int64{0, 1, 1, 2, 3, 5, 8, 13}},
+		{input: -1, want: []int64{0, 1}},
+	}
+
+	for _, tc := range tests {
+
```
juliaogris, line 48: Unnecessary blank line.

01_fib/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,54 @@
...
+		got := (fib(tc.input))
+		if !reflect.DeepEqual(tc.want, got) {
```
juliaogris, line 50: Consider using `testify/assert`: it will turn these three lines into a one-liner and print useful diff information in case of failure.

juliaogris (changes requested):

juliaogris:

PR #271, opened 2019-06-05
Implement letter frequency counting and printing

### LINKED ISSUE #270: lab03 for @lexual

lexual:
Fixes #270

Review of colleague's PR #235

### DIFF
--- a/03_letters/lexual/freq.go
+++ b/03_letters/lexual/freq.go
@@ -28,7 +28,7 @@ func sortLetters(m map[rune]int) []string {
 	sort.Strings(keys)
 	sorted := make([]string, len(m))
 	for i, k := range keys {
-		sorted[i] = fmt.Sprintf("%s:%d", string(k), m[rune(k[0])])
+		sorted[i] = fmt.Sprintf("%s:%d", k, m[rune(k[0])])
 	}
 	return sorted
 }
--- a/03_letters/lexual/freq.go
+++ b/03_letters/lexual/freq.go
@@ -19,19 +19,16 @@ func letters(s string) map[rune]int {
 }
 
 func sortLetters(m map[rune]int) []string {
-	keys := make([]rune, len(m))
+	keys := make([]string, len(m))
 	i := 0
 	for k := range m {
-		keys[i] = k
+		keys[i] = string(k)
 		i++
 	}
-	// sort keys
-	sort.Slice(keys, func(i, j int) bool {
-		return keys[i] < keys[j]
-	})
+	sort.Strings(keys)
 	sorted := make([]string, len(m))
 	for i, k := range keys {
-		sorted[i] = fmt.Sprintf("%s:%d", string(k), m[k])
+		sorted[i] = fmt.Sprintf("%s:%d", string(k), m[rune(k[0])])
 	}
 	return sorted
 }
--- a/03_letters/lexual/freq.go
+++ b/03_letters/lexual/freq.go
@@ -19,17 +19,19 @@ func letters(s string) map[rune]int {
 }
 
 func sortLetters(m map[rune]int) []string {
-	var keys []rune
+	keys := make([]rune, len(m))
+	i := 0
 	for k := range m {
-		keys = append(keys, k)
+		keys[i] = k
+		i++
 	}
 	// sort keys
 	sort.Slice(keys, func(i, j int) bool {
 		return keys[i] < keys[j]
 	})
-	var sorted []string
-	for _, k := range keys {
-		sorted = append(sorted, fmt.Sprintf("%s:%d", string(k), m[k]))
+	sorted := make([]string, len(m))
+	for i, k := range keys {
+		sorted[i] = fmt.Sprintf("%s:%d", string(k), m[k])
 	}
 	return sorted
 }
--- a/03_letters/lexual/freq.go
+++ b/03_letters/lexual/freq.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"unicode/utf8"
 )
 
 var out io.Writer = os.Stdout
@@ -28,7 +29,8 @@ func sortLetters(m map[rune]int) []string {
 	sort.Strings(keys)
 	sorted := make([]string, len(m))
 	for i, k := range keys {
-		sorted[i] = fmt.Sprintf("%s:%d", k, m[rune(k[0])])
+		r, _ := utf8.DecodeRuneInString(k)
+		sorted[i] = fmt.Sprintf("%s:%d", k, m[r])
 	}
 	return sorted
 }
--- a/03_letters/lexual/freq.go
+++ b/03_letters/lexual/freq.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	m := make(map[rune]int)
+	for _, r := range s {
+		m[r]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	var keys []rune
+	for k := range m {
+		keys = append(keys, k)
+	}
+	// sort keys
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i] < keys[j]
+	})
+	var sorted []string
+	for _, k := range keys {
+		sorted = append(sorted, fmt.Sprintf("%s:%d", string(k), m[k]))
+	}
+	return sorted
+}
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
--- a/03_letters/lexual/freq_test.go
+++ b/03_letters/lexual/freq_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"reflect"
 	"strconv"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


03_letters/lexual/freq.go
```diff
@@ -0,0 +1,39 @@
...
+func letters(s string) map[rune]int {
+	m := make(map[rune]int)
+	for _, r := range s {
+		m[r]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	var keys []rune
```
golangcibot: Consider preallocating `keys` (from `prealloc`)

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,39 @@
...
+	for k := range m {
+		keys = append(keys, k)
+	}
+	// sort keys
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i] < keys[j]
+	})
+	var sorted []string
```
golangcibot: Consider preallocating `sorted` (from `prealloc`)

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,41 @@
...
+	for _, r := range s {
+		m[r]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]rune, len(m))
+	i := 0
+	for k := range m {
```
n0npax, line 25: `for i, k : range ... ` may be a better fit here

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,41 @@
...
+		keys[i] = k
+		i++
+	}
+	// sort keys
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i] < keys[j]
+	})
+	sorted := make([]string, len(m))
+	for i, k := range keys {
```
n0npax, line 31: I believe sorting should be done by `value`(occurrences) not by the `key`.

n0npax (changes requested):

03_letters/lexual/freq.go
lexual: isn't the behaviour that when you use range over a map the resulting pair is key,value not index,value ?

03_letters/lexual/freq.go
lexual: The readme example seems to be sorted by key:

https://github.com/anz-bank/go-course/blob/master/03_letters/README.md

03_letters/lexual/freq.go
n0npax: aww right, It's a map. sorry for that

03_letters/lexual/freq.go
n0npax: I just checked merged examples and they are done in the both ways, so it's fine.

n0npax (dismissed):

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,41 @@
...
+	i := 0
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	// sort keys
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i] < keys[j]
+	})
+	sorted := make([]string, len(m))
```
rokane, line 30: By converting the map into a `[]rune` as the first step, we are introducing an intermediate step. Is there a need to convert to the `[]rune` first??? Perhaps we can create a `[]string` directly from the map, which will simplify this quite a lot. If that were the case, you could even perhaps use the `sort.Strings(input)` method to sort the slice, rather than providing your comparison function to the `sort.Slice()` method. 

Have a go at eliminating this `[]rune` and see if you can go directly to a `[]string`

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]rune, len(m))
+	i := 0
+	for k := range m {
+		keys[i] = k
+		i++
+	}
+	// sort keys
```
rokane: comment not required, please remove

03_letters/lexual/freq_test.go
```diff
@@ -0,0 +1,94 @@
...
+	}{
+		{
+			map[rune]int{
+				'a': 2,
+				'b': 1,
+				'c': 3,
+			},
+			[]string{"a:2", "b:1", "c:3"},
+		},
+		// empty map as argument tested
```
rokane: comment not required.

03_letters/lexual/freq_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, test := range tests {
+		actual := letters(test.input)
+		if len(actual) != len(test.expected) {
+			t.Errorf("len of actual and expected differ")
+		}
+		for k, v := range test.expected {
+			if v != actual[k] {
+				t.Errorf("Unexpected result")
+				t.Errorf("\nActual: %v\nExpected: %v", actual, test.expected)
+				for k, v := range actual {
```
rokane: I really don't think the looping over the `actual` and `expected` is beneficial here. It looks like you are doing this just for the purpose of printing, but you have already printed both the actual and expected above. This just introduces two iterations which are not required. Please remove this.

I would recommend running these test and providing an input which will make them fail. That way you will see what is printed and you will be able to determine if this is overkill. You will see only one `t.Errorf` should be required

03_letters/lexual/freq_test.go
```diff
@@ -0,0 +1,94 @@
...
+	for _, test := range tests {
+		actual := sortLetters(test.input)
+		if len(test.expected) != len(actual) {
+			t.Errorf("expected and actual different lengths")
+		}
+		for i := range test.expected {
+			if actual[i] != test.expected[i] {
+				t.Errorf("actual and expected don't match")
+				t.Errorf("\nActual: %s\nExpected: %s", actual, test.expected)
+
```
rokane: Little nit here. But there is an extra line. Please remove this.

Also is there a reason why you are breaking here?

rokane (dismissed):
Looking good so far. I have left a hint for how to achieve the `sortLetters` function without the need for an intermediate slice. See if you can get this working, it will simplify and reduce your code a lot.

Can you also please update the description of the pull request so you are outlining the changes proposed in this piece of work. Thanks

03_letters/lexual/freq_test.go
lexual: its because printing the map, shows the keys as integers instead of letters.

i.e. 

```Actual: map[97:1 98:2 99:3]
Expected: map[97:1 98:22 99:3]```

Isn't very useful for debugging, because I need to determine in my head that 97 corresponds to 'a'.

Full output gives this:

```freq_test.go:35: 
    Actual: map[97:1 98:2 99:3]
    Expected: map[97:1 98:22 99:3]
freq_test.go:37: Actual a = 1
freq_test.go:37: Actual b = 2
freq_test.go:37: Actual c = 3
freq_test.go:40: Expect c = 3
freq_test.go:40: Expect a = 1
freq_test.go:40: Expect b = 22
```

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,38 @@
...
+	keys := make([]string, len(m))
+	i := 0
+	for k := range m {
+		keys[i] = string(k)
+		i++
+	}
+	sort.Strings(keys)
+	sorted := make([]string, len(m))
+	for i, k := range keys {
+		sorted[i] = fmt.Sprintf("%s:%d", string(k), m[rune(k[0])])
```
golangcibot: unnecessary conversion (from `unconvert`)

03_letters/lexual/freq.go
lexual: gone in new commit

03_letters/lexual/freq.go
lexual: gone in new commit

03_letters/lexual/freq_test.go
lexual: gone in new commit

lexual:
@rokane Can you follow up on PR. I believe I had responded to your past comments ;)

03_letters/lexual/freq.go
camscale: The readme example is sorted both by key (ascending) and by count (descending). Which should you do? Either way, you should document your assumptions.

03_letters/lexual/freq.go
```diff
@@ -0,0 +1,38 @@
...
+	keys := make([]string, len(m))
+	i := 0
+	for k := range m {
+		keys[i] = string(k)
+		i++
+	}
+	sort.Strings(keys)
+	sorted := make([]string, len(m))
+	for i, k := range keys {
+		sorted[i] = fmt.Sprintf("%s:%d", k, m[rune(k[0])])
```
camscale: A question for you to answer yourself - why do you need the `rune` cast here? what does that mean for the value returned by the `k[0]` expression? It is related to the test case addition I suggested. Hope I'm not being too obtuse - just trying to get you to think about this some more.

03_letters/lexual/freq_test.go
```diff
@@ -0,0 +1,91 @@
...
+		expected []string
+	}{
+		{
+			map[rune]int{
+				'a': 2,
+				'b': 1,
+				'c': 3,
+			},
+			[]string{"a:2", "b:1", "c:3"},
+		},
```
camscale, line 66: You should add a test case with unicode characters that are encoded as multi-byte values as utf8. You might be surprised by the result (and then you can fix your code - you may want to look at `utf8.DecodeRuneInString()`)

03_letters/lexual/freq_test.go
```diff
@@ -0,0 +1,91 @@
...
+			},
+		},
+		{
+			"",
+			map[rune]int{},
+		},
+	}
+	for _, test := range tests {
+		actual := letters(test.input)
+		if len(actual) != len(test.expected) {
```
camscale: I feel this whole test is too verbose. It should be simple to understand what is being tested. You could just use `reflect.DeepEqual`. I understand your comment about the output being integers instead of characters, but you need to balance that with the readability of the test. If you really still want prettier output, I suggest adding a function that creates a string representation of the map as you want and using that in your error message. That keeps the test simple and moves the map formatting logic elsewhere.

camscale (changes requested):
Your implementation is subtly wrong and has gone unnoticed because your test suite is too short. Have a read of https://blog.golang.org/strings for the gory details of strings which may help you know where you've gone wrong.

camscale:
Also, I noticed you asked for a re-review but you still had the "Changes Requested" tag and did not have the "Ready for Review" tag. We won't pick things up for review tagged like this.

03_letters/lexual/freq.go
lexual: adopted DecodeRuneInString ;)

03_letters/lexual/freq_test.go
lexual: simplified

camscale (changes requested):
The changes you said you made are not here. Did you forget to push?

lexual:
> The changes you said you made are not here. Did you forget to push?

D'oh! 

Yes, rebased and pushed just now ;(

camscale (approved):
LGTM.

When you merge this, you can either do a squash merge (give your commit a decent comment at that time) or you can rebase-squash-force-push down to a single commit and do a merge commit. I personally prefer to do the former (I like merge commits and I often have multiple commits in a PR, but I still need to squash fixups into the relevant commit). This is to avoid committing fixup commits to master to keep the history clean. If you do the latter, ping me on slack and I'll try to get to re-approving quickly (which is needed because a rebase-force-push needs re-approval) so you can merge.

juliaogris (dismissed):
Please [fix commit messages](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages) with `git rebase -i HASH` and add a PR description.

anzdaddy:
Please merge or close.

PR #216, opened 2019-06-02
Implement numeronyms

nicholascross:
Fixes: #231

Review of colleague's PR #211

#### Changes proposed in this PR:
- Implement function that returns a slice of numeronyms for its input strings
- Create an executable to demonstrate
- Add unit tests

PR #274, opened 2019-06-06
Implement numeronym & numeronyms functions

### LINKED ISSUE #273: lab04 for @lexual

lexual:
Fixes #273

Review of colleague's PR #216


04_numeronym/lexual/numeronym_test.go
```diff
@@ -0,0 +1,78 @@
...
+		{"Kubernetes", "K8s"},
+		{"abc", "abc"},
+		{"abcd", "a2d"},
+		{"ab", "ab"},
+		{"a", "a"},
+		{"", ""},
+	}
+	for _, test := range tests {
+		t.Run(test.input, func(t *testing.T) {
+			actual := numeronym(test.input)
```
golangcibot, line 25: Using the variable on range scope `test` in function literal (from `scopelint`)

04_numeronym/lexual/numeronym_test.go
```diff
@@ -0,0 +1,78 @@
...
+			if actual != test.expected {
```
golangcibot, line 26: Using the variable on range scope `test` in function literal (from `scopelint`)

04_numeronym/lexual/numeronym_test.go
```diff
@@ -0,0 +1,78 @@
...
+				t.Errorf("input: %s, expected: %s, actual: %s",
+					test.input, test.expected, actual)
```
golangcibot, line 28: Using the variable on range scope `test` in function literal (from `scopelint`)

lexual:
looks good

04_numeronym/lexual/numeronym_test.go
n0npax: you shouldn't mark PR as ready for review as long as there are lint issues

lexual: OK, I'm a little confused. I thought the CI had run against an older version of the code. I believe  the 1st line of the for-loop is ```test := test``` now

04_numeronym/lexual/numeronym.go
```diff
@@ -0,0 +1,29 @@
...
+	"strconv"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronym(s string) string {
+	if len(s) <= 3 {
+		return s
+	}
+	return s[:1] + strconv.Itoa(len(s)-2) + s[len(s)-1:]
```
anzboi, line 16: You don't need to access an entire slice for a single element. `s[0] + ... + s[len(s)-1]` would do.
Since we are dealing with strings, the result of the `+` ends up the same anyway though, so all good.

04_numeronym/lexual/numeronym.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+func numeronyms(vals ...string) []string {
```
anzboi, line 19: Perfect 👍

04_numeronym/lexual/numeronym_test.go
anzboi: Your second test (`TestNumeronyms`) does contain `test := test`, but this one doesn't.

anzboi (approved):
LGTM, apologies for the delay.
I've left a comment about accessing single elements. It's ok since here we are luckily dealing with strings so don't feel the need to change it. I would also ignore the `test := test`, I find it super annoying and dumb that line has to even exist.

PR #277, opened 2019-06-06
Implement Stringer - Lab 5

### LINKED ISSUE #276: Lab 5 - Stringer
Create an executable go program in directory 05_stringer/brvishnu92
Make the IPAddr type implement fmt.Stringer to print the address as a dotted quad.

brvishnu92:
Fixes #276

Review of colleague's PR #337

-Create an executable go program in directory 05_stringer/brvishnu92
-Make the IPAddr type implement fmt.Stringer to print the address as a dotted quad.


05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+//main() test
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := `127.0.0.1
+`
+	actual := buf.String()
+	if expected != actual {
```
marksomething: Have a look at [testify/assert](https://godoc.org/github.com/stretchr/testify/assert), you've used it in your second test!

05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+		t.Errorf("Unexpected output in main()\nexpected: %q\nactual: %q", expected, actual)
+	}
+}
+
+func TestNumeronyms(t *testing.T) {
```
marksomething: Rename

05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+		expected string
+	}{
+		"test1":  {input: IPAddr{127, 0, 0, 1}, expected: "127.0.0.1"},
+		"empty":  {input: IPAddr{}, expected: "0.0.0.0"},
+		"oneval": {input: IPAddr{127}, expected: "127.0.0.0"},
+	}
+	for name, test := range testCases {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			result := test.input.String()
```
marksomething: The thing that is special about the Stringer interface is that it is called by functions in the `fmt` library. Your test should probably use a function from `fmt`

05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+			assert.Equalf(t, result, test.expected, "got %v want %v", result, test.expected)
```
marksomething: If you are not going to do a generic error message you can use `assert.Equal(t, expected,  actual)`.
The signature for [`assert.Equalf`](https://godoc.org/github.com/stretchr/testify/assert#Equalf) puts the expected before actual

marksomething (changes requested):
Your code looks good, but there's a few things with your tests, please also link to a peer review in your description.

05_stringer/brvishnu92/main_test.go
brvishnu92: Done, swapped them around but would like to keep the equalf since i dont like the formatting of the default message.

05_stringer/brvishnu92/main_test.go
brvishnu92: Sure, changed it.. :-)

05_stringer/brvishnu92/main_test.go
brvishnu92: Done.

marksomething (dismissed):
👍

05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"bytes"
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+//main() test
+func TestMainOutput(t *testing.T) {
```
rokane, line 11: Again here I don't see the benefit in this comment. The name of the function is documenting the same thing.

05_stringer/brvishnu92/main_test.go
```diff
@@ -0,0 +1,38 @@
...
+	actual := buf.String()
+	assert.Equalf(t, expected, actual, "Got: %v Want: %v", actual, expected)
+}
+
+func TestStringer(t *testing.T) {
+	testCases := map[string]struct {
+		input    IPAddr
+		expected string
+	}{
+		"test1":  {input: IPAddr{127, 0, 0, 1}, expected: "127.0.0.1"},
```
rokane: I get the feeling that you have copied the same test skeleton code from your previous labs 3 and 4 into this as they all have the same three test cases (`test1, empty, oneval`).

I can see that it is beneficial to copy the skeleton code of a generic test involving multiple cases like these have been across the labs. However when constructing specific cases for the test to run you should be thinking about the different edge cases and possible outputs that might occur from inputs to this function. Rather than just assigning a generic three cases to each function, you should analyse the function and see what can happen. 

For this, what happens if you give it a value of empty, single value, two values, three values etc. And use descriptive names for test cases, `test1` is non descriptive and you won't get much out of seeing a failure message with that test name.

Can you please add more cases, and update the name of the tests accordingly.

rokane (changes requested):
Code looks really good, but some advice with testing has been given on the construction of test cases and naming of them. If you can update according to those suggestions then I will take another look. 👍

05_stringer/brvishnu92/main_test.go
brvishnu92: Done.

05_stringer/brvishnu92/main_test.go
brvishnu92: Done. Yeah from now on will look into the boundary cases. Thanks for that. :-)

rokane (approved):
Looks good to me. Nice work!

PR #261, opened 2019-06-05
Implement bubble and insertion sort

lexual:
Fixes #260

Review of colleague's PR #204

PR #279, opened 2019-06-06
Lab2 - BubbleSort and InsertionSort plus testing

### LINKED ISSUE #278: Lab 2 Bubble and Insertion Sort
Create an executable go program in directory 02_bubble/USERNAME
Write a function that returns a sorted copy of int slice s using Bubble sort
Bonus points: implement Insertion sort

Sarahdavy1983:
Fixes #278

Review of colleague's PR #261 

Create an executable go program in directory 02_bubble/USERNAME
Write a function that returns a sorted copy of int slice s using Bubble sort
Bonus points: implement Insertion sort

### DIFF
--- a/02_bubble/sarahdavy1983/main.go
+++ b/02_bubble/sarahdavy1983/main.go
@@ -18,7 +18,21 @@ func bubbleSort(s []int) []int {
 	}
 	return s
 }
+func insertionSort(s []int) []int {
+	n := len(s)
+	for i := 1; i < n; i++ {
+		j := i
+		for j > 0 {
+			if s[j-1] > s[j] {
+				s[j-1], s[j] = s[j], s[j-1]
+			}
+			j--
+		}
+	}
+	return s
+}
 
 func main() {
 	fmt.Println(bubbleSort([]int{3, 2, 1, 5}))
+	fmt.Println(insertionSort([]int{3, 2, 1, 5, -1}))
 }
--- a/02_bubble/sarahdavy1983/main.go
+++ b/02_bubble/sarahdavy1983/main.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"fmt"
+)
+
+func bubbleSort(s []int) []int {
+	n := len(s)
+	swapped := true
+	for swapped {
+		swapped = false
+		for i := 1; i < n; i++ {
+			if s[i-1] > s[i] {
+				s[i], s[i-1] = s[i-1], s[i]
+				swapped = true
+			}
+		}
+	}
+	return s
+}
+
+func main() {
+	fmt.Println(bubbleSort([]int{3, 2, 1, 5}))
+}
--- a/02_bubble/sarahdavy1983/main.go
+++ b/02_bubble/sarahdavy1983/main.go
@@ -5,34 +5,36 @@ import (
 )
 
 func bubbleSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
 	n := len(s)
 	swapped := true
 	for swapped {
 		swapped = false
 		for i := 1; i < n; i++ {
-			if s[i-1] > s[i] {
-				s[i], s[i-1] = s[i-1], s[i]
+			if sliceCopy[i-1] > sliceCopy[i] {
+				sliceCopy[i], sliceCopy[i-1] = sliceCopy[i-1], sliceCopy[i]
 				swapped = true
 			}
 		}
 	}
-	return s
+	return sliceCopy
 }
 func insertionSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
 	n := len(s)
 	for i := 1; i < n; i++ {
 		j := i
 		for j > 0 {
-			if s[j-1] > s[j] {
-				s[j-1], s[j] = s[j], s[j-1]
+			if sliceCopy[j-1] > sliceCopy[j] {
+				sliceCopy[j-1], sliceCopy[j] = sliceCopy[j], sliceCopy[j-1]
 			}
 			j--
 		}
 	}
-	return s
+	return sliceCopy
 }
 
 func main() {
 	fmt.Println(bubbleSort([]int{3, 2, 1, 5}))
-	fmt.Println(insertionSort([]int{3, 2, 1, 5, -1}))
+	fmt.Println(insertionSort([]int{3, 2, 1, 5}))
 }
--- a/02_bubble/sarahdavy1983/main.go
+++ b/02_bubble/sarahdavy1983/main.go
@@ -2,13 +2,17 @@ package main
 
 import (
 	"fmt"
+	"io"
+	"os"
 )
 
+var out io.Writer = os.Stdout
+
 func bubbleSort(s []int) []int {
-	sliceCopy := append(s[:0:0], s...)
+
+	var sliceCopy = append(s[:0:0], s...)
 	n := len(s)
-	swapped := true
-	for swapped {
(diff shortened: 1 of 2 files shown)


02_bubble/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+var tests = []struct {
+	name  string
+	input []int
+	want  []int
+}{
+	{"valid", []int{3, 2, 1, 5}, []int{1, 2, 3, 5}},
+	{"same number", []int{5, 5, 5, 5}, []int{5, 5, 5, 5}},
+	{"blank", []int{}, []int{}},
+	{"sorted", []int{1, 2, 3, 4}, []int{1, 2, 3, 4}},
+}
```
jessodri, line 33: Looks great to me!  My only suggestion would be to add a test case for if the array contains negative numbers :)

Sarahdavy1983: Thanks!!  This is now updated :)

02_bubble/sarahdavy1983/main.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"fmt"
+)
+
+func bubbleSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
```
joshcarp: Should use `var SliceCopy = s` or ` Slicecopy := s` to copy slice

02_bubble/sarahdavy1983/main.go
```diff
@@ -0,0 +1,40 @@
...
+			if sliceCopy[i-1] > sliceCopy[i] {
+				sliceCopy[i], sliceCopy[i-1] = sliceCopy[i-1], sliceCopy[i]
+				swapped = true
+			}
+		}
+	}
+	return sliceCopy
+}
+func insertionSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
```
joshcarp: same here

02_bubble/sarahdavy1983/main.go
joshcarp: also there isn't any reason why you needed sliceCopy to start with; you can do it in place

02_bubble/sarahdavy1983/main.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+)
+
+func bubbleSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
+	n := len(s)
+	swapped := true
+	for swapped {
```
joshcarp: I like the use of this; although to simplify we could use 
`for swapped := true; swapped;{ `if we wanted to condense a little more

02_bubble/sarahdavy1983/main.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+	}
+	return sliceCopy
+}
+func insertionSort(s []int) []int {
+	sliceCopy := append(s[:0:0], s...)
+	n := len(s)
+	for i := 1; i < n; i++ {
+		j := i
+		for j > 0 {
```
joshcarp: this for loop can be consensed to declare j inline:
`for j := i; j > 0; {` 
Note that you might have tried this, but maybe just needed to add the trailing `;`

02_bubble/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestMainOut(t *testing.T) {
+
+	main()
+
+	expected := []int{1, 2, 3, 5}
+	actual := []int{1, 2, 3, 5}
```
joshcarp: You're not testing anything here: you're going to need to capture the output

02_bubble/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,48 @@
...
+}{
+	{"valid", []int{3, 2, 1, 5}, []int{1, 2, 3, 5}},
+	{"same number", []int{5, 5, 5, 5}, []int{5, 5, 5, 5}},
+	{"blank", []int{}, []int{}},
+	{"sorted", []int{1, 2, 3, 4}, []int{1, 2, 3, 4}},
+	{"negative number", []int{1, -5, 6, 9}, []int{-5, 1, 6, 9}},
+}
+
+func TestBubble(t *testing.T) {
+	for _, tc := range tests {
```
joshcarp: nitpick: tc could be changed with `test` as you're iterating through a list of `tests` and the singular of `tests` could be `test`
just gives more context to what `tc` is, although this is probablty really pedantic

joshcarp (changes requested):
Nice!

02_bubble/sarahdavy1983/main.go
Sarahdavy1983: @Joshcarp 
Thanks for the feedback, I have updated as per most of your comments.  I just want to clarify what you mean by " also there isn't any reason why you needed sliceCopy to start with; you can do it in place"?  Per my understanding of the requirements we need to create a copy

02_bubble/sarahdavy1983/main.go
```diff
@@ -0,0 +1,43 @@
...
+}
+func insertionSort(s []int) []int {
+	var sliceCopy = append(s[:0:0], s...)
+	n := len(s)
+	for i := 1; i < n; i++ {
+		for j := i; j > 0; {
+			if sliceCopy[j-1] > sliceCopy[j] {
+				sliceCopy[j-1], sliceCopy[j] = sliceCopy[j], sliceCopy[j-1]
+			}
+			j--
```
camscale, line 34: just a small nitpick: this should go in the `for` loop:
```
for j := i; j > 0; j-- {
...
}
```

02_bubble/sarahdavy1983/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	name  string
+	input []int
+	want  []int
+}{
+	{"valid", []int{3, 2, 1, 5}, []int{1, 2, 3, 5}},
+	{"same number", []int{5, 5, 5, 5}, []int{5, 5, 5, 5}},
+	{"blank", []int{}, []int{}},
+	{"sorted", []int{1, 2, 3, 4}, []int{1, 2, 3, 4}},
+	{"negative number", []int{1, -5, 6, 9}, []int{-5, 1, 6, 9}},
+}
```
camscale, line 33: I would suggest adding a test case for a 1-element slice and a 2-element slice. This is because "1" is a boundary condition for the length of the slice - you start counting at 1 and use `<`. If you get the logic slightly wrong in the code, it's easy to be off by one.

These off by one errors (boundary conditions) are common enough that you should look at the comparisons in your code and add test cases for either side of that boundary.

This is a general software engineering comment, not Go-specific, so this won't block approval if you would prefer to merge and move on. Just consider it for later labs.

camscale (changes requested):
First, I am so sorry we've dropped the ball on reviewing your PR. The amount of time that this has been waiting for approval is way too long. We're working through the backlog and hope to keep the PR review wait time much shorter.

This code is very good. There's really only one little thing I would change which makes the insertion sort loop a little clearer. I'm keeping an eye on this so I'm ready to approve as soon as you push a new change. If you are on our slack channel (#go-course) you can ping me there too (camh) for a quick re-ewview.

juliaogris:

PR #283, opened 2019-06-06
Make the IPAddr type implement fmt.Stringer

lexual:
Fixes #282 

Review of colleague's PR #247

TODO

PR #309, opened 2019-06-09
Implement puppy CRUD

### LINKED ISSUE #310: Puppy CRUD
- [X] Implement a `Puppy` struct containing `ID`, `Breed`, `Colour`, `Value`
- [X] Create `Storer` interface with crud methods for `Puppy`
- [X] Implement `MapStore` a `Storer` backed by a `map`
- [X] implementation `SyncStore` a `Storer` backed by a `sync.Map`
- [X] Add test suite for `Storer` interface testing both implementations

nicholascross:
Fixes #310

Review of colleague's PR #283

- Create `Storer` interface with crud methods for `Puppy`
- Implement `MapStore` a `Storer` backed by a `map`
- implementation `SyncStore` a `Storer` backed by a `sync.Map`
- Add test suite for `Storer` interface testing both implementations

### DIFF
--- a/06_puppy/nicholascross/main.go
+++ b/06_puppy/nicholascross/main.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+type Puppy struct {
+	ID     uint64
+	Breed  string
+	Colour string
+	Value  float64
+}
+
+type Storer interface {
+	CreatePuppy(p Puppy)
+	RetrievePuppy(uint64) *Puppy
+	UpdatePuppy(uint64, Puppy)
+	DeletePuppy(uint64)
+}
+
+type MapStore map[uint64]Puppy
+
+var out io.Writer = os.Stdout
+
+func main() {
+	p := Puppy{1, "Jack Russel Terrier", "white and brown", 550.0}
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) {
+	if _, exists := ms[p.ID]; !exists {
+		ms[p.ID] = p
+	}
+}
+
+func (ms MapStore) RetrievePuppy(id uint64) *Puppy {
+	if _, exists := ms[id]; exists {
+		pup := ms[id]
+		return &pup
+	}
+	return nil
+}
+
+func (ms MapStore) UpdatePuppy(id uint64, p Puppy) {
+	if _, exists := ms[id]; exists {
+		p.ID = id
+		ms[p.ID] = p
+	}
+}
+
+func (ms MapStore) DeletePuppy(id uint64) {
+	if _, exists := ms[id]; exists {
+		delete(ms, id)
+	}
+}
+
+type SyncStore struct {
+	store *sync.Map
+}
+
+func (s SyncStore) CreatePuppy(p Puppy) {
+	s.store.LoadOrStore(p.ID, p)
+}
+
+func (s SyncStore) RetrievePuppy(id uint64) *Puppy {
+	if pup, exists := s.store.Load(id); exists {
+		p := pup.(Puppy)
+		return &p
+	}
+	return nil
+}
+
+func (s SyncStore) UpdatePuppy(id uint64, p Puppy) {
+	if _, exists := s.store.Load(id); exists {
+		p.ID = id
+		s.store.Store(p.ID, p)
+	}
+}
+
+func (s SyncStore) DeletePuppy(id uint64) {
+	if _, exists := s.store.Load(id); exists {
+		s.store.Delete(id)
+	}
+}
--- a/06_puppy/nicholascross/main_test.go
+++ b/06_puppy/nicholascross/main_test.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type StorerTestSuite struct {
+	suite.Suite
+	storer Storer
+}
+
+func (suite *StorerTestSuite) SetupTest() {
+	suite.storer.CreatePuppy(Puppy{1, "Jack Russel", "white", 500.0})
+}
+
+func (suite *StorerTestSuite) TestCreate() {
+	suite.storer.CreatePuppy(Puppy{12, "Komondor", "white", 1000.0})
+	puppy := suite.storer.RetrievePuppy(12)
+	suite.Equal(uint64(12), puppy.ID)
+	suite.Equal("Komondor", puppy.Breed)
+	suite.Equal("white", puppy.Colour)
+	suite.Equal(1000.00, puppy.Value)
+}
+
+func (suite *StorerTestSuite) TestRead() {
+	puppy := suite.storer.RetrievePuppy(1)
+	suite.Equal(uint64(1), puppy.ID)
+	suite.Equal("Jack Russel", puppy.Breed)
+	suite.Equal("white", puppy.Colour)
+	suite.Equal(500.00, puppy.Value)
+}
+
+func (suite *StorerTestSuite) TestUpdate() {
+	suite.storer.UpdatePuppy(1, Puppy{1, "Jack Russel Terrier", "white and brown", 550.0})
+	puppy := suite.storer.RetrievePuppy(1)
+	suite.Equal(uint64(1), puppy.ID)
+	suite.Equal("Jack Russel Terrier", puppy.Breed)
+	suite.Equal("white and brown", puppy.Colour)
+	suite.Equal(550.00, puppy.Value)
+}
+
+func (suite *StorerTestSuite) TestDelete() {
+	suite.storer.DeletePuppy(1)
+	puppy := suite.storer.RetrievePuppy(1)
+	suite.Nil(puppy)
+}
+
+func TestMapStorer(t *testing.T) {
+	s := StorerTestSuite{
+		storer: MapStore{},
+	}
+	suite.Run(t, &s)
+}
+
+func TestSyncStorer(t *testing.T) {
+	store := sync.Map{}
+	s := StorerTestSuite{
+		storer: SyncStore{
+			store: &store,
+		},
+	}
+	suite.Run(t, &s)
+}
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("1 - Jack Russel Terrier [white and brown]: 550.000000")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output. Expected: %q - Actual: %q", expected, actual)
+	}
+}


06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+import "sync"
+
+type ID uint64
+type Breed string
+type Colour string
+type Value float64
+
+type Puppy struct {
+	Id     ID
```
golangcibot: struct field `Id` should be `ID` (from `golint`)

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,99 @@
...
+	pexist := ms[p.ID]
+	if pexist != nil {
+		ms[p.ID] = &p
+	}
+}
+
+func (ms MapStore) DeletePuppy(id ID) {
+	pexist := ms[id]
+	if pexist != nil {
+		ms[id] = nil
```
kasunfdo: This will not delete the map entry. Instead, use `delete(ms, id)`

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (ms MapStore) UpdatePuppy(id ID, p Puppy) {
+	pexist := ms[p.ID]
+	if pexist != nil {
+		ms[p.ID] = &p
+	}
+}
+
+func (ms MapStore) DeletePuppy(id ID) {
+	pexist := ms[id]
```
kasunfdo: Instead of checking null, checking whether key exists will be better.
```
if val, ok := ms[id]; ok {
    ...
}
```

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,99 @@
...
+		ms[id] = nil
+	}
+}
+
+type SyncStore struct {
+	store *sync.Map
+}
+
+func (s SyncStore) CreatePuppy(p Puppy) {
+	_, found := s.store.Load(p.ID)
```
kasunfdo: `LoadOrStore` method comes in handy for this scenario

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,99 @@
...
+	if pexist == nil {
+		ms[p.ID] = &p
+	}
+}
+
+func (ms MapStore) RetrievePuppy(id ID) *Puppy {
+	return ms[id]
+}
+
+func (ms MapStore) UpdatePuppy(id ID, p Puppy) {
```
kasunfdo: First argument is unused and can be removed. However, with a little bit of foresight, following on lab 10
`PUT    /api/puppy/{id}      Payload: Puppy JSON without ID`
Therefore, setting id as p.ID would be better. 😉

06_puppy/nicholascross/main.go
kasunfdo: Use it in create, update, and read as well.

kasunfdo (changes requested):
Nice work and good coding style.
It is a good practice to check key existence in map instead of comparing returned value with default value.

I do not think this is mandatory, but having a bool return with success/failure status will be awesome. However, you'll have to test both scenarios in test suit

06_puppy/nicholascross/main.go
nicholascross: Oh, I didn't notice that 😄

@kasunfdo are you suggesting that I should do something like

```
p.ID = id
```

or just remove the parameter from the function?

```
func (ms MapStore) UpdatePuppy(p Puppy) 
```

kasunfdo: Yes. Better using a puppy instance without id as the second arg, setting `p.ID = id` if id is not in store. 
It will be beneficial for the future.

kasunfdo (dismissed):
LGTM

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,88 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+type ID uint64
```
anzboi: I wouldn't bother creating new type aliases for these.

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type Storer interface {
+	CreatePuppy(p Puppy)
+	RetrievePuppy(ID) *Puppy
+	UpdatePuppy(ID, Puppy)
+	DeletePuppy(ID)
+}
+
+type MapStore map[ID]*Puppy
```
anzboi: I'd advise against pointers for in memory storage. The reason is that it is safer to handle ([playspace](https://goplay.space/#O4dHwcB5_Ak))

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,88 @@
...
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) {
+	if _, ok := ms[p.ID]; !ok {
+		ms[p.ID] = &p
+	}
+}
+
+func (ms MapStore) RetrievePuppy(id ID) *Puppy {
```
anzboi: By returning a pointer, you allow the caller to modify the underlying data. You can get around this by copying the stored data into a new variable and returning a pointer to that
```
pup := *ms[id] // dereference and copy raw into new variable
return &pup // return pointer to new variable
```
or using a raw store.

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,88 @@
...
+		delete(ms, id)
+	}
+}
+
+type SyncStore struct {
+	store *sync.Map
+}
+
+func (s SyncStore) CreatePuppy(p Puppy) {
+	s.store.LoadOrStore(p.ID, p)
```
anzboi, line 65: Neat

anzboi (changes requested):
Great work so far, Tests are perfect.
There is an issue with the map store, you can actually modify the stored data by modifying the pointer returned by `RetrievePuppy`. The recommended fix is to use raw structs in the map store (`type MapStore map[ID]Puppy`), though you can also fix it by carefully copying the data where needed.
Here are a couple of go play space links highlighting the issue and providing some solutions. I encourage you to play around with them and come up with a scheme that is safe to use.
[issue](https://goplay.space/#O4dHwcB5_Ak)
[rawstore](https://goplay.space/#wB4QssvTojs)

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	p := Puppy{1, "Jack Russel Terrier", "white and brown", 550.0}
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) {
+	if _, ok := ms[p.ID]; !ok {
+		ms[p.ID] = p
```
alfredxiao, line 35: In this context, the name of `ok` feels a bit strange. For Retrieve/Update/Delete it makes sense, for this, perhaps name it as `existing` or something like that.

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+	Colour string
+	Value  float64
+}
+
+type Storer interface {
+	CreatePuppy(p Puppy)
+	RetrievePuppy(uint64) *Puppy
+	UpdatePuppy(uint64, Puppy)
+	DeletePuppy(uint64)
+}
```
alfredxiao, line 22: None of your interface functions return errors. So how does function caller know something is wrong? For example, update or delete with non-existing id?

nicholascross: This is working as intended, I have already implemented error handling in the subsequent lab.  I don't want to double up on work so unless this is going to be a blocker I will leave it as is.

06_puppy/nicholascross/main.go
nicholascross: I have updated the variable name for all relevant cases from `ok` to `exists`

camscale: Actually @alfredxiao , `ok` is the most common name used here, and referred to as the "comma ok" idiom (as described in Effective Go, in the Maps section: https://golang.org/doc/effective_go.html#maps). You'll see it used quite widely across the Go standard library.

However, I find `exists` to be ok to, so don't take this comment as a reason to change it back.

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (ms MapStore) CreatePuppy(p Puppy) {
+	if _, exists := ms[p.ID]; !exists {
+		ms[p.ID] = p
+	}
+}
+
+func (ms MapStore) RetrievePuppy(id uint64) *Puppy {
+	if _, exists := ms[id]; exists {
+		pup := ms[id]
```
camscale, line 41: There is no need to look up the map again. You could have done:
```
if pup, exists := ms[id]; exists {
        return &pup
}
return nil
```

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+		return &pup
+	}
+	return nil
+}
+
+func (ms MapStore) UpdatePuppy(id uint64, p Puppy) {
+	if _, exists := ms[id]; exists {
+		p.ID = id
+		ms[p.ID] = p
```
camscale, line 50: If `ID` is changed, you leave the puppy in the map under the old ID too.

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+}
+
+func (ms MapStore) DeletePuppy(id uint64) {
+	if _, exists := ms[id]; exists {
```
camscale, line 55: There is no need for this check. It is safe to delete an element that does not exist: https://golang.org/pkg/builtin/#delete

06_puppy/nicholascross/main.go
```diff
@@ -0,0 +1,87 @@
...
+		p := pup.(Puppy)
+		return &p
+	}
+	return nil
+}
+
+func (s SyncStore) UpdatePuppy(id uint64, p Puppy) {
+	if _, exists := s.store.Load(id); exists {
+		p.ID = id
+		s.store.Store(p.ID, p)
```
camscale, line 79: This has the same problem I mentioned on `MapStore` - a changing ID leaves the map still with the old ID.

06_puppy/nicholascross/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+func TestMapStorer(t *testing.T) {
+	s := StorerTestSuite{
+		storer: MapStore{},
+	}
+	suite.Run(t, &s)
+}
+
+func TestSyncStorer(t *testing.T) {
+	store := sync.Map{}
```
camscale, line 61: Requiring the creator of the `SyncStore` to create a `sync.Map` feels like it is exposing too much to the caller. Instead you could have a constructor for it `NewSyncStore` that returns a `SyncStore` initialised with a `sync.Map` already created.

06_puppy/nicholascross/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+	suite.Run(t, &s)
+}
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("1 - Jack Russel Terrier [white and brown]: 550.000000")
```
camscale, line 76: This test is relying on the default of `%f` to be 6 (six decimal places). I don't know if this default is guaranteed to be stable, so this test is possibly fragile. You could use `%g` in `main` which prints with the minimum necessary precision (which would print `550` in this case), or you could specify the precision and test against what was specified. Or you could just leave it - perhaps you want main to use whatever the default is. But in that case, I'd suggest a comment here saying you are relying on the default precision of `%f` to be 6, so if the default does change, it is easy to understand the reason for the failure.

No need to change anything in response to this comment - just an observation and something to consider.

camscale (changes requested):
A couple of changes I'd like to see to make better use of Go's features and ways of working. The ones about updating the ID of a puppy I would like to see because you do mostly handle the case, but it just needs a little more.

juliaogris:

PR #296, opened 2019-06-08
Lab 3

joshcarp:
Fixes #295 

Review of colleague's PR #

#### Changes proposed in this PR:

- Complete Lab 3
 -  Add letters function
 -  Add sortLetters function
- Add Unit Tests for 100% test coverage

PR #320, opened 2019-06-09
Lab03

### LINKED ISSUE #319: implement letters for lab03
Complete Lab 3 with two functions
- letters function to return a map of runes with their counts
- sortLetters function to sort the map returned by letters
Add Unit Tests for 100% test coverage

jonathanheaden:
Fixes #319 

Review of colleague's PR #296 

Implement letters function
Implement sortletters 
## -
This uses a sorting functionality of ints from the sort package


03_letters/jonathanheaden/main.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func sortLetters(m map[rune]int) []string {
+	var rv = make([]string, 0)
+	var keys = make([]int, 0, len(m))
+	var sm = make(map[int]int)
+	for k := range m {
+		sm[int(k)] = m[k]
```
joel00wood, line 19: This loop seems somewhat unnecessary, by my understanding it loops through the map `m` and then transfer the counts from `m` to int map `sm` along with int slice `keys`, after which point it sorts the `keys` slice. Then it fills out the string slice `rv` by looping through `keys` and appending to `rv` accordingly. 

While I'm not 100% certain I'm fairly sure that any non alphanumeric characters will break this, particularly characters from non-English alphabets and/or emojis.

Similarly though this could be simplified down to:
```
for k, count := range m {
  rs := string(k)
  rc := strconv.Itoa(count)
  rv = append(r, fmt.Sprintf(%s:%d", rs, rc))
}
sort.Strings(rv)
```

joel00wood (commented):
I'm not sure that the `sortLetters()` function will work with anything outside of standard/English language based alphanumeric characters. I've included a comment that both simplifies and _should_ address this. Namely, you were sorting the keys as integers however to my knowledge the equivalent sort as strings should get the same result given the runes directly correspond to the 'value' of the alphanumeric character. 

That said, even when reviewing my own submission against the requirements it's not immediately obvious in the lab description as to whether it expects the slice to be sorted based on the value of the strings within the slice or the count of the rune per slice entry.

Additionally it doesn't look like there is explicit testing for `sortLetters()` although mysteriously the coverage checks don't seem to mind. To that effect while useful it is to allow for implicit coverage, it's always better to explicitly test individual functions to show they're behaving as designed under a range of inputs.

juliaogris:

PR #308, opened 2019-06-09
Lab 2 - Bubble sort

jessodri:
Fixes #307 

Review of colleague's PR #279 

#### Changes proposed in this PR:

- Add function to sort array by bubble sort algorithm
- Add functions to sort array by insertion and O(n log(n)) algorithms
- Add tests for all functions

PR #325, opened 2019-06-10
Lab 2 - Adding multiple sort functions

### LINKED ISSUE #398: Lab 2 - Multiple sort functions
Implementing the following sort functions:
Bubble
Insertion
Mergesort

brnbke:
Fixes #398

Review of colleague's PR #308

Implementing and testing following sort functions:
Bubble
Insertion 
Mergesort

### DIFF
--- a/02_bubble/brnbke/main.go
+++ b/02_bubble/brnbke/main.go
@@ -1,0 +1,91 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(arr []int) []int {
+	changed := true
+	r := make([]int, len(arr))
+	copy(r, arr)
+
+	for {
+		if !changed {
+			break
+		}
+		changed = false
+		for i := 0; i < len(r)-1; i++ {
+			if r[i] > r[i+1] {
+				r[i], r[i+1] = r[i+1], r[i]
+				changed = true
+			}
+		}
+	}
+
+	return r
+}
+
+func insertion(arr []int) []int {
+	result := make([]int, len(arr))
+	copy(result, arr)
+
+	for i := 1; i < len(result); i++ {
+		tmp := result[i]
+		for j := i - 1; j > -1; j-- {
+			if result[j] > tmp {
+				result[j+1], result[j] = result[j], tmp
+			} else {
+				break
+			}
+		}
+	}
+
+	return result
+}
+
+func mergesort(arr []int) []int {
+	if len(arr) <= 1 {
+		return arr
+	}
+
+	// work out the mid point of array
+	i := len(arr) / 2
+
+	// split input arr into two halves and recursively call merge sort to sort each half
+	l := mergesort(arr[:i])
+	r := mergesort(arr[i:])
+
+	// merge two sorted halves back together.
+	return merge(l, r)
+}
+
+func merge(l []int, r []int) []int {
+
+	result := []int{}
+
+	for len(l) > 0 && len(r) > 0 {
+		if l[0] <= r[0] {
+			result = append(result, l[0])
+			l = l[1:]
+		} else {
+			result = append(result, r[0])
+			r = r[1:]
+		}
+	}
+
+	if len(l) > 0 {
+		result = append(result, l...)
+	} else {
+		result = append(result, r...)
+	}
+
+	return result
+}
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
--- a/02_bubble/brnbke/main.go
+++ b/02_bubble/brnbke/main.go
@@ -9,20 +9,19 @@ import (
 var out io.Writer = os.Stdout
 
 func bubble(arr []int) []int {
-	changed := true
 	r := make([]int, len(arr))
 	copy(r, arr)
 
 	for {
-		if !changed {
-			break
-		}
-		changed = false
+		changed := false
 		for i := 0; i < len(r)-1; i++ {
 			if r[i] > r[i+1] {
 				r[i], r[i+1] = r[i+1], r[i]
 				changed = true
 			}
+		}
+		if !changed {
+			break
 		}
 	}
 
@@ -35,12 +34,11 @@ func insertion(arr []int) []int {
 
 	for i := 1; i < len(result); i++ {
 		tmp := result[i]
-		for j := i - 1; j > -1; j-- {
-			if result[j] > tmp {
-				result[j+1], result[j] = result[j], tmp
-			} else {
+		for j := i - 1; j >= 0; j-- {
+			if result[j] <= tmp {
 				break
 			}
+			result[j+1], result[j] = result[j], tmp
 		}
 	}
 
@@ -63,9 +61,9 @@ func mergesort(arr []int) []int {
 	return merge(l, r)
 }
 
-func merge(l []int, r []int) []int {
+func merge(l, r []int) []int {
 
-	result := []int{}
+	result := make([]int, 0, len(l)+len(r))
 
 	for len(l) > 0 && len(r) > 0 {
 		if l[0] <= r[0] {
--- a/02_bubble/brnbke/main_test.go
+++ b/02_bubble/brnbke/main_test.go
@@ -36,9 +36,6 @@ var testCases = map[string]struct {
 func TestBubble(t *testing.T) {
 	for name, test := range testCases {
 		test := test
-		var buf bytes.Buffer
-
-		out = &buf
 		t.Run(name, func(t *testing.T) {
 			b := bubble(test.input)
 			result := reflect.DeepEqual(test.want, b)
@@ -52,8 +49,6 @@ func TestBubble(t *testing.T) {
 func TestInsertion(t *testing.T) {
 	for name, test := range testCases {
 		test := test
-		var buf bytes.Buffer
-		out = &buf
 		t.Run(name, func(t *testing.T) {
 			b := insertion(test.input)
 			result := reflect.DeepEqual(test.want, b)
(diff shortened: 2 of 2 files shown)


02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,92 @@
...
+	return result
+}
+
+func mergesort(arr []int) []int {
+	if len(arr) <= 1 {
+		return arr
+	}
+
+	// work out the mid point of array
+	i := int(math.Floor(float64(len(arr) / 2)))
```
golangcibot: SA4015: calling math.Floor on a converted integer is pointless (from `staticcheck`)

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
```
joel00wood, line 9: Is the purpose of this purely to simplify testing?

02_bubble/brnbke/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+	expected := strconv.Quote("[1 2 3 5]\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main(), expected %v, sadly got %v", expected, actual)
+	}
+}
+
+var testCases = map[string]struct {
+	input []int
```
joel00wood, line 25: Given that these two pieces of the struct are of the same type this could be simplified to a single line

02_bubble/brnbke/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+			if !result {
+				t.Errorf("expected %v, got %v", test.want, b)
+			}
+		})
+	}
+}
+
+func TestMerge(t *testing.T) {
+	var testCases = map[string]struct {
+		left  []int
```
joel00wood, line 77: Similar to the above comment, this could be simplified to `left, right, want []int` as a single line

02_bubble/brnbke/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+		"right only":    {left: []int{}, right: []int{1, 6, 2, 0}, want: []int{1, 6, 2, 0}},
+	}
+	for name, test := range testCases {
+		test := test
+		var buf bytes.Buffer
+		out = &buf
+		t.Run(name, func(t *testing.T) {
+			b := merge(test.left, test.right)
+			result := reflect.DeepEqual(test.want, b)
+			if !result {
```
joel00wood, line 93: Is it necessary to assign the reflect response to a value, or is this matter of `gofmt/golangci-lint` eccentricity

joel00wood (commented):
Most of my comments are essentially nitpicking with respect to formatting and also some curiosity, looks good to me otherwise.

02_bubble/brnbke/main.go
brnbke: Yes pretty much, so I can change the output on subsequent Fprintln to "out" and grab that in the test.

02_bubble/brnbke/main_test.go
brnbke: Yes, you're right, did not notice that, though not sure what the expected way of doing this is in go TBH.

02_bubble/brnbke/main_test.go
brnbke: No good reason that I can see actually, just habit I think

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+func bubble(arr []int) []int {
+	changed := true
+	r := make([]int, len(arr))
+	copy(r, arr)
+
+	for {
+		if !changed {
```
samuong: If you moved this if condition to the bottom of the loop body, it would still mean the same thing. But then you could also move the `changed` variable into the scope of the `for` loop.

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
...
+	result := make([]int, len(arr))
+	copy(result, arr)
+
+	for i := 1; i < len(result); i++ {
+		tmp := result[i]
+		for j := i - 1; j > -1; j-- {
+			if result[j] > tmp {
+				result[j+1], result[j] = result[j], tmp
+			} else {
+				break
```
samuong, line 39: nitpick: I'd suggest only indenting the `break` block, and having the other one outside the `if` statement. https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// split input arr into two halves and recursively call merge sort to sort each half
+	l := mergesort(arr[:i])
+	r := mergesort(arr[i:])
+
+	// merge two sorted halves back together.
+	return merge(l, r)
+}
+
+func merge(l []int, r []int) []int {
```
samuong: You don't need to specify `[]int` for the first parameter.

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	result := []int{}
```
samuong: Given that you know the size of the result (`len(l)+len(r)`), you could create a pre-allocated array here. Then you could guarantee that you never need to regrow the array through a call to `append`.

02_bubble/brnbke/main_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"already sorted":  {input: []int{0, 1, 2, 6}, want: []int{0, 1, 2, 6}},
+	"reverse sorted":  {input: []int{6, 2, 1, 0}, want: []int{0, 1, 2, 6}},
+}
+
+func TestBubble(t *testing.T) {
+	for name, test := range testCases {
+		test := test
+		var buf bytes.Buffer
+
+		out = &buf
```
samuong: What's the purpose of overriding `out` in this test (as well a the others)? I think `main` is the only function that uses it, but let me know if I've missed something.

02_bubble/brnbke/main.go
camscale: As Sam says, since this condition is the first thing in the loop and it breaks the loop, you may as well put it in the condition of the loop:
```
for changed {
...
}
```
It does read a bit funny as Go doesn't use "while" for this, opting to use "for" for all loops.

02_bubble/brnbke/main.go
```diff
@@ -0,0 +1,91 @@
...
+	return r
+}
+
+func insertion(arr []int) []int {
+	result := make([]int, len(arr))
+	copy(result, arr)
+
+	for i := 1; i < len(result); i++ {
+		tmp := result[i]
+		for j := i - 1; j > -1; j-- {
```
camscale: nit: `j > -1` is a little odd. I'd use `j >= 0`, since 0 has meaning but -1 does not (except as being 1 off the meaningful number).

02_bubble/brnbke/main.go
camscale: I agree with Sam. By using:
```
if result[j] <= tmp {
  break
}
result[j+1], result[j] = result[j], tmp
```
It makes it clear what the terminating condition is - I don't need to mentally invert the condition in my head to understand the condition of the special case.

It is also a variation of aligning the happy path to the left edge: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88 (modulo loop indentation).

camscale (changes requested):
I didn't review the merge sort, just the ones requested for the lab (i ain't got time for that!).

I've chimed in with Sam on a couple of comments which relate to idiomatic Go, so I'm going to request changes for those. The rest you can take or leave (but I suggest taking)

brnbke:
Reopening PR as closed in error

02_bubble/brnbke/main.go
brnbke: have moved the check as suggested so var now just in the scope of the for loops

02_bubble/brnbke/main.go
brnbke: Done

02_bubble/brnbke/main_test.go
brnbke: A mistake to be honest, copy paste error, removed.

02_bubble/brnbke/main.go
brnbke: Moved the check to the bottom of the look to keep the changed var scoped to the loops, not sure I can implement it as "for changed" as need to reset the changed to false each loop and only exit if not changed

02_bubble/brnbke/main.go
brnbke: Implemented.

@samuong seems you were not being too nitpicky with this suggestion either.

02_bubble/brnbke/main.go
camh-anz: You could do:
```
for changed := true; changed; {
    changed = false
    ...
}
```
if you want to keep the var scoped to the loop. But this is fine as is. I won't insist on any further change here.

camscale: [sorry if you get this twice. I wanted to add it in the review, but the comment has gone from the review.]

You could do:
```
for changed := true; changed; {
    changed = false
    ...
}
```
if you want to keep the var scoped to the loop. But this is fine as is. I won't insist on any further change here.

camscale (approved):
LGTM. As per usual, could you either rebase-squash-force-push your changes to remove fixup commits, or just do a squash merge when you merge.

02_bubble/brnbke/main.go
samuong: Hah, good to know I'm not alone! 😄

02_bubble/brnbke/main.go
brnbke: Ah I see what you mean, I assume this is the prefered Go way of doing this. will take note for future, but will leave as is for this to avoid further reviews.

camscale: I'm not sure if Go-style really has much to say about this, but I prefer to avoid infinite loops with the terminating condition in the body - it's a bit easier to read and know that the loop will actually terminate. In this particular case it's easy enough to tell, but not always.

PR #327, opened 2019-06-10
Lab 4 - Numeronyms

### LINKED ISSUE #326: Lab 4 - Numeronyms

jessodri:
Fixes #326

Review of colleague's PR #554

- Add function to return numeronyms of strings
- Add tests


04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	numeronyms := []string{}
+	for _, v := range vals {
+		len := len(v) - 2
+		if len > 1 {
+			first := string(v[0])
```
kasunfdo: You can use string formatter to make concatenation easy.
`result := fmt.Sprintf("%c%d%c", v[0], len, v[len+1])`

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	expected := strconv.Quote("[a11y K8s abc]\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
+		t.Errorf("expected %v but got %v", expected, actual)
+	}
+}
+
+func TestSortLetters(t *testing.T) {
```
kasunfdo: Good unit test coverage.

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	testCases := []struct {
+		description string
+		word        string
```
kasunfdo: As `numeronyms(vals ...string)` is a variadic function, you can define input as `words []string` instead of `word string`. By doing that, you can add 'multiple string test' into testCases.
In that case, you can invoke function call as follows.
`numeronyms(tc.words...)`

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+	"strconv"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	numeronyms := []string{}
+	for _, v := range vals {
+		len := len(v) - 2
```
marksomething: Avoid using names that aren't quite what people would expect, perhaps move the `-2` to later in the code?

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		expected    []string
+	}{
+		{"happy path", "onomatopoeia", []string{"o10a"}},
+		{"3 letter word", "cat", []string{"cat"}},
+		{"empty", "", []string{""}},
+	}
```
marksomething, line 53: What happens if the string contains some multi-byte characters ❓

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for _, tc := range testCases {
+		expected := tc.expected
+		actual := numeronyms(tc.word)
+		t.Run(tc.description, func(t *testing.T) {
+			assert.Equal(t, actual, expected, "got %v but wanted %v", actual, expected)
+		})
+	}
+
+	t.Run("test with multiple strings passed in", func(t *testing.T) {
```
marksomething: Consider splitting this into a seperate test function or include in examples list above

04_numeronym/jessodri/main_test.go
marksomething: Rename?

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("[a11y K8s abc]\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
```
marksomething: You used `assert.equal()` in the second set of tests, why not here?

marksomething (changes requested):

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,31 @@
...
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	numeronyms := []string{}
+	for _, v := range vals {
+		len := len(v)
+		if len > 3 {
+			first := string(v[0])
+			middle := strconv.Itoa(len-2)
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+		})
+	}
+}
+func TestMultipleStrings(t *testing.T) {
+	t.Run("test with multiple strings passed in", func(t *testing.T) {
+		actual := numeronyms("K8bernetes", "dog", "chicken", "four", "hello")
+		expected := []string{"K8s", "dog", "c5n", "f2r", "h3o"}
+		assert.Equal(t, actual, expected, "got %v but wanted %v", actual, expected)
+	})
+}
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

04_numeronym/jessodri/main.go
jessodri: Thanks for your suggestions!

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"io"
+	"os"
+	"strconv"
+	"unicode"
+)
+
+var out io.Writer = os.Stdout
+
+func isASCII(s string) bool {
+	for i := 0; i < len(s); i++ {
```
marksomething: `for c := range s` ?

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,41 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+	"unicode"
+)
+
+var out io.Writer = os.Stdout
+
+func isASCII(s string) bool {
```
marksomething: It probably simpler to support unicode than test for multibyte.

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("[a11y K8s abc]\n")
+	actual := strconv.Quote(buf.String())
+
+	if expected != actual {
```
marksomething: The if statement here is not needed

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		assert.Equal(t, actual, expected, "got %v but wanted %v", actual, expected)
```
marksomething: The definition of [`asset.Equal()`](https://godoc.org/github.com/stretchr/testify/assert#Equal) is for expected second and actual as the third parameter. It can be called as `assert.Equal(t, expected, actual)` unless there's a reason to want a specific output message.

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, tc := range testCases {
+		expected := tc.expected
+		actual := numeronyms(tc.word)
+		t.Run(tc.description, func(t *testing.T) {
+			assert.Equal(t, actual, expected, "got %v but wanted %v", actual, expected)
+		})
+	}
+}
+func TestMultipleStrings(t *testing.T) {
+	t.Run("test with multiple strings passed in", func(t *testing.T) {
```
marksomething: You can leave out the t.Run(...) here, it's only needed for sub-tests

marksomething (changes requested):

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,41 @@
...
+	for _ , c := range s {
```
golangcibot: File is not `goimports`-ed (from `goimports`)

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,42 @@
...
+}
+
+func TestNumeronyms(t *testing.T) {
+	testCases := []struct {
+		description string
+		word        []string
+		expected    []string
+	}{
+		{"happy path - one string", []string{"onomatopoeia"}, []string{"o10a"}},
+		{"happy path - multiple strings", []string{"onomatopoeia", "K8bernetes", "dog", "chicken", "four"}, []string{"o10a", "K8s", "dog", "c5n", "f2r"}},
```
golangcibot: line is 148 characters (from `lll`)

04_numeronym/jessodri/main_test.go
jessodri: Thanks! That is very useful to know!

04_numeronym/jessodri/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+func TestNumeronyms(t *testing.T) {
+	testCases := []struct {
+		description string
+		word        []string
+		expected    []string
+	}{
+		{
+			"happy path - one string",
```
golangcibot: File is not `goimports`-ed (from `goimports`)

04_numeronym/jessodri/main.go
jessodri: Probably, but I'm happy to leave it how it is for now.

marksomething (dismissed):

04_numeronym/jessodri/main.go
```diff
@@ -0,0 +1,37 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+func numeronyms(vals ...string) []string {
+	numeronyms := []string{}
+	for _, v := range vals {
+		len := len(v)
```
rokane: I feel this is quite a dangerous action, because it overwrites the built in `len` function to be an int value. You are not running into any issues in this scenario because the memory of the int len is released after each iteration of the loop. But if you were going to use the len() function further down you would run into issues.

I would advise against this, as this is not a good practice to get into. Are you able to rename the variable to something different? Perhaps `length`,`l` or alternatively use the `len(v)` without storing the value ?

For an example of what I am talking about see [Overwrite example](https://goplay.space/#hVjLs2Vw_kF)

04_numeronym/jessodri/main.go
rokane: I agree, I think this is complicating things more than what was intended. I don't believe it asked for only strings involving ascii code.

> Write a function that returns a slice of numeronyms for its input strings

rokane (dismissed):
Looking good so far @jessodri. I have requested one change, in the `numeronyms` function. On top of this, I do think having the ASCII check is a little overkill for this task. I am happy for you to leave it there, but if you do I would like you to add further edge cases around this. These cases might be things like, adding unicode characters, strings with a combination of ASCII and unicode chars etc..

Alternatively you can simplify it by removing this and not worry about the extra test cases as the complexity will be removed. I will leave that up to you, I am happy with either approach.

juliaogris (approved):
Lab is solved - LGTM.

Just some nitpicky feedback for the future:
IMO this lab is factored nicer if you have two functions, one taking care of slice/batch conversion, one doing the actual numeronym conversion, that way you can benefit from `return` vs more verbose `if-else`.

Also, when you instantiate your result numeronyms you already know its length - it's runtime effective to take advantage of it with `make([]string, len(input))` or at least specify capacity with `make([]string, 0, len(input))`:

```go
func numeronyms(s ...string) []string {
	result := make([]string, len(s))
	for i, str := range s {
		result[i] = numeronym(str)
	}
	return result
}

func numeronym(input string) string {
     if utf8.RuneCountInString(input) <= 3 {
        return input
    }
    firstRune, _ := utf8.DecodeRuneInString(v)
    lastRune, _ := utf8.DecodeLastRuneInString(v)
    return fmt.Sprintf("%c%d%c", firstRune, utf8.RuneCountInString(input), lastRune )
}
```

PR #251, opened 2019-06-04
Lab5 - Stringer

CameronSchafer:
Fixes #250

Review of colleague's PR #355

#### Changes proposed in this PR:

- Add fmt.stringer functionality
- Test fmt.stringer implementation

PR #329, opened 2019-06-10
Lab 5 - Stringer

### LINKED ISSUE #328: Lab 5 - Stringer

jessodri:
Fixes #328 

Review of colleague's PR #251 

- Add tests


05_stringer/jessodri/main.go
```diff
@@ -0,0 +1,19 @@
...
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type IPAddr [4]byte
+
+func (i IPAddr) String() string {
+	return fmt.Sprintf("%d.%d.%d.%d", i[0], i[1], i[2], i[3])
```
pfannerj, line 14: Mine is exactly the same... I was kinda hoping that you could use , i... rather than i[0], i[1], i[2], i[3] but it seems you can't do that without have something like this...  
func (i IPAddr) String() string {
    tmp := make([]interface{}, len(i))
    for j, val := range ip {
        tmp[j] = val
    }
    return fmt.Sprintf("%d.%d.%d.%d", tmp...)
}
So I ended up leaving as I had it...

05_stringer/jessodri/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+		description string
+		ip          IPAddr
+		expected    string
+	}{
+		{"happy path", IPAddr{127, 0, 0, 1}, "127.0.0.1"},
+		{"one value", IPAddr{127}, "127.0.0.0"},
+		{"two value", IPAddr{127, 0}, "127.0.0.0"},
+		{"three value", IPAddr{127, 0, 0}, "127.0.0.0"},
+		{"empty", IPAddr{}, "0.0.0.0"},
+	}
```
pfannerj, line 33: I included a max test too with 255.255.255.255 but it's not necessary

rokane: I agree with @pfannerj, can you please put a test for the max case `255.255.255.255` this ensures the implementation is done correctly using `byte`

05_stringer/jessodri/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+	for _, tc := range testCases {
```
rokane: Alternatively if you use the `map[string]struct{input IPAddr, expected string}` you can use the key to describe the test name and you don't have to access it from the struct.

```go
for name, tc := range testCases {
    t.Run(name, func(t *testing.T) { ... })
}
```

05_stringer/jessodri/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+		actual := tc.ip
+		expected := tc.expected
+		t.Run(tc.description, func(t *testing.T) {
+			actual := actual.String()
```
rokane: You can remove this and directly assign when initialising the variable above.
```go
actual := tc.ip.String()
```

rokane (changes requested):
Code is looking really good. If you can please add another test case as @pfannerj mentioned then this should be good to merge after that. You also haven't completed a review of someone elses PR. Can you please do that and attach the reference in the description.

rokane (approved):
Good work. Thanks for fixing up those last few things I suggested and hanging in there with the review process! Great stuff :)

PR #342, opened 2019-06-11

### LINKED ISSUE #341: Lab 6 - CRUD puppy with interface
Implement CRUD puppy using interfaces

kasunfdo:
Fixes #341

Review of colleague's PR #309

- Implement CRUD puppy using interfaces
- Test suite for both implementations of Storer


06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+func (m MapStore) CreatePuppy(puppy *Puppy) bool {
+	if _, ok := m[puppy.ID]; !ok {
```
lexual: "exists" might be a better variable name here than "ok", makes the if logic become "!exists"

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
...
+// Storer defines standard CRUD operations for Puppy
+type Storer interface {
+	CreatePuppy(puppy *Puppy) bool
+	ReadPuppy(id uint64) (*Puppy, bool)
+	UpdatePuppy(id uint64, puppy *Puppy) bool
+	DeletePuppy(id uint64) bool
+}
+
+// MapStore type and implementation
+type MapStore map[uint64]*Puppy
```
lexual: I'm unsure (especially as I'm still a golang newbie) if this would be better as a value with type ```Puppy``` or ```*Puppy```.

Downside of having as a pointer, means, you store something in the Store, and then update that variable you passed to CreatePuppy, and what is stored in the Store has changed.

I'm unsure if this is a problem or not, if you feel confident in this design choice, feel free to make a comment, and resolve. It might something the higher-level reviewers might have an opinion on.

06_puppy/kasunfdo/syncstore.go
```diff
@@ -0,0 +1,37 @@
+package main
+
+func (s *SyncStore) CreatePuppy(puppy *Puppy) bool {
+	_, ok := s.Map.LoadOrStore(puppy.ID, puppy)
```
lexual: While I can see what this does, it took a little while to step through what was happening and why we were using LoadOrStore.

Potentially you could consider just doing a separate line that checks for existence by calling Load, and then calling plain old Store instead of LoadOrStore.

lexual:
I've conducted a review.

Work looks pretty great really.

I made a few comments, none of them prevent merging, but just for some things for you to consider, and up to you whether to take my comments on board, or ignore them.

Good work.

I'll label as 'changes-requested', but if you don't think changes are required then resolve my comments, and i'll mark as colleague-approved ;)

06_puppy/kasunfdo/mapstore.go
kasunfdo: I agree. :)

06_puppy/kasunfdo/types.go
kasunfdo: Thank you for pointing out the potential issue. Working with pointers should be done with extra care. Both pointer and value implementations have their on pros and cons as well. I'm open for a discussion about that.

06_puppy/kasunfdo/syncstore.go
kasunfdo: Valid point. Initially I was using Store method with checking for existence. After going through the sync map reference, I thought of using existing implementation and changed Store to LoadOrStore and added a comment to make it much clear.

06_puppy/kasunfdo/types.go
alfredxiao: Until requirement is clear, can't say either `Puppy` or `*Puppy` is right or wrong.

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,33 @@
...
+func (m MapStore) CreatePuppy(puppy *Puppy) bool {
+	if _, exists := m[puppy.ID]; !exists {
+		m[puppy.ID] = puppy
+		return true
+	}
+	return false
+}
+
+func (m MapStore) ReadPuppy(id uint64) (*Puppy, bool) {
+	if puppy, exists := m[id]; exists {
```
camscale: Why do you need this `if` statement. If the puppy exists, then `puppy, exists` is equal to the puppy and `true. If the puppy does not exist, then `puppy, exists` is equal to `nil, false`, so you could just do:
```
puppy, exists := m[id]
return puppy, exists
```

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,33 @@
...
+		return puppy, true
+	}
+	return nil, false
+}
+
+func (m MapStore) UpdatePuppy(id uint64, puppy *Puppy) bool {
+	if _, exists := m[id]; exists {
```
camscale: Should this be `m[id]` or `m[puppy.ID]`? On the next line, you set `puppy.ID` to id, which is good, because it keeps the `id` in the argument list and the ID in the puppy consistent, and also allows updating the ID of the puppy, but this means you need to look up the map with the old puppy ID, and then remove it from the map with the old id. Otherwise you now have the puppy stored with two IDs in the map.

06_puppy/kasunfdo/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+	suite.Suite
+	store Storer
+}
+
+func (suite *StorerTest) SetupTest() {
+	suite.store.CreatePuppy(&Puppy{101, "Labrador", "Cream", 2999.99})
+}
+
+func (suite *StorerTest) TestCreatePuppy() {
+	status := suite.store.CreatePuppy(&Puppy{101, "Labrador", "Cream", 2999.99})
```
camscale: `status` is not a good name for this variable. `created` makes a lot more sense and is more readable. I don't know what it means to say "the status was false" without understanding more about the code. "created is false" is clearer in its meaning.

06_puppy/kasunfdo/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+	suite.False(status)
+
+	status = suite.store.CreatePuppy(&Puppy{102, "German Shepard", "Brown", 3499.90})
+	suite.True(status)
+}
+
+func (suite *StorerTest) TestReadPuppy() {
+	puppy, status := suite.store.ReadPuppy(101)
```
camscale: Similarly, `exists` makes more sense than `status`.

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import (
+	"sync"
+)
+
+// Data structure for puppy
```
camscale: This is not the proper format for a doc string. See https://blog.golang.org/godoc-documenting-go-code for details.

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
...
+type Puppy struct {
+	ID     uint64
+	Breed  string
+	Colour string
+	Value  float64
+}
+
+// Storer defines standard CRUD operations for Puppy
```
camscale: This is _not quite_ the right format for a doc string. It needs a full stop at the end to make it a complete sentence.

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
...
+type Storer interface {
+	CreatePuppy(puppy *Puppy) bool
+	ReadPuppy(id uint64) (*Puppy, bool)
+	UpdatePuppy(id uint64, puppy *Puppy) bool
+	DeletePuppy(id uint64) bool
+}
+
+// MapStore type and implementation
```
camscale: This is also not the correct format for a doc string. Furthermore it is wrong because there is no implementation here.

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
...
+type MapStore map[uint64]*Puppy
+
+// SyncStore struct and implementation
```
camscale: This is also not the correct format for a doc string. Furthermore it is wrong because there is no implementation here.

06_puppy/kasunfdo/types.go
camscale: @lexual is correct that with this implementation, you can alter what is stored without going through the storer interface, but that is not a given due to this type. The implementation could make a copy and still store pointers, and also return a copy. This may make sense because a typical implementation of a "storer" would be to store in an external database, in which case you could not alter the underlying storage through the pointer. But that more describes the interface than this `MapStore` implementation type.

I think it the most sensible approach is to use pointers in the interface and store values in the map.

@alfredxiao There are no requirements on this. It is an implementation detail. But as usual, any assumptions made should be documented.

06_puppy/kasunfdo/syncstore.go
```diff
@@ -0,0 +1,37 @@
+package main
+
+func (s *SyncStore) CreatePuppy(puppy *Puppy) bool {
+	_, exists := s.Map.LoadOrStore(puppy.ID, puppy)
+	// exists is true if a record exists with given key, false if value stored.
+	return !exists
+}
+
+func (s *SyncStore) ReadPuppy(id uint64) (*Puppy, bool) {
+	value, exists := s.Map.Load(id)
```
camscale: This should be folded into the `if` statement as none of the variables are needed outside the body of the if:

```
if value, exists := s.Map.Load(id); exists {
        return value.(*Puppy), true
}
return nil, false
```

06_puppy/kasunfdo/syncstore.go
```diff
@@ -0,0 +1,37 @@
...
+		puppy := value.(*Puppy)
+		return puppy, true
+	}
+
+	return nil, false
+}
+
+func (s *SyncStore) UpdatePuppy(id uint64, puppy *Puppy) bool {
+	if _, exists := s.Map.Load(id); exists {
+		puppy.ID = id
```
camscale: Same comment as on the `MapStore` implementation. Are you updating the ID correctly with respect to the map key?

camscale (dismissed):
This is quite good Go code. There are a couple of places to fix things up (doc strings, more idiomatic code), but otherwise well structured. The test suite is also well done.

I've raised a few points about the implementation of the stores, with respect to updating values. It is also worth considering whether the storer implementations should be allocating the IDs instead of leaving it up to the caller, as the caller really cannot know what a valid ID is. This is more of a design issue that more advanced developers should be thinking about, and I think you are one of those developers. So I would like to see some thought in this area. It is quite ok to just add some comments saying you have decided to leave the ID allocation up to the caller - I just want to see some evidence that you have thought about the issue.

I am not requesting any changes in this design area, but do consider it for the later labs which follow the same patterns.

06_puppy/kasunfdo/main.go
```diff
@@ -0,0 +1,18 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	store := MapStore{}
+	puppy := Puppy{ID: 1, Breed: "Husky", Colour: "White", Value: 4999.98}
+
+	if store.CreatePuppy(&puppy) {
+		fmt.Fprintln(out, "Puppy added to store")
+	}
+}
```
nicko-lee, line 17: Was wondering does this lab even need a main function? Many people seem to have a main function but it doesn't seem necessary based on the lab instructions

juliaogris: @nicko-lee in the [requirements](https://github.com/anz-bank/go-course/blob/master/06_puppy/README.md) it says: 
> Create an executable go program in directory 06_puppy/USERNAME (see hints)

So the answer is yes :)

06_puppy/kasunfdo/main_test.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("Puppy(1) added to store\n")
+	actual := strconv.Quote(buf.String())
```
juliaogris: Drop `Quote` on both.

06_puppy/kasunfdo/main_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+	if expected != actual {
+		t.Errorf("Unexpected output\nexpected: %v\nactual: %v", expected, actual)
+	}
```
juliaogris: Use testify/assert here too.

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,34 @@
...
+	m.store[puppy.ID] = puppy
+	return puppy.ID
+}
+
+func (m *MapStore) ReadPuppy(id uint64) (Puppy, bool) {
+	puppy, exists := m.store[id]
+	return puppy, exists
+}
+
+func (m *MapStore) UpdatePuppy(id uint64, puppy Puppy) bool {
```
juliaogris: Better return an `error` - doesn't need to be a custom error type (that's part of the next lab I believe), but at least:
```go
fmt.Errorf("no puppy with ID %d", id)
```
To me it's nicer to only keep the ID encoded in `puppy`. It feels a bit dodgy to accept a puppy whose ID doesn't match the separate id field. You could add an error check for it, but really just use only one ID imo.  
Feel free to argue against it :)

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,34 @@
...
+	if _, exists := m.store[id]; exists {
+		puppy.ID = id
+		m.store[id] = puppy
+		return true
+	}
+	return false
+}
+
+func (m *MapStore) DeletePuppy(id uint64) bool {
```
juliaogris: error if not found IMO

06_puppy/kasunfdo/store_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	isSuccessful = suite.store.DeletePuppy(suite.id)
+	suite.False(isSuccessful)
+}
+
+func TestMapStore(t *testing.T) {
+	s := StorerTest{
+		store: NewMapStore(),
+	}
+	suite.Run(t, &s)
```
juliaogris: IMO you should just fold this into one line (fee free to argue against it):
```go
suite.Run(t, &StorerTest{store: NewMapStore()})
```
same below.

06_puppy/kasunfdo/types.go
```diff
@@ -0,0 +1,29 @@
...
+
+type MapStore struct {
+	store map[uint64]Puppy
+	count uint64
+}
+
+type SyncStore struct {
+	store sync.Map
+	count uint64
+}
```
juliaogris, line 17: IMO SyncStore & MapStore should sit with their implementation files (`syncstore.go`, `mapstore.go`).
Also, consider renaming `count` to `nextID` (that's how it seems to be used).

juliaogris (changes requested):
A few nitpicks, overall this really nice code, very enjoyable to read IMO.

If you don't mind, could you please add [godoc comments](https://blog.golang.org/godoc-documenting-go-code) to your public types and functions? (the linter should pick this up, not quite sure why it doesn't).

06_puppy/kasunfdo/mapstore.go
kasunfdo: I did not use error as it has to be implemented in lab 7 😃
Anyway, I'll return error.

06_puppy/kasunfdo/types.go
kasunfdo: Agree. :)

kasunfdo: Agree

06_puppy/kasunfdo/store_test.go
kasunfdo: Done

06_puppy/kasunfdo/mapstore.go
kasunfdo: Done

06_puppy/kasunfdo/main_test.go
kasunfdo: Done

06_puppy/kasunfdo/main_test.go
```diff
@@ -0,0 +1,20 @@
...
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := "Puppy(1) added to store\n"
+	actual := buf.String()
+
+	assert.Equalf(t, expected, actual, "Unexpected output")
```
juliaogris: No need for this failure message, testify has you covered unless you print something specific with context that makes sense to you.

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// NewMapStore creates a new MapStore
+func NewMapStore() *MapStore {
+	return &MapStore{nextID: 0, store: map[uint64]Puppy{}}
+}
+
+// CreatePuppy inserts given puppy in store and returns given id to the puppy
+func (m *MapStore) CreatePuppy(puppy Puppy) uint64 {
+	m.nextID++
```
juliaogris: Semantics, but imo nextID should hold the next I'd.
So sign it and increment after assignment.

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,52 @@
...
+	puppy.ID = m.nextID
+	m.store[puppy.ID] = puppy
+	return puppy.ID
+}
+
+// ReadPuppy reads puppy with given id from the store
+func (m *MapStore) ReadPuppy(id uint64) (Puppy, error) {
+	if puppy, exists := m.store[id]; exists {
+		return puppy, nil
```
juliaogris: Very nit picky, but imo logic should be inverted: if !exist ...
Is bail early on error. below too

juliaogris (dismissed):
Please fix commit messages (imperative mood , no fixup commits - rework history instead)

A few minor nitpicky comments inline. Rest looks perfect, thank you.

06_puppy/kasunfdo/mapstore.go
kasunfdo: I did in this way because otherwise I'll have to use `else` to retain `puppy` value or I'll have to make `puppy, exists := m.store[id]` a statement.

I'll change as requested.

06_puppy/kasunfdo/mapstore.go
```diff
@@ -0,0 +1,56 @@
...
+func (m *MapStore) CreatePuppy(puppy Puppy) uint64 {
+	puppy.ID = m.nextID
+	m.nextID++
+	m.store[puppy.ID] = puppy
+	return puppy.ID
+}
+
+// ReadPuppy reads puppy with given id from the store
+func (m *MapStore) ReadPuppy(id uint64) (Puppy, error) {
+	puppy, exists := m.store[id]
```
rokane, line 28: If you flip the logic you can simply this further to fit on 3 lines.
```go
func (m *MapStore) ReadPuppy(id uint64) (Puppy, error) {
  if puppy, exists := m.store[id]; exists {
    return puppy, nil
  }
  return Puppy{}, fmt.Errorf("no puppy with id: %d", id)
}
```

06_puppy/kasunfdo/store_test.go
```diff
@@ -0,0 +1,71 @@
...
+	_, err = suite.store.ReadPuppy(suite.id)
+	suite.Equal(fmt.Sprintf("no puppy with id: %d", suite.id), err.Error())
+
+	err = suite.store.DeletePuppy(suite.id)
+	suite.Equal(fmt.Sprintf("no puppy with id: %d", suite.id), err.Error())
+}
+
+func TestMapStore(t *testing.T) {
+	suite.Run(t, &StorerTest{store: NewMapStore()})
+}
```
rokane, line 67: Up to you and how you want your output formatted , but you can have these suites trigger in the same test. 

```go
func TestStore(t *testing.T) {
  suite.Run(t, &StorerTest{store: NewMapStore()})
  suite.Run(t, &StorerTest{store: NewSyncStore()})
}
```

rokane (approved):
Looks good to me. Really nice code :) I left a couple of minor suggestions, none of which are major enough to block this from being merged so I will leave it up to you as to whether you include them or not.

06_puppy/kasunfdo/mapstore.go
kasunfdo: That's how I had done it. But changed as per Julia's comment. 😃
https://github.com/anz-bank/go-course/pull/342#discussion_r314966815

06_puppy/kasunfdo/store_test.go
kasunfdo: Thank you! I will do so in Lab 7 😊

PR #344, opened 2019-06-12
Lab2 Bubble sort

zhukaisjtu:
Fixes #

Review of colleague's PR #

Implemented the bubble sort function.

### DIFF
--- a/02_bubble/kaizhu/main.go
+++ b/02_bubble/kaizhu/main.go
@@ -1,0 +1,31 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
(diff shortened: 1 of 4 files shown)


02_bubble/kaizhu/main.go
```diff
@@ -0,0 +1,31 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(sc []int) []int {
+	s := make([]int, len(sc))
+	copy(s, sc)
+	for i := 0; i < len(s)-1; i++ {
```
hsy3418, line 18: The len(s) could be called outside the for loop by assigning the value to a variable, otherwise, this function will be called every loop.

PR #349, opened 2019-06-12
Implement Puppy - Lab 6

### LINKED ISSUE #348: Lab 6 - CRUD puppy with interface
Create an executable go program in directory 06_puppy/brvishnu92
Implement a Puppy struct containing ID, Breed, Colour, Value.
Create Storer interface with CRUD methods for Puppy
Write a MapStore implementation of Storer backed by a map
Write a SyncStore implementation of Storer backed by a sync.Map

brvishnu92:
Fixes #348

Review of colleague's PR #235

 - Implement Puppy with test

### DIFF
--- a/06_puppy/brvishnu92/main.go
+++ b/06_puppy/brvishnu92/main.go
@@ -1,0 +1,104 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+var out io.Writer = os.Stdout
+
+type Puppy struct {
+	id    uint32
+	breed string
+	color string
+	value string
+}
+
+type Storer interface {
+	CreatePuppy(*Puppy)
+	ReadPuppy(id uint32) *Puppy
+	UpdatePuppy(puppy *Puppy)
+	DeletePuppy(id uint32) bool
+	HasPuppy(id uint32) bool
+}
+
+type MapStore map[uint32]Puppy
+
+type SynStore struct {
+	store *sync.Map
+}
+
+func (ms MapStore) CreatePuppy(pup *Puppy) {
+	if !ms.HasPuppy(pup.id) {
+		ms[pup.id] = *pup
+	}
+}
+func (ms MapStore) ReadPuppy(id uint32) *Puppy {
+	if puppy, has := ms[id]; has {
+		return &puppy
+	}
+	return nil
+}
+
+func (ms MapStore) UpdatePuppy(puppy *Puppy) {
+	if ms.HasPuppy(puppy.id) {
+		ms[puppy.id] = *puppy
+	}
+}
+
+func (ms MapStore) DeletePuppy(id uint32) bool {
+	if ms.HasPuppy(id) {
+		delete(ms, id)
+		return true
+	}
+	return false
+}
+func (ms MapStore) HasPuppy(id uint32) bool {
+	_, has := ms[id]
+	return has
+}
+
+func (ss SynStore) CreatePuppy(pup *Puppy) {
+	if !ss.HasPuppy(pup.id) {
+		ss.store.Store(pup.id, pup)
+	}
+}
+
+func (ss SynStore) ReadPuppy(id uint32) *Puppy {
+	if pup, ok := ss.store.Load(id); ok {
+		return pup.(*Puppy)
+	}
+	return nil
+}
+
+func (ss SynStore) UpdatePuppy(puppy *Puppy) {
+	if ss.HasPuppy(puppy.id) {
+		ss.store.Store(puppy.id, puppy)
+	}
+}
+
+func (ss SynStore) DeletePuppy(id uint32) bool {
+	if ss.HasPuppy(id) {
+		ss.store.Delete(id)
+		return true
+	}
+	return false
+}
+
+func (ss SynStore) HasPuppy(id uint32) bool {
+	_, ok := ss.store.Load(id)
+	return ok
+}
+
+func main() {
+	tempPuppy := Puppy{id: 1, breed: "dog", color: "Black", value: "1"}
+	var store Storer = make(MapStore)
+	store.CreatePuppy(&tempPuppy)
+	fmt.Fprintln(out, *store.ReadPuppy(tempPuppy.id))
+	tempPuppy2 := Puppy{id: 1, breed: "cat", color: "Brown", value: "1"}
+	var store2 Storer = SynStore{store: &sync.Map{}}
+	store2.CreatePuppy(&tempPuppy2)
+	fmt.Fprintln(out, *store2.ReadPuppy(tempPuppy.id))
+}
--- a/06_puppy/brvishnu92/main_test.go
+++ b/06_puppy/brvishnu92/main_test.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"bytes"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type StorerTestSuite struct {
+	suite.Suite
+	storer Storer
+}
+
+func (suite *StorerTestSuite) SetupTest() {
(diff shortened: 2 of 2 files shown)


06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,95 @@
...
+func (ms MapStore) DeletePuppy(id uint32) bool {
+	if ms[id] != nil {
+		delete(ms, id)
+		return true
+	}
+	return false
+}
+
+func (ss SynStore) CreatePuppy(pup Puppy) {
+
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)

06_puppy/brvishnu92/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+func (suite *StorerTestSuite) TestCreate() {
+	crePup := Puppy{id: 1, breed: "cat", color: "Black", value: "val1"}
+	suite.storer.CreatePuppy(crePup)
+	returnPup := suite.storer.ReadPuppy(crePup.id)
+	suite.Equal(returnPup.id, crePup.id)
+	suite.Equal(returnPup.breed, crePup.breed)
+}
+
+func (suite *StorerTestSuite) TestRead() {
+	ExistingPup := *suite.storer.ReadPuppy(1)
```
CameronSchafer: is this supposed to be a global variable?

CameronSchafer (commented):
Just a little nitpick otherwise all good

06_puppy/brvishnu92/main_test.go
brvishnu92: Done. Thanks 👍

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,92 @@
...
+	if ms[pup.id] == nil {
+		ms[pup.id] = &pup
+	}
+}
+func (ms MapStore) ReadPuppy(id uint32) *Puppy {
+	return ms[id]
+}
+
+func (ms MapStore) UpdatePuppy(id uint32, puppy *Puppy) {
+	if ms[id] != nil {
```
anzdaddy: Why not `puppy.id`?

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,92 @@
...
+func (ms MapStore) CreatePuppy(pup Puppy) {
+	if ms[pup.id] == nil {
+		ms[pup.id] = &pup
+	}
+}
+func (ms MapStore) ReadPuppy(id uint32) *Puppy {
+	return ms[id]
+}
+
+func (ms MapStore) UpdatePuppy(id uint32, puppy *Puppy) {
```
anzdaddy: Not strictly required, since you can modify the result of `ReadPuppy`, but that is a problem in its own right.

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+type Storer interface {
+	CreatePuppy(Puppy)
+	ReadPuppy(id uint32) *Puppy
+	UpdatePuppy(id uint32, puppy *Puppy)
+	DeletePuppy(id uint32) bool
+}
+
+type MapStore map[uint32]*Puppy
```
anzdaddy: `Puppy` is more robust than `*Puppy`.

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,92 @@
...
+func (ms MapStore) DeletePuppy(id uint32) bool {
+	if ms[id] != nil {
+		delete(ms, id)
+		return true
+	}
+	return false
+}
+
+func (ss SynStore) CreatePuppy(pup Puppy) {
+	if _, ok := ss.store.Load(pup.id); !ok {
```
anzdaddy: This is common code. I suggest adding a `hasPuppy(id uint32)` (or `HasPuppy`) method and using it everywhere.

anzdaddy (changes requested):

06_puppy/brvishnu92/main.go
brvishnu92: I cant do the validations if the mapstore type is changed to puppy, for eg `if ms[pup.id] == nil` . I would have to validate against a puppy whose value I wouldnt know.

anzdaddy:
```go
if _, has := ms[pup.id]; has {
```

06_puppy/brvishnu92/main.go
brvishnu92: Done. :-)

06_puppy/brvishnu92/main.go
brvishnu92: Oh yeah, thanks for that :-D Done 👍

06_puppy/brvishnu92/main.go
brvishnu92: Done.

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,104 @@
...
+	color string
+	value string
+}
+
+type Storer interface {
+	CreatePuppy(*Puppy)
+	ReadPuppy(id uint32) *Puppy
+	UpdatePuppy(puppy *Puppy)
+	DeletePuppy(id uint32) bool
+	HasPuppy(id uint32) bool
```
juliaogris, line 24: All of these functions should also return `error`

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,104 @@
...
+	store *sync.Map
+}
+
+func (ms MapStore) CreatePuppy(pup *Puppy) {
+	if !ms.HasPuppy(pup.id) {
+		ms[pup.id] = *pup
+	}
+}
+func (ms MapStore) ReadPuppy(id uint32) *Puppy {
+	if puppy, has := ms[id]; has {
```
juliaogris, line 39: use `ok` instead of `has` ([comma go idiom](https://golang.org/doc/effective_go.html#maps))

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,104 @@
...
+		ms[puppy.id] = *puppy
+	}
+}
+
+func (ms MapStore) DeletePuppy(id uint32) bool {
+	if ms.HasPuppy(id) {
+		delete(ms, id)
+		return true
+	}
+	return false
```
juliaogris, line 56: return error instead

06_puppy/brvishnu92/main.go
```diff
@@ -0,0 +1,104 @@
...
+}
+func (ms MapStore) HasPuppy(id uint32) bool {
+	_, has := ms[id]
+	return has
+}
+
+func (ss SynStore) CreatePuppy(pup *Puppy) {
+	if !ss.HasPuppy(pup.id) {
```
juliaogris, line 64: The point of using sync.Map is to protect your code from race conditions (make it "thread/goroutine safe").
In your case HasPuppy reads before Store updates. Another go routine could in the meantime alter the result of HasPuppy.
To mitigate this you can use [Mutex](https://golang.org/pkg/sync/#Mutex) or [RWMutex](https://golang.org/pkg/sync/#RWMutex) - use lock at beginning of method and unlock at end where instructions need to be executed together.
Same for all other methods that use two or more access to the underlying syncmap

juliaogris (changes requested):
I've fixed the Markdown formatting in your PR description.
Please add some basic error handling (no need for custom errors) - this will make lab 7 easier too.
Please also add mutexes in the syncmap implementation to mitigate race conditions.
IMO it's also time to split up your code into several files, I'd suggest something like: main.go, storer.go | types.go, mapstore.go, syncmapstore.go, ...

PR #347, opened 2019-06-12
Lab 3 - Letter Frequency

marksomething:
Fixes #346

Review of colleague's PR #333

#### Changes proposed in this PR:
* letters(string) map[rune]int
  returns a map of the frequency of runes
* sortLetters(map[rune]int) [] string
  returns a slice of rune:frequency pairs formatted as string
  and ordered by descending frequency
* main()
  prints an ordered list of rune:freuency pairs

PR #351, opened 2019-06-12
Lab3 Letter Frequency

### LINKED ISSUE #350: Lab3 Letter Frequency

pfannerj:
Fixes #350

Review of colleague's PR #347

- build letter frequency solution
- build letter frequency test cases


03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+	input    string
+	expected map[rune]int
+}{
+	"Standard": {input: "aba", expected: map[rune]int{97: 2, 98: 1}},
+	"Same":     {input: "cccccccccccccccccccc", expected: map[rune]int{99: 20}},
+	"Empty":    {input: "", expected: map[rune]int{}},
+	"Return":   {input: "ab\na\n", expected: map[rune]int{10: 2, 97: 2, 98: 1}},
+	"Slash":    {input: "ab/na/n", expected: map[rune]int{47: 2, 97: 2, 98: 1, 110: 2}},
+	"LongMixed": {input: "zabaaarzzzASYbcd@&^!*JHDJKHHGFGD^!*Dar",
+		expected: map[rune]int{33: 2, 38: 1, 42: 2, 64: 1, 65: 1, 68: 3, 70: 1, 71: 2, 72: 3, 74: 2, 75: 1, 83: 1,
```
runnerdave: not too fussed about but this lines are very long, I tend to stick to 80, but it is a very minor issue, the only I can see, albeit very minor

runnerdave (dismissed):
Very simple and clear, just added a minor issue but please disregard if you don't consider it important.

03_letters/pfannerj/main_test.go
pfannerj: Thanks!  Yeah, I got an error from the lint and split the lines so that it was happy but there still could be a better way to format the long lines.. will investigate but leaving it as-is for now...

03_letters/pfannerj/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+}
+
+func letters(s string) map[rune]int {
+
+	mapOfRunes := map[rune]int{}
+	allRunes := []rune(s)
+	count := 0
+	for i := range allRunes {
+		count = mapOfRunes[allRunes[i]]
```
n0npax: I believe you can improve this with sth like:
`mapOfRunes[allRunes[i]]++`

03_letters/pfannerj/main.go
```diff
@@ -0,0 +1,42 @@
...
+		mapOfRunes[allRunes[i]] = count + 1
+	}
+	return mapOfRunes
+}
+
+func sortLetters(m map[rune]int) []string {
+
+	lettersCount := make([]string, len(m))
+	i := 0
+	for thisRune := range m {
```
n0npax: please consider sth like `for i, thisRune := range m {`

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+		})
+	}
+}
+
+var sortLettersCases = map[string]struct {
+	input    map[rune]int
+	expected []string
+}{
+	"Standard": {input: map[rune]int{98: 1, 97: 2}, expected: []string{"a:2", "b:1"}},
+	"Long": {input: map[rune]int{72: 3, 74: 2, 75: 1, 83: 1, 89: 1, 94: 2, 97: 5, 98: 2, 99: 1, 100: 1,
```
n0npax: using `rune` instead of `int` will improve readability in my mind.
please consider difference between:
`map[rune]int{'о': 3, 'р': 2, 'т': 1, 'у': 1, ' ': 1, 'е': 1, 'б': 1, 'д': 1}}`
and
`map[rune]int{72: 3, 74: 2, 75: 1, 83: 1, 89: 1, 94: 2, 97: 5, 98: 2, 99: 1, 100: 1,...`

n0npax (changes requested):

03_letters/pfannerj/main.go
pfannerj: I tried out that but it doesn't get me what I need... if I say for thisRune, i := range m { then i will be the count and not the index of the slice.  I've defined and incremented i separately so that I can use it as an index to the lettersCount slice and insert the strings before sorting lettersCount... will leave this one as is...

n0npax: awww yes, m is map. sorry for this.

03_letters/pfannerj/main.go
pfannerj: Yes, thx!  I have done that change and I also further simplified because I realised after reviewing someone else's solution that I don't need to create the allRunes slice at all.. I just have this now...
`	for _, thisRune := range s {
		mapOfRunes[thisRune]++
	}`

03_letters/pfannerj/main_test.go
pfannerj: Thx again... I agree it is better and have updated all my test cases.  I also stole your test characters and added it as an extra test case :)

n0npax (dismissed):
lgtm

03_letters/pfannerj/main.go
```diff
@@ -0,0 +1,39 @@
...
+		lettersCount[i] = string(thisRune) + ":" + strconv.Itoa(m[thisRune])
```
rokane: Nice can also do a direct string format here to achieve in another way. No need to update just comes down to personal preference.
```go
lettersCount[i] = fmt.Sprintf("%v:%v", string(thisRune), string(m[thisRune]))
```

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+	b:1`, expected: map[rune]int{9: 1, 10: 1, 49: 1, 50: 1, 58: 2, 97: 1, 98: 1}},
+	"Numeric":      {input: "11235", expected: map[rune]int{49: 2, 50: 1, 51: 1, 53: 1}},
+	"EscapedQuote": {input: "a\":2\"b:\"1", expected: map[rune]int{34: 3, 49: 1, 50: 1, 58: 2, 97: 1, 98: 1}},
+	"RawQuote":     {input: `a":2"b:"1`, expected: map[rune]int{34: 3, 49: 1, 50: 1, 58: 2, 97: 1, 98: 1}},
+}
+
+func TestLetters(t *testing.T) {
+	for name, tc := range lettersCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
```
rokane, line 48: Really thorough test cases. Also not a requirement of this lab but if you look at using the `testify/assert` package you can simplify the running / assertion of each test case quite a lot. Don't worry about changing, but try to incorporate into your next lab :)

```go
for name, test := range lettersCases {
   input := test.input
   expected := test.expected
      assert.Equalf(t, letters(input), expected, fmt.Sprintf("Expected %v, but got %v", expected, letters(input)))
   })
}
```

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+	if expected != actual {
+		t.Errorf(expected, actual, "Unexpected output in main()")
+	}
+}
+
+var lettersCases = map[string]struct {
+	input    string
+	expected map[rune]int
+}{
+	"Standard": {input: "aba", expected: map[rune]int{97: 2, 98: 1}},
```
rokane: I think it makes it a lot easier to read if you use the actual char representing the rune, rather than the unicode code point. Can you please update this to reflect?

Example:
```go
"Standard": { input: "aba", "expected": map[rune]int{'a': 2, 'b':1 }
```

rokane (dismissed):
Good work on this. If you can update the one thing regarding the expected Rune maps in the test cases that will be great :)

03_letters/pfannerj/main_test.go
pfannerj: Yeah, I used that for my Lab5... thx again..

03_letters/pfannerj/main_test.go
pfannerj: Will do... I did that already for the SortLetters cases after feedback but I didn't think of doing for the Letters function.

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+}{
+	"Standard": {input: "aba", expected: map[rune]int{'a': 2, 'b': 1}},
+	"Same":     {input: "cccccccccccccccccccc", expected: map[rune]int{'c': 20}},
+	"Empty":    {input: "", expected: map[rune]int{}},
+	"Return":   {input: "ab\na\n", expected: map[rune]int{'\n': 2, 'a': 2, 'b': 1}},
+	"Slash":    {input: "ab/na/n", expected: map[rune]int{'/': 2, 'a': 2, 'b': 1, 'n': 2}},
+	"LongMixed": {input: "zabaaarzzzASYbcd@&^!*JHDJKHHGFGD^!*Dar",
+		expected: map[rune]int{'!': 2, '&': 1, '*': 2, '@': 1, 'A': 1, 'D': 3, 'F': 1, 'G': 2, 'H': 3, 'J': 2, 'K': 1, 'S': 1,
+			'Y': 1, '^': 2, 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2, 'z': 4}},
+	"Punctuation": {input: ".:,!;'?.,:!'", expected: map[rune]int{'!': 2, '\'': 2, ',': 2, '.': 2, ':': 2, ';': 1, '?': 1}},
```
golangcibot: line is 121 characters (from `lll`)

01_fib/lexual/main_test.go
```diff
@@ -62,3 +62,27 @@ func TestMain(t *testing.T) {
 		t.Errorf("\nActual: %q\nExpected: %q", actual, expected)
 	}
 }
+
+func TestMain(t *testing.T) {
```
golangcibot: `TestMain` redeclared in this block (from `typecheck`)

01_fib/lexual/main_test.go
```diff
@@ -0,0 +1,112 @@
...
+		actual := buf.String()
+
+		if test.expected != actual {
+			// %q (raw-string), we don't want to print '\n' as newline.
+			t.Errorf("\nActual: %q\nExpected: %q", actual, test.expected)
+		}
+	}
+}
+
+func TestMain(t *testing.T) {
```
golangcibot: other declaration of TestMain (from `typecheck`)

davidbroughsmyth (commented):
Looks like the review has gone stale. I've noted the changes you made and comments. Yeah, the using the testify/assert is a preference thing. Nice coverage on tests. LGTM.

03_letters/pfannerj/main.go
```diff
@@ -0,0 +1,39 @@
...
+
+func main() {
+
+	fmt.Fprintf(mainout, strings.Join(sortLetters(letters("aba")), "\n"))
+
+}
+
+func letters(s string) map[rune]int {
+
+	mapOfRunes := map[rune]int{}
```
camscale: Using `map` in a variable name (or any type) is discouraged in Go: https://dave.cheney.net/2019/01/29/you-shouldnt-name-your-variables-after-their-types-for-the-same-reason-you-wouldnt-name-your-pets-dog-or-cat

03_letters/pfannerj/main.go
```diff
@@ -0,0 +1,39 @@
...
+	for _, thisRune := range s {
```
camscale: Naming is a bit of a contentious point with Go, but they do encourage short names for short scopes, so `thisRune` would be best named something like `r`, given that it is used over only two lines.

To see some early thoughts from Russ Cox (a Go luminary): https://research.swtch.com/names, and a slide deck from Andrew Gerrand (another Go luminary): https://talks.golang.org/2014/names.slide

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,79 @@
...
+			'Y': 1, '^': 2, 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2, 'z': 4}},
+	"Punctuation": {input: ".:,!;'?.,:!'",
+		expected: map[rune]int{'!': 2, '\'': 2, ',': 2, '.': 2, ':': 2, ';': 1, '?': 1}},
+	"Raw": {input: `a:2b:1`, expected: map[rune]int{'1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"RawReturn": {input: `a:2
+	b:1`, expected: map[rune]int{'	': 1, '\n': 1, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"Numeric":      {input: "11235", expected: map[rune]int{'1': 2, '2': 1, '3': 1, '5': 1}},
+	"EscapedQuote": {input: "a\":2\"b:\"1", expected: map[rune]int{'"': 3, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"RawQuote":     {input: `a":2"b:"1`, expected: map[rune]int{'"': 3, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+}
```
camscale, line 43: I know Ryan has already said "Really thorough test cases", but I'm going to counter that and say I think there are too many test cases here. Most of them are not testing anything different to the other test cases. For instance, `EscapedQuote` and `RawQuote` are the same test case - the different quoting is a property of the compiler, not a property of your code, so it makes little sense to have both tests. Ideally you have the smallest set of test cases to test your code.

Your code make no distinction between all the different types of characters you have in your tests, so you do not need all the tests that make that distinction. Every test is a test that needs to be validated and maintained so having too many tests can be a burden.

What is missing though is some unicode characters, where the value of the rune does not fit into a byte. When you index a string, you get a byte. When you range over it, you get runes (multi-byte characters). So when dealing with carving a string up into runes, it is worthwhile to have tests containing multi-byte characters to ensure you have not incorrectly indexed a string to get a byte, rather than correctly extracting a rune.

I would like to see some multi-byte character runes in the test cases before approving - I've seen one lab do this incorrectly (although I think it was numeronyms). I'll leave it up to you as to whether you want to reduce the size of the test cases or just take it into consideration for future labs.

The gory details about strings in Go, if you're interested: https://blog.golang.org/strings

camscale (changes requested):
I'm really sorry for the delay in reviewing this. I hope we can do better from here out.

There are a couple of changes I'd like to see - not using "types" in variable names, as that is quite discouraged in Go, and adding the multi-byte test case I mentioned. The other comment about short names can be left as is.

Another general comment is that some of the code is a bit more widely spaced vertically than you'd usually see for Go code (blank lines at the start of functions, and the extra one in `main()`). That and the short name comment both fall into a "personal style" category, so I'll leave it up to you because my usual rule is "my project, my style". But be aware that your style may be contrary to more common Go code you'll see and if you're working on a shared code base, you may need to adapt your style to be closer to the Go idioms. Andrew Gerrand has a nice slide deck from a presentation on naming here: https://talks.golang.org/2014/names.slide

Ping me on slack (camh) when you're ready for a re-review and I'll get right on it. I'll also keep an eye on the PR too. Again, sorry for the delay.

03_letters/pfannerj/main.go
pfannerj: Well actually, my friend had a cat called 'Kitten' :)  but that's cool... I will change it and check for others like that in my newer lab solutions.

03_letters/pfannerj/main.go
pfannerj: Yeah, I've seen some discussion like this pop up on our slack channel but that wasn't til after I submitted this lab...

03_letters/pfannerj/main_test.go
pfannerj: OK, I was a bit surprised to see that I don't have any multi-byte tests because I really thought I had included some. 
 I can see now that I did include them for the sortLetters tests but not for the other tests..

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,80 @@
...
+			'Y': 1, '^': 2, 'a': 5, 'b': 2, 'c': 1, 'd': 1, 'r': 2, 'z': 4}},
+	"Punctuation": {input: ".:,!;'?.,:!'",
+		expected: map[rune]int{'!': 2, '\'': 2, ',': 2, '.': 2, ':': 2, ';': 1, '?': 1}},
+	"Raw": {input: `a:2b:1`, expected: map[rune]int{'1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"RawReturn": {input: `a:2
+	b:1`, expected: map[rune]int{'	': 1, '\n': 1, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"Numeric":      {input: "11235", expected: map[rune]int{'1': 2, '2': 1, '3': 1, '5': 1}},
+	"EscapedQuote": {input: "a\":2\"b:\"1", expected: map[rune]int{'"': 3, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"RawQuote":     {input: `a":2"b:"1`, expected: map[rune]int{'"': 3, '1': 1, '2': 1, ':': 2, 'a': 1, 'b': 1}},
+	"MultiByteMix": {input: "日本語A⌘дтB語A⌘д", expected: map[rune]int{'A': 2, 'B': 1, 'д': 2, 'т': 1, '⌘': 2, '日': 1, '本': 1, '語': 2}},
```
golangcibot: line is 129 characters (from `lll`)

camscale (approved):
LGTM

PR #352, opened 2019-06-12
Letter frequency

runnerdave:
Fixes #[63](https://github.com/anz-bank/go-course/issues/63) 

Review of colleague's PR [#351](https://github.com/anz-bank/go-course/pull/351#pullrequestreview-248590162) 

- Implement test cases for sorting and frequency functions
- Write functions to pass tests

### DIFF
--- a/03_letters/runnerdave/main.go
+++ b/03_letters/runnerdave/main.go
@@ -5,41 +5,35 @@ import (
 	"io"
 	"os"
 	"sort"
+	"strconv"
+	"strings"
 )
 
 var out io.Writer = os.Stdout
 
 func letters(s string) map[rune]int {
 	stats := make(map[rune]int)
-	for pos, char := range s {
-		stats[char] = pos
+	for _, char := range s {
+		stats[char]++
 	}
 	return stats
 }
 
 func sortLetters(m map[rune]int) []string {
-	var keys []rune
+	keys := make([]rune, 0)
 	for k := range m {
 		keys = append(keys, k)
 	}
 	sorted(keys)
 	sortedArray := make([]string, len(keys))
-	i := 0
-	for _, k := range keys {
-		sortedArray[i] = string(k)
-		i++
+	for i, k := range keys {
+		sortedArray[i] = fmt.Sprintf("%s:%s", string(k), strconv.Itoa(m[k]))
 	}
 	return sortedArray
 }
 
 func main() {
-	printRuneMap(letters("aba"))
-}
-
-func printRuneMap(m map[rune]int) {
-	for key, value := range m {
-		fmt.Fprintf(out, "%s:%d\n", string(key), value)
-	}
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
 }
 
 type runeSlice []rune
--- a/03_letters/runnerdave/main.go
+++ b/03_letters/runnerdave/main.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	stats := make(map[rune]int)
+	for pos, char := range s {
+		stats[char] = pos
+	}
+	return stats
+}
+
+func sortLetters(m map[rune]int) []string {
+	var keys []rune
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sorted(keys)
+	sortedArray := make([]string, len(keys))
+	i := 0
+	for _, k := range keys {
+		sortedArray[i] = string(k)
+		i++
+	}
+	return sortedArray
+}
+
+func main() {
+	printRuneMap(letters("aba"))
+}
+
+func printRuneMap(m map[rune]int) {
+	for key, value := range m {
+		fmt.Fprintf(out, "%s:%d\n", string(key), value)
+	}
+}
+
+type runeSlice []rune
+
+func (p runeSlice) Len() int           { return len(p) }
+func (p runeSlice) Less(i, j int) bool { return p[i] < p[j] }
+func (p runeSlice) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+
+func sorted(runes []rune) string {
+	sort.Sort(runeSlice(runes))
+	return string(runes)
+}
--- a/03_letters/runnerdave/main_test.go
+++ b/03_letters/runnerdave/main_test.go
@@ -51,31 +51,19 @@ func TestSortingLetters(t *testing.T) {
 		x map[rune]int
 		n []string
 	}{
-		{map[rune]int{97: 1, 98: 3}, []string{"a:1", "b:3"}},
-		{map[rune]int{99: 1, 98: 3}, []string{"b:3", "c:1"}},
-		{map[rune]int{109: 1, 98: 3, 100: 5}, []string{"b:3", "d:5", "m:1"}},
-		{map[rune]int{97: 1, 65: 3}, []string{"A:3", "a:1"}},
-		{map[rune]int{110: 1, 78: 3}, []string{"N:3", "n:1"}},
-		{map[rune]int{110: 0, 78: 3}, []string{"N:3", "n:0"}},
+		{map[rune]int{'a': 1, 'b': 3}, []string{"a:1", "b:3"}},
+		{map[rune]int{'c': 1, 'b': 3}, []string{"b:3", "c:1"}},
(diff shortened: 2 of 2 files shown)


03_letters/runnerdave/main.go
```diff
@@ -0,0 +1,54 @@
...
+		if freq, ok := stats[char]; ok {
+			stats[char] = freq + 1
+		} else {
+			stats[char] = 1
+		}
+	}
+	return stats
+}
+
+func sortLetters(m map[rune]int) []string {
```
n0npax, line 22: As far as i remember sorting should be done by `values` not by the `key`. You can find confirmation in the slack channel

n0npax (changes requested):

03_letters/runnerdave/main.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	stats := make(map[rune]int)
+	for _, char := range s {
+		if freq, ok := stats[char]; ok {
+			stats[char] = freq + 1
+		} else {
```
n0npax: you can skip `else` statement. `stats[c]` where `c` is not in `stats` evaluates to 0

03_letters/runnerdave/main.go
```diff
@@ -0,0 +1,54 @@
...
+func sortLetters(m map[rune]int) []string {
+	keys := make([]rune, 0)
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sorted(keys)
+	sortedArray := make([]string, len(keys))
+	i := 0
+	for _, k := range keys {
+		sortedArray[i] = string(k) + ":" + strconv.Itoa(m[k])
```
n0npax: In my mind, `fmt.Fprintf` will be better here. I'm not `go` dev, but most of the languages suggest avoiding raw string concatenation.

03_letters/runnerdave/main.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]rune, 0)
+	for k := range m {
+		keys = append(keys, k)
+	}
+	sorted(keys)
+	sortedArray := make([]string, len(keys))
+	i := 0
```
n0npax: keys is an array. In my mind you should move `i` to loop and iterate over `for i, k : range ....`

03_letters/runnerdave/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestLetters(t *testing.T) {
+	tables := []struct {
+		n string
+		x map[rune]int
+	}{
+		{"aaa", map[rune]int{97: 3}},
```
n0npax: using `a` instead of `97` would improve readability

03_letters/runnerdave/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+		{"bbbb", map[rune]int{98: 4}},
+		{"B", map[rune]int{66: 1}},
+		{"bbbB", map[rune]int{98: 3, 66: 1}},
+		{"bbbBCcccc", map[rune]int{98: 3, 66: 1, 67: 1, 99: 4}},
+		{"bbbBCbbbb", map[rune]int{98: 7, 66: 1, 67: 1}},
+		{"", map[rune]int{}},
+	}
+	for _, table := range tables {
+		result := letters(table.n)
+		if !reflect.DeepEqual(result, table.x) {
```
n0npax, line 43: you've got `equal` function. Do you need `reflect.DeepEqual` in this case?

n0npax (changes requested):

03_letters/runnerdave/main_test.go
runnerdave: the equal function is for arrays, the DeepEqual is for maps

03_letters/runnerdave/main.go
runnerdave: I had to do that a lot in the secure code warrior training ;)

03_letters/runnerdave/main.go
runnerdave: there is no specification on what to use to sort, from the **readme**:

> Write a function that returns a sorted slice of strings with elements "{key}:{val}". Use package sort:

The only spec is to use the package sort, I guess to practice using packages, which I did.

03_letters/runnerdave/main_test.go
runnerdave: good idea

03_letters/runnerdave/main.go
runnerdave: I need to initialise the map value for frequency with 1 for the first time, can you please elaborate what do you mean?

03_letters/runnerdave/main.go
runnerdave: spot on

03_letters/runnerdave/main.go
n0npax: 👍

03_letters/runnerdave/main.go
n0npax: I mean `0+1 = 1`
Lets go step by step here:
1: Stats is empty at the beginning, so you're hitting else branch `stats[c] = 1`
2: Stats have value. You're taking this value and incrementing by 1 `freq = stats[c]; stats[c] = freq+ 1`
My suggestion is to do sth like:
`stats[c]++`


----
please check example:
```
func letters(s string) map[rune]int {
        freqMap := map[rune]int{}
        ss := []rune(s)
        for i := range ss {
                freqMap[ss[i]]++
        }
        return freqMap
}
```

03_letters/runnerdave/main_test.go
n0npax: aww yes, you're right, but in this case you're using `DeepEqual` for maps and custom function for slices. In my mind it should be unified. You should use custom functions for both or DeepEqual for both.

runnerdave: I disagree, that would be an overkill, DeepEqual is a language function there is not a similar function for arrays, so I had to write one for my tests.

03_letters/runnerdave/main.go
runnerdave: got you, much simpler that way

runnerdave:
@n0npax Hi mate, I have addressed all your comments can you please approve?

03_letters/runnerdave/main_test.go
n0npax: Comparing arrays using reflect seems to be possible. 

Due to docs: https://golang.org/pkg/reflect/#DeepEqual
DeepEqual reports whether x and y are “deeply equal,” defined as follows. Two values of identical type are deeply equal if one of the following cases applies. Values of distinct types are never deeply equal.

Please check the following example:
```
   8   │ func main() {
   9   │     a := []int{1, 2, 3, 4}
  10   │     b := []int{1, 2, 3, 4}
  11   │     fmt.Println(reflect.DeepEqual(a, b))
```
I'm able to compare slices using the DeepEqua function here.

n0npax:
Hi @runnerdave, Added one comment the code. cheers:)

03_letters/runnerdave/main_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestSortingLetters(t *testing.T) {
+	tables := []struct {
+		x map[rune]int
+		n []string
+	}{
+		{map[rune]int{97: 1, 98: 3}, []string{"a:1", "b:3"}},
+		{map[rune]int{99: 1, 98: 3}, []string{"b:3", "c:1"}},
+		{map[rune]int{109: 1, 98: 3, 100: 5}, []string{"b:3", "d:5", "m:1"}},
+		{map[rune]int{97: 1, 65: 3}, []string{"A:3", "a:1"}},
```
n0npax: providing run as `'c'` instead of '`99` would be better more readable my mind. Please keep convention from line 35

03_letters/runnerdave/main_test.go
runnerdave: you are right, thanks for the tip I won't forget this one ever. I remember reading that DeepEqual did not work for slices, but I did not actually tried, my bad

03_letters/runnerdave/main_test.go
runnerdave: sorry you are right, fixed

runnerdave:
@n0npax hi mate, if you can help me finish this off that would be great, I have implemented the changes you have requested, thanks

n0npax (dismissed):
lgtm

anzboi (approved):
LGTM

PR #358, opened 2019-06-12
Complete lab06

### LINKED ISSUE #356: lab 06
Create an executable go program in directory 06_puppy/USERNAME (see hints)
* [x] Implement a Puppy struct containing ID, Breed, Colour, Value
* [x] Create Storer interface with crud methods for Puppy
* [x] Write a MapStore implementation of Storer backed by a map
* [x] Write a SyncStore implementation of Storer backed by a sync.Map
* [x] Keep all implementation files in the same folder and in package main
* [x] Test against the Storer interface and run in suite with both implementations

n0npax:
Implement CRUD for pet structure
Implement Sync and mem storer
Implement tests

Fixes #356 

Review of colleague's PR #352 

## Complete lab06
Implement CRUD for puppy structure
Implement tests

### DIFF
--- a/06_puppy/n0npax/memStore.go
+++ b/06_puppy/n0npax/memStore.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	"errors"
+	"fmt"
 )
 
 func NewMemStore() MemStore {
@@ -15,21 +15,26 @@ func (m MemStore) CreatePuppy(puppy *Puppy) int {
 }
 
 func (m MemStore) ReadPuppy(id int) (*Puppy, error) {
-	puppy, ok := m[id]
-	if !ok {
-		return nil, errors.New("doesn't exists")
+	if puppy, ok := m[id]; ok {
+		return puppy, nil
 	}
-	return puppy, nil
+	return nil, fmt.Errorf("puppy with ID: %d does not exist", id)
 }
 
 func (m MemStore) UpdatePuppy(id int, puppy *Puppy) error {
+	if _, ok := m[id]; !ok {
+		return fmt.Errorf("puppy with ID: %d does not exist", id)
+	}
+	if id != puppy.ID {
+		return fmt.Errorf("puppy ID corrupted")
+	}
 	m[id] = puppy
 	return nil
 }
 
 func (m MemStore) DeletePuppy(id int) (bool, error) {
 	if _, ok := m[id]; !ok {
-		return false, errors.New("doesn't exist")
+		return false, fmt.Errorf("puppy with ID: %d does not exist", id)
 	}
 	delete(m, id)
 	return true, nil
--- a/06_puppy/n0npax/memStore.go
+++ b/06_puppy/n0npax/memStore.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"errors"
+)
+
+func NewMemStore() MemStore {
+	return MemStore{}
+}
+
+func (m MemStore) CreatePuppy(puppy *Puppy) int {
+	id := len(m)
+	m[id] = puppy
+	return id
+}
+
+func (m MemStore) ReadPuppy(id int) (*Puppy, error) {
+	puppy, ok := m[id]
+	if !ok {
+		return nil, errors.New("doesn't exists")
+	}
+	return puppy, nil
+}
+
+func (m MemStore) UpdatePuppy(id int, puppy *Puppy) error {
+	m[id] = puppy
+	return nil
+}
+
+func (m MemStore) DeletePuppy(id int) (bool, error) {
+	if _, ok := m[id]; !ok {
+		return false, errors.New("doesn't exist")
+	}
+	delete(m, id)
+	return true, nil
+}
--- a/06_puppy/n0npax/store_test.go
+++ b/06_puppy/n0npax/store_test.go
@@ -11,7 +11,7 @@ type storerImpl int
 
 const (
 	syncStorer storerImpl = iota
-	memStorer  storerImpl = iota
+	memStorer
 )
 
 var (
@@ -41,15 +41,13 @@ func (s *storerSuite) SetupTest() {
 	switch s.impl {
 	case syncStorer:
 		s.store = NewSyncStore()
-		puppy := puppy0()
-		s.store.CreatePuppy(&puppy)
 	case memStorer:
 		s.store = NewMemStore()
-		puppy := puppy0()
-		s.store.CreatePuppy(&puppy)
 	default:
 		panic("Unrecognised storer implementation")
 	}
+	puppy := puppy0()
+	s.store.CreatePuppy(&puppy)
 }
 
 func (s *storerSuite) TestCreatePuppySuccessful() {
@@ -87,6 +85,22 @@ func (s *storerSuite) TestUpdatePuppy() {
 	assert.Equal(existingPuppy.Colour, puppy.Colour, "Updated colour missmatch")
 }
 
+func (s *storerSuite) TestUpdatePuppyCorruptedID() {
+	assert := tassert.New(s.T())
+	existingPuppy := puppy0()
+	oldID := existingPuppy.ID
+	existingPuppy.ID = 1000
+	err := s.store.UpdatePuppy(oldID, &existingPuppy)
+	assert.Error(err, "Should get an error when attempting to update with corrupted id")
+}
+
+func (s *storerSuite) TestUpdatePuppyIDDoesNotExist() {
+	assert := tassert.New(s.T())
+	newPuppy := puppy0()
+	err := s.store.UpdatePuppy(1000, &newPuppy)
+	assert.Error(err, "Should get an error when attempting to update an non-existing puppy")
+}
+
 func (s *storerSuite) TestDeletePuppySuccessful() {
 	assert := tassert.New(s.T())
 	existingPuppy := puppy0()
@@ -98,24 +112,13 @@ func (s *storerSuite) TestDeletePuppySuccessful() {
 	assert.Error(err, "Should not be able to read a deleted ID")
 }
 
-func (s *storerSuite) TestDeletePuppy_IDDoesNotExist() {
+func (s *storerSuite) TestDeletePuppyIDDoesNotExist() {
 	assert := tassert.New(s.T())
 	_, err := s.store.DeletePuppy(1000)
 	assert.Error(err, "Should not be able to delete puppy with non existing ID")
 }
 
-func TestStorerSyncMap(t *testing.T) {
-	syncSuite := storerSuite{
-		store: NewSyncStore(),
-		impl:  syncStorer,
-	}
-	suite.Run(t, &syncSuite)
-}
-
-func TestStorerMemoryMap(t *testing.T) {
-	syncSuite := storerSuite{
-		store: NewMemStore(),
-		impl:  memStorer,
-	}
-	suite.Run(t, &syncSuite)
+func TestStorer(t *testing.T) {
+	suite.Run(t, &storerSuite{impl: syncStorer})
+	suite.Run(t, &storerSuite{impl: memStorer})
 }
--- a/06_puppy/n0npax/store_test.go
+++ b/06_puppy/n0npax/store_test.go
@@ -1,0 +1,121 @@
+package main
+
+import (
+	"testing"
+
+	tassert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type storerImpl int
+
+const (
+	syncStorer storerImpl = iota
+	memStorer  storerImpl = iota
+)
+
+var (
+	puppy0 = func() Puppy {
+		return Puppy{
+			Breed:  "Norwegian Forest cat",
+			Colour: "Grey",
+			Value:  "$300",
+		}
+	}
+	puppy1 = func() Puppy {
+		return Puppy{
+			Breed:  "Maine Coon",
+			Colour: "Brown",
+			Value:  "€400",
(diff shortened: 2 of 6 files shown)


06_puppy/n0npax/store_test.go
```diff
@@ -0,0 +1,121 @@
...
+	assert := tassert.New(s.T())
+	expectedPuppy := puppy0()
+	puppy, err := s.store.ReadPuppy(expectedPuppy.ID)
+
+	if assert.NoError(err, "Should be able to read puppy0 from store") {
+		assert.Equal(&expectedPuppy, puppy, "store should return a puppy identical to puppy0")
+	}
+}
+
+func (s *storerSuite) TestReadPuppy_IDDoesNotExist() {
```
runnerdave: naming is a bit inconsistent, this test name has an underscore while the others do not

06_puppy/n0npax/syncStore.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"errors"
+)
+
+func NewSyncStore() *SyncStore {
+	return &SyncStore{}
+}
+
+func (m *SyncStore) CreatePuppy(puppy *Puppy) int {
+	// generate new id
```
runnerdave: unnecessary comment

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,37 @@
...
+	return puppy, nil
+}
+
+func (m MemStore) UpdatePuppy(id int, puppy *Puppy) error {
+	m[id] = puppy
+	return nil
+}
+
+func (m MemStore) DeletePuppy(id int) (bool, error) {
+
```
runnerdave: minor, but this empty line is not needed

runnerdave (dismissed):
good job, very minor suggestions

06_puppy/n0npax/memStore.go
n0npax: fixed. Thanks

06_puppy/n0npax/syncStore.go
n0npax: fixed. Thanks

06_puppy/n0npax/store_test.go
n0npax: You're right, I should avoid underscores in names. Fixed

06_puppy/n0npax/types.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import (
+	"sync"
+)
+
```
rokane, line 6: Great work separating the types out into their own file. Are you able to add some documentation to these exported types? It will be good to get in the habit of documenting any exported types, interfaces and functions as we start to move towards building an API.

General convention is a comment above the exported type/function, which starts with the type/function name. Eg.

```go
// Puppy contains information about a single Puppy ...
type Puppy struct {
    ...
}
```

06_puppy/n0npax/types.go
```diff
@@ -0,0 +1,25 @@
...
+type Puppy struct {
+	ID     int    `json:"id"`
+	Value  string `json:"value"`
+	Breed  string `json:"breed"`
+	Colour string `json:"colour"`
+}
+
+type Storer interface {
+	ReadPuppy(ID int) (*Puppy, error)
```
rokane, line 17: Minor little nit. But the naming of input parameters here is inconsistent. We have `ID` and `id` across the different interface functions. Pick one and use that consistently.

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+func (m MemStore) CreatePuppy(puppy *Puppy) int {
+	id := len(m)
+	m[id] = puppy
+	return id
+}
+
+func (m MemStore) ReadPuppy(id int) (*Puppy, error) {
+	puppy, ok := m[id]
```
rokane: If you flip the logic here, you can save a line of code.

```go
if puppy, ok := m[id]; ok {
    return puppy, nil
}
return nil, errors.New("doesn't exist")
```

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,36 @@
...
+	if !ok {
+		return nil, errors.New("doesn't exists")
+	}
+	return puppy, nil
+}
+
+func (m MemStore) UpdatePuppy(id int, puppy *Puppy) error {
+	m[id] = puppy
```
rokane, line 31: Can you include an error check here to see if the puppy exists? Since we are returning an error, it suggests we should be doing this kind of check.
```go
if _, ok := m[id]; !ok {
    return errors.New(fmt.Sprintf("Puppy with ID: %d does not exist", id))
}
m[id] = puppy
return nil
```

06_puppy/n0npax/syncStore.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	"errors"
+)
+
+func NewSyncStore() *SyncStore {
+	return &SyncStore{}
+}
+
+func (m *SyncStore) CreatePuppy(puppy *Puppy) int {
```
rokane, line 11: Here you should be making use of the `sync.Mutex` capability to put a mutex lock on the resource to ensure race conditions can't occur. If you extend your definition of the `SyncStore` you will be able to do this.

```go
type SyncStore struct {
    sync.Map
    sync.Mutex
}

func (m *SyncStore) CreatePuppy(puppy *Puppy) int {
    m.Lock()
    defer m.Unlock()

    // Continue with functionality below
}

```

06_puppy/n0npax/syncStore.go
```diff
@@ -0,0 +1,42 @@
...
+	var newID int
+	m.Range(func(_, _ interface{}) bool {
```
rokane: I also don't think it is required to iterate over the map everytime, we can turn this from a linear, to a constant runtime by storing the number of items as another variable in the `SyncStore`.

```go
type SyncStore struct {
    sync.Map
    sync.Mutex
    total    int
}
```

06_puppy/n0npax/syncStore.go
```diff
@@ -0,0 +1,42 @@
...
+func (m *SyncStore) ReadPuppy(id int) (*Puppy, error) {
+	puppyData, ok := m.Load(id)
+	if !ok {
+		return nil, errors.New("doesn't exists")
+	}
+	puppy := puppyData.(Puppy)
+	return &puppy, nil
+}
+
+func (m *SyncStore) UpdatePuppy(id int, puppy *Puppy) error {
```
rokane, line 29: Again here, I think there should be a check to see if the id exists, and if not a suitable error condition

06_puppy/n0npax/syncStore.go
```diff
@@ -0,0 +1,42 @@
...
+	m.Range(func(_, _ interface{}) bool {
+		newID++
+		return true
+	})
+	puppy.ID = newID
+	m.Store(puppy.ID, *puppy)
+	return puppy.ID
+}
+
+func (m *SyncStore) ReadPuppy(id int) (*Puppy, error) {
```
rokane, line 20: Again can flip the logic and save a line here as well. Same as described in the memstore

06_puppy/n0npax/store_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+	tassert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type storerImpl int
+
+const (
+	syncStorer storerImpl = iota
+	memStorer  storerImpl = iota
```
rokane: You don't need to specify this as two `iota`'s. You can just define this as below and they will be seperate enum values.

```go
const (
    syncStorer storerImpl = iota
    memStorer
)
```

06_puppy/n0npax/store_test.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+var (
+	puppy0 = func() Puppy {
```
rokane, line 18: Nice work 👍

06_puppy/n0npax/store_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+type storerSuite struct {
+	suite.Suite
+	store Storer
+	impl  storerImpl
+}
+
+func (s *storerSuite) SetupTest() {
+	switch s.impl {
+	case syncStorer:
```
rokane, line 42: Awesome work!! Great use of the SetupTest function.

Can you please extract the inserting of a puppy so it is outside the switch statement, that way you won't have to repeat it for each implementation. Saves a couple of lines of code.

06_puppy/n0npax/store_test.go
```diff
@@ -0,0 +1,121 @@
...
+
+func TestStorerSyncMap(t *testing.T) {
+	syncSuite := storerSuite{
+		store: NewSyncStore(),
+		impl:  syncStorer,
+	}
+	suite.Run(t, &syncSuite)
+}
+
+func TestStorerMemoryMap(t *testing.T) {
```
rokane: You can combine these two functions into one, and just run the suite against the different configurations. Also since you are using the `SetupTest` method, you are ensuring in there that the store is set. You should be able to remove this and delegate it to that function as a pre run task.

```go
func TestStorer(t *testing.T) {
    suite.Run(t, &storerSuite{impl: syncStorer})
    suite.Run(t, &storerSuite{impl: memStorer})
}
```

rokane (changes requested):
Awesome work. Your tests and code are looking really succinct and clean. I have suggested a few tidy up tasks throughout, can you please address them and let me know, I will aim to look at it so you can continue with the other lab.

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,39 @@
...
+	id := len(m)
+	m[id] = puppy
+	return id
+}
+
+func (m MemStore) ReadPuppy(id int) (*Puppy, error) {
+	if puppy, ok := m[id]; ok {
+		return puppy, nil
+	}
+	return nil, errors.New(fmt.Sprintf("Puppy with ID: %d does not exist", id))
```
golangcibot: should replace `errors.New(fmt.Sprintf(...))` with `fmt.Errorf(...)` (from `golint`)

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (m MemStore) UpdatePuppy(id int, puppy *Puppy) error {
+	if _, ok := m[id]; !ok {
+		return errors.New(fmt.Sprintf("Puppy with ID: %d does not exist", id))
```
golangcibot: should replace `errors.New(fmt.Sprintf(...))` with `fmt.Errorf(...)` (from `golint`)

06_puppy/n0npax/memStore.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	m[id] = puppy
+	return nil
+}
+
+func (m MemStore) DeletePuppy(id int) (bool, error) {
+	if _, ok := m[id]; !ok {
+		return false, errors.New(fmt.Sprintf("Puppy with ID: %d does not exist", id))
```
golangcibot: should replace `errors.New(fmt.Sprintf(...))` with `fmt.Errorf(...)` (from `golint`)

06_puppy/n0npax/memStore.go
n0npax: fixed. Thanks for that!

06_puppy/n0npax/store_test.go
n0npax: fixed Thanks for that!

06_puppy/n0npax/store_test.go
n0npax: Fixed. Good hint. Thanks!

06_puppy/n0npax/types.go
n0npax: Fixed. All interfaces have capitalized `ID` now. Thanks!

06_puppy/n0npax/types.go
n0npax: Fixed. Thanks!

06_puppy/n0npax/syncStore.go
n0npax: fixed. Thanks!

06_puppy/n0npax/syncStore.go
n0npax: fixed. Thanks!

06_puppy/n0npax/syncStore.go
n0npax: `create`, `update` & `delete` are using mutex now. I believe Read can be done without a mutex.

06_puppy/n0npax/store_test.go
n0npax: :rocket: Thanks!

06_puppy/n0npax/store_test.go
n0npax: fixed. Thanks

06_puppy/n0npax/memStore.go
n0npax: fixed. Thanks

n0npax:
Added `ID` corruption check in the midtime. I'd like to disallow updating `ID` to prevent the situation when the object is under `id`: 8 but internal `id`: is 9

rokane (approved):
Looks good, keep it up!!

PR #359, opened 2019-06-12
Lab 3 - Built Letter Frequency Analyser

### LINKED ISSUE #357: Lab 3 - Letter Frequency
Implement two functions:
- The first takes a string as input and produces a map of runes to integers that lists a count of characters as they appear in the string
- The second takes a map of runes to integers as input and produces an ordered slice of strings that are functionally equivalent to the map

joel00wood:
Created two functions, `letters(s string)` and `sortLetters(m map[rune]int)`, the first to output a map of runes and their frequencies in a given string, the second to provide a sorted slice of letters and their frequency in a given map.

Fixes #357 

Review of colleague's PR #320

- Implementation of `letters (s string)` and unit tests
- Implementation of `sortLetters(m map[rune]int)` and unit tests

### DIFF
--- a/03_letters/joel00wood/main.go
+++ b/03_letters/joel00wood/main.go
@@ -1,0 +1,29 @@
+package main
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+func main() {
+	fmt.Println(strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	out := map[rune]int{}
+	for _, runeValue := range s {
+		out[runeValue]++
+	}
+	return out
+}
+
+func sortLetters(m map[rune]int) []string {
+	out := make([]string, 0, len(m))
+	for r, c := range m {
+		str := fmt.Sprintf("%c:%v", r, c)
+		out = append(out, str)
+	}
+	sort.Strings(out)
+	return out
+}
--- a/03_letters/joel00wood/main_test.go
+++ b/03_letters/joel00wood/main_test.go
@@ -1,0 +1,99 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+
+	output "github.com/joel00wood/test-helpers/capture"
+)
+
+func TestMain(t *testing.T) {
+	expected := "a:2\nb:1\n"
+	actual := output.CaptureOutput(func() { main() })
+	if expected != actual {
+		t.Errorf("Unexpected result in main(), expected=%q, got=%q", expected, actual)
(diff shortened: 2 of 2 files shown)


03_letters/joel00wood/main.go
```diff
@@ -0,0 +1,30 @@
...
+	for _, runeValue := range s {
+		out[runeValue]++
+	}
+	return out
+}
+
+func sortLetters(m map[rune]int) []string {
+	out := []string{}
+	for r, c := range m {
+		str := string(r) + ":" + strconv.Itoa(c)
```
brnbke: Not sure what's "right", but as you already include fmt, can you use Sprintf rather than manual conversion & concatenation

03_letters/joel00wood/main.go
```diff
@@ -0,0 +1,30 @@
...
+func letters(s string) map[rune]int {
+	out := map[rune]int{}
+	for _, runeValue := range s {
+		out[runeValue]++
+	}
+	return out
+}
+
+func sortLetters(m map[rune]int) []string {
+	out := []string{}
```
brnbke: As you know the length of the array you want, you can allocate the size when defining, has some perf benefits - https://github.com/alexkohler/prealloc

brnbke (changes requested):

03_letters/joel00wood/main.go
joel00wood: seems reasonable to me, I'll look into it

03_letters/joel00wood/main.go
joel00wood: That's true, definitely seems worth implementing if not solely to get into the habit of doing so where appropriate

03_letters/joel00wood/main.go
joel00wood: Implemented

03_letters/joel00wood/main.go
joel00wood: Implemented

joel00wood:
@brnbke Implemented changes as described, please re-review when able

brnbke (dismissed):
LGTM

pentaphobe (dismissed):
Good stuff!

Am approving so I don't block the process, but will not mark as ready to merge just yet - please:

1. Rebase so you've got no more than 2 commits
2. Ensure the branch is updated from `master`
3. Label as "ready to merge" once the above steps are done

Thanks!

rokane (approved):
LGTM

PR #98, opened 2019-02-14
Lab4 and Lab5

osrnsree:
Fixes #70

#### Changes proposed in this PR:
* Add program to implement numeronym
* Add implemented test cases for numeronym implementation
* Add program to implement IP Address Type and also Stringer interface
* Add implemented test cases for Stringer interface implementation

PR #360, opened 2019-06-12
Lab4

### LINKED ISSUE #401: lab4
#360

hsy3418:
Fixes #401

Review of colleague's PR #98 

- Create a function for converting a list of string to a list of numeronyms


04_numeronym/hsy3418/main.go
```diff
@@ -0,0 +1,37 @@
...
+		countLetters := len(val)
+		if countLetters > 3 {
+			var buf strings.Builder
+			firstChar := val[:1]
+			lastChar := val[len(val)-1:]
+			midLength := strconv.Itoa(len(val) - 2)
+			buf.WriteString(firstChar)
+			buf.WriteString(midLength)
+			buf.WriteString(lastChar)
+			val = buf.String()
```
pfannerj: Not essential but IMO it might be tidier if you build the new numeronym value in a single line... like this 
numeronymsList[i] = val[0] + strconv.Itoa(len(val)-2) + val[len(val)-1].  And if you just set your numeronymsList := vals at the beginning then you don't need to do anything when the length is less than 3.

04_numeronym/hsy3418/main_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestNumeronym(t *testing.T) {
+
+	actual := numeronyms("accessibility", "Kubernetes", "abc")
+	expected := []string{"a11y", "K8s", "abc"}
```
pfannerj: Additional tests are required beyond the ones mentioned in the README... include some tests with spaces and other values so that you can be sure your function works as you expect with other non-standard inputs.  It will be useful to create your additional tests as 'table driven' tests as suggested in the following link...
https://github.com/anz-bank/go-course/wiki/Lab-2:-Collected-Code-Review-Comments

pfannerj (dismissed):
Not sure why this PR shows the files of lab3 also... I only reviewed the lab4 files.
 Also refer to the lab4 collected review comments too via this like...

https://github.com/anz-bank/go-course/wiki/Lab-4:-Collected-Code-Review-Comments

04_numeronym/hsy3418/main_test.go
```diff
@@ -0,0 +1,33 @@
...
+
+}
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("[a11y K8s a1c]\n")
```
pfannerj: Sorry I didn't notice this before but... the requirement is that 'abc' should return as 'abc' (not 'a1c').  Anything three chars or less should return as per how it was received..

04_numeronym/hsy3418/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"strconv"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+//numeronyms implements a function that returns a slice of numeronyms for its input strings
```
camscale: This comment is a little verbose. The "implements a function" part does not really add anything meaningful as we can see it is a function. Just describe what it does:
```
// numeronyms returns a slice of numeronyms for its input strings.
```
Also note, the full-stop at the end. Write these as complete sentences as they can be extracted as documentation.

04_numeronym/hsy3418/main.go
```diff
@@ -0,0 +1,30 @@
...
+func numeronyms(vals ...string) []string {
+
+	numeronymsList := vals
+
+	for index, val := range vals {
+		if len(val) >= 3 {
+			numeronymsList[index] = val[:1] + strconv.Itoa(len(val)-2) + val[len(val)-1:]
```
camscale: The two ends of the numeronyms are single characters, so using a colon in them makes no sense:
```
numernumsList[index] = val[0] + strconv.Itoa(len(val)-2) + val[len(val)-1]
```

04_numeronym/hsy3418/main.go
```diff
@@ -0,0 +1,30 @@
...
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+//numeronyms implements a function that returns a slice of numeronyms for its input strings
+func numeronyms(vals ...string) []string {
+
+	numeronymsList := vals
```
camscale: This is dangerous. It is not making a copy of the input so it is actually changing it. In your case, the input is an anonymous slice so you don't notice, but if `numeronyms` was called like this:
```
vals := []string{"accessibility", "Kubernetes", "abc}
fmt.Fprintln(out, numeronyms(vals...))
```
then your function would be changing the contents of the `vals` slice in the caller.

You should be making a new slice for the output value - `result := make([]string, len(vals))` and putting the numeronyms in that, and then returning it.

camscale: You shouldn't name variables with "types" in their name, so in this case, do not add "List" to the variable name. Just name it for what it is. In this case, just `result` would be sufficient. `numeronyms` would also be ok too, but that is the name of the function, and it's best not to shadow names like that.

04_numeronym/hsy3418/main_test.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestNumeronym(t *testing.T) {
+
+	actual := numeronyms("accessibility", "Kubernetes", "abc", "Android Enterprise", "", "ab")
+	expected := []string{"a11y", "K8s", "a1c", "A16e", "", "ab"}
```
camscale: As @pfannerj has said, "abc" should stay as "abc", not "a1c". Please fix the test case and the code.

camscale (dismissed):
A little bit of work still needed here.

Could you also remove all the commits from the other labs from this PR? I am happy to help if you want to visit in person - git can be a bit daunting for somethings, so please come by (ping me first on slack - camh) and I can show you how.

04_numeronym/hsy3418/main.go
```diff
@@ -0,0 +1,30 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+// numeronyms returns a slice of numeronyms for its input strings.
+func numeronyms(vals ...string) []string {
+
```
juliaogris: Unnecessary blank line.
(below as well).

04_numeronym/hsy3418/main_test.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestNumeronym(t *testing.T) {
```
juliaogris, line 9: Please add tests with non-ascii unicode chars, e.g. emojis or non-english chars.

juliaogris (changes requested):
Please add PR description, link to review of colleague's PR and update commit message to start with capital letter.

juliaogris (approved):
LGTM

I've fixed the Markdown formatting of your PR description, check out the history under the timestamps. It might be worthwhile to read up on Github Flavoured MarkDown:
* https://guides.github.com/features/mastering-markdown/
* https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet

PR #375, opened 2019-06-13
Implement tests and function to produce numeronyms

runnerdave:
Fixes [#67](https://github.com/anz-bank/go-course/issues/67)

Review of colleague's PR [#358 ](https://github.com/anz-bank/go-course/pull/358)

 - Implement tests for numeronyms
 - Implement function to pass tests for numeronyms


brnbke:
Looks good to me!

04_numeronym/runnerdave/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{[]string{"11111111"}, []string{"161"}},
+		{[]string{"?"}, []string{"?"}},
+		{[]string{"?$"}, []string{"?$"}},
+		{[]string{"?$&%"}, []string{"?2%"}},
+	}
+
+	for _, table := range tables {
+		numys := numeronyms(table.x...)
+		if !equal(numys, table.n) {
+			t.Errorf("Sorting of (%v) was incorrect, got: %v, want: %v.", table.x, numys, table.n)
```
rokane: I think this error message might not be indicative of the true failure. What if it were the case that the test failed because of the length's not being the same? Then it would still say "Sorting of ... was incorrect.... " I think it is fine to remove the "sorting" and just say "Unexpected output, expected ..., got ..."

04_numeronym/runnerdave/main.go
```diff
@@ -0,0 +1,27 @@
...
+	"unicode/utf8"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
```
rokane, line 16: Great work!

04_numeronym/runnerdave/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	t.Logf("actual:%s", actual)
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestNumeronyms(t *testing.T) {
+	tables := []struct {
+		x []string
```
rokane: I think these variable names are a little confusing, I don't know what the x is supposed to stand for. I understand it is the input to the function, but why `x`? Can we update these so it is a little easier to read?
Perhaps, `x -> input` and `n -> expected | output | exp` ?

Will help later on with readability when looking at the line where we check for equality
```go
if !equal(numys, table.expected) {
    ...
}
```

rokane (changes requested):
Looking really good. I have just a couple of requests to clean up the tests a little, and then should be fine afterwards :)

04_numeronym/runnerdave/main_test.go
runnerdave: good point, thanks

04_numeronym/runnerdave/main_test.go
runnerdave: You are right, fixed

04_numeronym/runnerdave/main.go
runnerdave: thanks a lot

rokane (approved):
LGTM 👍 well done!

PR #376, opened 2019-06-13
Lab4 Numeronym

### LINKED ISSUE #377: Lab4 Numeronym

pfannerj:
Fixes #377

Review of colleague's PR #360

- build lab4 numeronyms function
- build lab4 numeronyms test cases

### DIFF
--- a/04_numeronym/pfannerj/main.go
+++ b/04_numeronym/pfannerj/main.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+)
+
+var mainout io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(mainout, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
+	//Regular expression can be changed or removed altogether if additional chars need to be included
+	regEx := regexp.MustCompile("[^a-zA-Z0-9]+")
+	valsOut := vals
+	for i := range vals {
+		alphaNumVal := regEx.ReplaceAllString(vals[i], "") //Strips all non-alpha-numeric characters
+		length := len(alphaNumVal)
+		if length > 3 {
+			valsOut[i] = fmt.Sprintf("%v%d%v", string(alphaNumVal[0]), length-2, string(alphaNumVal[length-1]))
+		}
+	}
+	return valsOut
+}
--- a/04_numeronym/pfannerj/main_test.go
+++ b/04_numeronym/pfannerj/main_test.go
@@ -1,0 +1,47 @@
+package main
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	mainout = &buf
+	main()
+	expected := "[a11y K8s abc]\n"
+	actual := buf.String()
+	if expected != actual {
+		t.Errorf(expected, actual, "Unexpected output in main()")
+	}
+}
+
+var numeronymsCases = map[string]struct {
+	input, expected []string
+}{
+	"Standard":         {input: []string{"accessibility", "Kubernetes", "abc"}, expected: []string{"a11y", "K8s", "abc"}},
+	"OneTwoThreeFour":  {input: []string{"x", "yz", "xyz", "wxyz"}, expected: []string{"x", "yz", "xyz", "w2z"}},
+	"WithSpaceAndPunc": {input: []string{"I like you", "you are very NICE!"}, expected: []string{"I6u", "y12E"}},
+	"OneLongArg":       {input: []string{"supercalifragilisticexpialidocious"}, expected: []string{"s32s"}},
+	"Return":           {input: []string{"AB\nBA"}, expected: []string{"A2A"}},
+	"RawAndStd":        {input: []string{`a11y`, "Kubernetes", `abc`}, expected: []string{"a2y", "K8s", "abc"}},
+	"Numeric":          {input: []string{"1234567890"}, expected: []string{"180"}},
+	"EscapedQuote":     {input: []string{"a\":2\"b:\"1"}, expected: []string{"a21"}},
+	"RawQuote":         {input: []string{`a":2"b:"1`}, expected: []string{"a21"}},
+	"EmptyStrings":     {input: []string{``, ""}, expected: []string{"", ""}},
+	"RawReturn": {input: []string{`a1
+	1y`}, expected: []string{"a2y"}},
+}
+
+func TestNumeronyms(t *testing.T) {
+	for name, tc := range numeronymsCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			actual := numeronyms(tc.input...)
+			if !reflect.DeepEqual(tc.expected, actual) {
+				t.Errorf("Numeronyms function failed. Input: %v, Actual: %v, Expected: %v", tc.input, actual, tc.expected)
+			}
+		})
+	}
+}


04_numeronym/pfannerj/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+var numeronymsCases = map[string]struct {
+	input    []string
+	expected []string
+}{
+	"Standard":        {input: []string{"accessibility", "Kubernetes", "abc"}, expected: []string{"a11y", "K8s", "abc"}},
+	"OneTwoThreeFour": {input: []string{"x", "yz", "xyz", "wxyz"}, expected: []string{"x", "yz", "xyz", "w2z"}},
+	// "Empty":       {input: "", expected: map[rune]int{}},
+	// "Return":      {input: "ab\na\n", expected: map[rune]int{10: 2, 97: 2, 98: 1}},
+	// "Slash":       {input: "ab/na/n", expected: map[rune]int{47: 2, 97: 2, 98: 1, 110: 2}},
+	// "Huge":        {input: "zabaaarzzzASYbcd@&^!*JHDJKHHGFGD^!*Dar", expected: map[rune]int{33: 2, 38: 1, 42: 2, 64: 1, 65: 1, 68: 3, 70: 1, 71: 2, 72: 3, 74: 2, 75: 1, 83: 1, 89: 1, 94: 2, 97: 5, 98: 2, 99: 1, 100: 1, 114: 2, 122: 4}},
```
golangcibot: line is 236 characters (from `lll`)

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+		expected: []string{"!:2", "&:1", "*:2", "@:1", "A:1", "D:3", "F:1", "G:2", "H:3", "J:2", "K:1",
+			"S:1", "Y:1", "^:2", "a:5", "b:2", "c:1", "d:1", "r:2", "z:4"}},
+	"Empty": {input: map[rune]int{}, expected: []string{}},
+}
+
+func TestSortLetters(t *testing.T) {
+	for name, tc := range sortLettersCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			actual := sortLetters(tc.input)
```
golangcibot: undeclared name: `sortLetters` (from `typecheck`)

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"Numeric":      {input: "11235", expected: map[rune]int{49: 2, 50: 1, 51: 1, 53: 1}},
+	"EscapedQuote": {input: "a\":2\"b:\"1", expected: map[rune]int{34: 3, 49: 1, 50: 1, 58: 2, 97: 1, 98: 1}},
+	"RawQuote":     {input: `a":2"b:"1`, expected: map[rune]int{34: 3, 49: 1, 50: 1, 58: 2, 97: 1, 98: 1}},
+}
+
+func TestLetters(t *testing.T) {
+	for name, tc := range lettersCases {
+		tc := tc
+		t.Run(name, func(t *testing.T) {
+			actual := letters(tc.input)
```
golangcibot: undeclared name: `letters` (from `typecheck`)

03_letters/pfannerj/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	mainout = &buf
```
golangcibot: undeclared name: `mainout` (from `typecheck`)

04_numeronym/pfannerj/main.go
```diff
@@ -0,0 +1,33 @@
...
+func numeronyms(vals ...string) []string {
+
+	valsOut := vals
+	for i := range vals {
+		//Regular expression can be changed or removed altogether if additional chars need to be included
+		regEx := regexp.MustCompile("[^a-zA-Z0-9]+")
+		alphaNumVal := regEx.ReplaceAllString(vals[i], "") //Strips all non-alpha-numeric characters
+		if len(alphaNumVal) > 3 {
+			alphaNumChar := strings.Split(alphaNumVal, "")
+			valsOut[i] = alphaNumChar[0] + strconv.Itoa(len(alphaNumVal)-2) + alphaNumChar[len(alphaNumVal)-1]
```
joel00wood: These two lines could be simplified to one as the string doesn't need to be split into chars. Seeing as the first char and the last char are the only values it requires, it could equivalently be completed like so:
```
valOut[i] = alphaNumVal[:1] + strconv.Itoa(len(alphaNumVal)-2) + alphaNumVal[len(alphaNumVal)-1:]
```

joel00wood (commented):
The approach taken to this lab is quite comprehensive, including accounting for non-alphanumeric characters.
I only really had one comment towards simplifying it a bit further but otherwise it looks good.

04_numeronym/pfannerj/main.go
```diff
@@ -0,0 +1,33 @@
...
+
+	fmt.Fprintln(mainout, numeronyms("accessibility", "Kubernetes", "abc"))
+
+}
+
+func numeronyms(vals ...string) []string {
+
+	valsOut := vals
+	for i := range vals {
+		//Regular expression can be changed or removed altogether if additional chars need to be included
```
rokane: I think this might be making the task a little more complicated that what was asked, as the lab never stated anything about ensuring the input was alphanumeric. So I think this isn't necessarily required, but if you want to proceed with this I am happy to let it go. Feel free to do either, just as long as the test cases reflect the implementation.

If you do keep it, I think the `regEx` variable should be moved so it is initialised outside the loop, rather than re initialising it each iteration, it will be constant for all.

04_numeronym/pfannerj/main.go
rokane: Here you can simplify this, there shouldn't be a need to use the `strings.Split()` method, you should be able to get the values directly from the `alphaNumVal`. If you use the `fmt.Sprintf()` method as well you can do something like below:

```go
length := len(alphaNumVal)
if length > 3 {
    valsOut[i] = fmt.Sprintf("%v%d%v", string(alphaNumVal[0]), length-2, string(alphaNumVal[length-1]))
}
```

And if you combine this with also removing the regex statements above you can simplify the logic heaps by using the value returned from the `range` function.

```go
func numeronyms(vals ...string) []string {
    output := vals
    for i, v := range vals {
        length := len(v)
        if length > 3 {
            output[i] = fmt.Sprintf("%v%d%v", string(v[0]), length-2, string(v[length-1]))
        }
    }
    return output
}
```

04_numeronym/pfannerj/main_test.go
```diff
@@ -0,0 +1,46 @@
...
+	expected := "[a11y K8s abc]\n"
+	actual := buf.String()
+	if expected != actual {
+		t.Errorf(expected, actual, "Unexpected output in main()")
+	}
+}
+
+var numeronymsCases = map[string]struct {
+	input, expected []string
+}{
```
rokane, line 22: Good set of test cases. Maybe the empty string could be another edge case you could consider

04_numeronym/pfannerj/main.go
```diff
@@ -0,0 +1,33 @@
...
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+var mainout io.Writer = os.Stdout
+
+func main() {
+
+	fmt.Fprintln(mainout, numeronyms("accessibility", "Kubernetes", "abc"))
```
rokane, line 13: Little nit, but I don't think the extra white space is required here, please remove.

rokane (changes requested):
Looking really good so far! Take a look at my comments, I requested a couple of changes and also a suggestion for simplifying the function. Once those are addressed I'll take another look 👍

04_numeronym/pfannerj/main_test.go
pfannerj: Yes, I will include an empty string test too....

04_numeronym/pfannerj/main.go
pfannerj: Oh, that's heaps better... I thought I needed to do the string.Split in order to be able to refer to the [0] values etc.  Will definitely make these changes.  Thx!

04_numeronym/pfannerj/main.go
pfannerj: Cool.. I wasn't sure if it can really be called a Numeronym if it contains weird characters so I thought this was a good way to cater for that by just stripping them.  I think I will leave the regex in place but will defs move it outside the loop... thx!!

04_numeronym/pfannerj/main.go
pfannerj: no probs.. will remove...

rokane (approved):
Really good work!! 👍

PR #366, opened 2019-06-13
Lab1

srihariInfy:
Fixes #

Review of colleague's PR #

PR #385, opened 2019-06-15
Implement type for IP address with stringer interface plus tests

runnerdave:
Fixes [#70](https://github.com/anz-bank/go-course/issues/70) 

Review of colleague's PR #366 

 - Write tests for a function int the spec in Readme.md
 - Implement type for IPAddr that represents an IP address
 - Implement stringer interface for this type
 - Pass tests

### DIFF
--- a/05_stringer/runnerdave/main_test.go
+++ b/05_stringer/runnerdave/main_test.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("127.0.0.1\n")
+	actual := strconv.Quote(buf.String())
+	t.Logf("expected:%s", expected)
+	t.Logf("actual:%s", actual)
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestStringer(t *testing.T) {
+	tables := []struct {
+		input  IPAddr
+		output string
+	}{
+		{IPAddr{127, 0, 0, 1}, "127.0.0.1"},
+		{IPAddr{10, 0, 0, 1}, "10.0.0.1"},
+		{IPAddr{191, 255, 255, 255}, "191.255.255.255"},
+		{IPAddr{10, 10}, "10.10.0.0"},
+		{IPAddr{}, "0.0.0.0"},
+	}
+
+	for _, table := range tables {
+		expected := table.output
+		actual := table.input.String()
+		if expected != actual {
+			t.Errorf("Stringer of (%#v) was incorrect, got: %s, want: %s.", table.input, actual, expected)
+		}
+	}
+}
--- a/05_stringer/runnerdave/main.go
+++ b/05_stringer/runnerdave/main.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// IPAddr represents an IP address
+type IPAddr [4]byte
+
+var out io.Writer = os.Stdout
+
+func (ip IPAddr) String() string {
+	return fmt.Sprintf("%d.%d.%d.%d", ip[0], ip[1], ip[2], ip[3])
+}
+
+func main() {
+	fmt.Fprintln(out, IPAddr{127, 0, 0, 1})
+}


05_stringer/runnerdave/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func TestFibo(t *testing.T) {
+	tables := []struct {
+		x IPAddr
+		n string
+	}{
+		{IPAddr{127, 0, 0, 1}, "127.0.0.1"},
+		{IPAddr{10, 0, 0, 1}, "10.0.0.1"},
+		{IPAddr{191, 255, 255, 255}, "191.255.255.255"},
```
n0npax, line 32: good job mate. Just one minor nit. Could you add test for empty and half initialized IPAddr
`IPAddr{10,10}, "10.10.0.0"`
`IPAddr{}, "0.0.0.0"`

n0npax (commented):
Good job. 1 minor thing and I can approve. Cheers!

05_stringer/runnerdave/main_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := strconv.Quote("127.0.0.1\n")
+	actual := strconv.Quote(buf.String())
+	t.Logf("expected:%s", expected)
```
brnbke, line 17: More of a question/comment than a change, but in this test you log the expected and actual always, and maybe an error, on your other test you only log when there's an error. Expending this out if you have a lot of tests, would it get a bit noisy to log during success criteria?

runnerdave: hi @brnbke this reflects my TDD approach, note that the logging is only in the single test not on the table one, because I write a failing test and then write the function to pass it, if I added logging in the other test I agree the noise would not be useful at all.

05_stringer/runnerdave/main_test.go
runnerdave: implemented, thanks for the input

n0npax (dismissed):
lgtm

05_stringer/runnerdave/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+	t.Logf("actual:%s", actual)
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
+	}
+}
+
+func TestFibo(t *testing.T) {
```
pentaphobe: Is it really testing fibonacci? ;)

Totally a nitpick, but good to watch out for such things when copy-pasting especially if they're publicly exported.

(Or totally avoid copy-pasting when first acclimatising to a new language)

pentaphobe (dismissed):
LGTM also

keep an eye out for incorrect copypasta though!

05_stringer/runnerdave/main_test.go
anzboi: Please change this.

anzboi (dismissed):

anzboi:
Change the test function name then all good.

05_stringer/runnerdave/main_test.go
runnerdave: hehe good catch

05_stringer/runnerdave/main_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{IPAddr{10, 10}, "10.10.0.0"},
+		{IPAddr{}, "0.0.0.0"},
+	}
+
+	for _, table := range tables {
+		var buf bytes.Buffer
+		out = &buf
+		fmt.Fprintln(out, table.x)
+		expected := strconv.Quote(table.n)
+		actual := strconv.Quote(strings.Replace(buf.String(), "\n", "", -1))
```
camscale: Using `Fprintln` to call the `String` method and then stripping off the `\n` added by `Fprintln` seems unnecessary. You could just call `Fprint` instead (https://golang.org/pkg/fmt/#Fprint) and not have a newline to strip off.

I'd also consider removing `Fprint(ln)` altogether and just call `table.x.String()`. Using a buffer to capture the output is required for `main`, but when testing a function with inputs and outputs, it makes more sense to deal directly with those inputs and outputs with nothing in between.

Finally, since there are no special characters in the string literals now, there is no need to use `strconv.Quote`.

With all this in place, your test should just be:
```
expected := table.n
actual := table.x.String()
```

That is much simpler to read an understand.

On the subject of reading and understanding, using `n` and `x` for the fields in your struct are rather meaningless and even misleading. `n` is usually some sort of count (int). I think naming them something like `input` (or `in`) and `expected` (or `out`) would make this more readable. Go encourages short variable names, but they need to be meaningful. In this context, `n` and `x` are not meaningful. They just seem to be random characters.

If you are unsure of naming, it is often useful to look at the Go library source code. E.g. https://golang.org/src/strconv/atob_test.go (the strconv package) uses `in` and `out`, so following that means you likely wont go wrong.

camscale (commented):
Just a general comment on making tests simple/minimal and readable. Let me know if you disagree - I'll approve given a reasonable counter-argument.

05_stringer/runnerdave/main_test.go
runnerdave: @camh-anz  this is really valuable feedback thanks a lot, will implement later

05_stringer/runnerdave/main_test.go
```diff
@@ -0,0 +1,44 @@
...
+		{IPAddr{191, 255, 255, 255}, "191.255.255.255"},
+		{IPAddr{10, 10}, "10.10.0.0"},
+		{IPAddr{}, "0.0.0.0"},
+	}
+
+	for _, table := range tables {
+		expected := table.output
+		actual := table.input.String()
+		if expected != actual {
+			t.Errorf("Stringer of (%d) was incorrect, got: %s, want: %s.", table.input, actual, expected)
```
camscale: Sorry I didn't notice this before, but the `%d` format specifier here is incorrect. I would suggest `%#v` which will give a Go-syntax representation of the value (https://golang.org/pkg/fmt/) as shown here: https://goplay.space/#BiRitf92ruM (`%v` would call the `String()` method which is counter-productive here)

camscale (changes requested):
Sorry, one thing I missed last time...

05_stringer/runnerdave/main_test.go
runnerdave: implemented

camscale (approved):
LGM. Thanks for your patience working through this. Looks great now.

PR #390, opened 2019-06-16
Lab 4 - Implemented Numeronyms

### LINKED ISSUE #361: Lab 4 - Numeronyms
Implements function `numeronyms()` which takes a slice of strings as input and outputs a slice of strings converted to numeronyms. Based on the sample response it also requires the output slice to be in the same order as the input slice.

joel00wood:
Created `numeronym()` and `numeronyms()` functions. `numeronym()` takes a string as input and returns a string with the first and last character wrapped around it's length -2. `numeronyms()` takes a variadic string slice as input and returns a slice of those strings converted to numeronyms.

Fixes #361 

Review of colleague's PR #376 

- Numeronym converter

### DIFF
--- a/04_numeronym/joel00wood/main.go
+++ b/04_numeronym/joel00wood/main.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+)
+
+func main() {
+	fmt.Println(numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronym(s string) string {
+	out := s
+	sLen := len(s)
+	midLen := sLen - 2
+	if midLen > 1 {
+		out = fmt.Sprintf("%s%v%s",
+			s[:1], midLen, s[sLen-1:])
+	}
+	return out
+}
+
+func numeronyms(vals ...string) []string {
+	out := make([]string, len(vals))
+	for i, val := range vals {
+		out[i] = numeronym(val)
+	}
+	return out
+}
--- a/04_numeronym/joel00wood/main_test.go
+++ b/04_numeronym/joel00wood/main_test.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"bytes"
+	"io"
+	"log"
+	"os"
+	"reflect"
+	"sync"
+	"testing"
+)
+
+// enter borrowed code
+// https://gist.github.com/hauxe/e935a7f9012bf2649710cf75af323dbf
+// stdOut capture helper
+func captureOutput(f func()) string {
+	reader, writer, err := os.Pipe()
+	if err != nil {
+		panic(err)
+	}
+	stdout := os.Stdout
+	stderr := os.Stderr
+	defer func() {
+		os.Stdout = stdout
+		os.Stderr = stderr
+		log.SetOutput(os.Stderr)
+	}()
+	os.Stdout = writer
+	os.Stderr = writer
+	log.SetOutput(writer)
+	out := make(chan string)
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	go func() {
+		var buf bytes.Buffer
+		wg.Done()
+		_, err := io.Copy(&buf, reader)
+		if err != nil {
+			panic(err)
+		}
+		out <- buf.String()
+	}()
+	wg.Wait()
+	f()
+	writer.Close()
+	return <-out
+}
+
+func TestMain(t *testing.T) {
+	expected := "[a11y K8s abc]\n"
+	actual := captureOutput(func() { main() })
+	if expected != actual {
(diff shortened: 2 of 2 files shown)


04_numeronym/joel00wood/main_test.go
```diff
@@ -0,0 +1,101 @@
...
+		[]string{"a", "ab", "abc", "a2d"},
+	},
+	"Some more complicated words": {
+		[]string{"feremone", "guttulous", "synchronous", "isopropenyl", "mericarp",
+			"spleened", "Universalian", "appeacher", "antidisestablishmentarianism"},
+		[]string{"f6e", "g7s", "s9s", "i9l", "m6p", "s6d", "U10n", "a7r", "a26m"},
+	},
+}
+
+func testNumeronym(t *testing.T) {
```
golangcibot: `testNumeronym` is unused (from `deadcode`)

04_numeronym/joel00wood/main.go
```diff
@@ -0,0 +1,35 @@
...
+func numeronym(s string) string {
+	var out string
+	len := len(s)
+	midLen := len - 2
+	if midLen > 1 {
+		start := s[:1]
+		mid := strconv.Itoa(midLen)
+		end := s[len-1:]
+		out = start + mid + end
+	} else {
```
pentaphobe: purely philosophical / nitpick:

I tend to avoid fallback `else` wherever I can, ie. your `out` declaration could just be `out := s` at the top of the function and you could get rid of the `else`.

What are your thoughts?

04_numeronym/joel00wood/main.go
```diff
@@ -0,0 +1,35 @@
...
+	"strconv"
+)
+
+func main() {
+	fmt.Println(numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronym(s string) string {
+	var out string
+	len := len(s)
```
pentaphobe: stylistic / nitpick: I tend to avoid reusing names like this (`len`) - smart compilers can work out needs from context, but it can still tack on a little extra cognitive load when reading long code

and hypothetically, if `len()` returned a function - what would be behaviour of future calls (in scope) to it after this line?

pentaphobe (commented):
LGTM!

Nice work

04_numeronym/joel00wood/main.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	return out
+}
+
+func numeronyms(vals ...string) []string {
+	out := []string{}
+	for _, val := range vals {
+		nym := numeronym(val)
+		out = append(out, nym)
```
pentaphobe: FWIW: `append` is quite slow since it creates a whole new array every iteration

Since you know the length of the output array is the same as your input array you could just as easily do:

```go
out := make([]string, len(vals))
for idx, val := range values {
    out[idx] = numeronym(val)
}

return out
```

pentaphobe: This (`make()`) makes a sized slice, you could also just use a regular array

pentaphobe: Live example:

https://goplay.space/#ok7pqrqoCL_W

04_numeronym/joel00wood/main.go
joel00wood: In terms of code readability that does make more sense. Whereby you don't need to wade through a more complex outcome to read through to the simpler one.

In the context of this function it seems particularly fitting given that the end result is still the input, whether it's been manipulated into a numeronym or not. I think part of my reasoning here was that in an earlier lab I got burned by not making the separation between input slice and output slice. But given that this isn't slices it seems reasonable to make this change.

04_numeronym/joel00wood/main.go
joel00wood: I'd imagine they'd return the value of len as set here and my unfortunate shortcut would have a less than desirable result, as such I've switched up for `sLen`.

04_numeronym/joel00wood/main.go
joel00wood: Made the change as described

pentaphobe:
Oh, huh - I'm not a codeowner.

Will seek to rectify that!

04_numeronym/joel00wood/main.go
pentaphobe: Oh I wasn't trying to force that one one ye, was more just an "interesting" side note, but cool :)

pentaphobe (dismissed):

pentaphobe:
@joel00wood pending branch update, but otherwise good to go

juliaogris (approved):
lgtm

PR #391, opened 2019-06-16
Lab3

### LINKED ISSUE #399: Lab3 - Letter frequency functions
Creating two function one that returns a mapping of each letter to its frequency for a given string and another that returns a sorted slice of strings with elements "{key}:{val}" for a given Map

brnbke:
Fixes #399

Review of colleague's PR #359

Adds following functions:
letters -  For a given string returns a mapping of each letter to its frequency
sortLetters - For a given map returns a sorted slice of strings with elements "{key}:{val}"

### DIFF
--- a/03_letters/brnbke/main.go
+++ b/03_letters/brnbke/main.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	result := make(map[rune]int)
+
+	for _, r := range s {
+		result[r]++
+	}
+
+	return result
+}
+
+func sortLetters(m map[rune]int) []string {
+	var result = make([]string, 0, len(m))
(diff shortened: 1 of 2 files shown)


pentaphobe (changes requested):
Looking good (epic tests!)

Not sure how strict we're being on the "create an issue for your PR" rule after Lab 1, but don't see that as a problem (it's mostly an exercise in process)

Do need to include a colleague PR review reference though [see here](https://github.com/anz-bank/go-course/wiki/Lab-4:-Collected-Code-Review-Comments)

Also, believe we're being fairly strict about the two-commit limit (likewise an exercise in process)

brnbke:
@pentaphobe Thanks for pointing that out, actually missed the whole "create an issue for your PR" thing, have updated this and other labs accordingly. 

Squashed commits into single one for now and added review of other PR (was doing it whilst looking for comments on this)

pentaphobe (approved):

pentaphobe:
LGTM!

pentaphobe (approved):
(Second approval)

PR #408, opened 2019-06-17
Lab 3 - Letter frequency

### LINKED ISSUE #407: lab3 - Letter frequency

```
```

```

### PR DESCRIPTION
**Pull request by mohitnag**
Fixes #407 

Review of colleague's PR #371

### CODE REVIEW

File: 03_letters/mohitnag/main_test.go [hunk 1]
```diff
@@ -0,0 +1,57 @@
...
+func TestMain(t *testing.T) {
+	assert := assert.New(t)
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := "a:2,b:1"
+	actual := buf.String()
+	actual = strings.Replace(actual, "\n", ",", -1)
+	actual = strings.TrimRight(actual, ",")
+	assert.Equal(expected, actual)
```
alfredxiao, line 37: I would probably just assign expected as `a:2\nb:1\n` instead of replacing and trimming.

03_letters/mohitnag/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	m := map[rune]int{}
+	rs := []rune(s)
```
rokane: This step is not required. In go when you iterate over a string like `s`, it will treat each character in the string as a rune. Therefore you can directly assign to the `map[rune]int{}` from `s`. Can you please update this?

```go
func letters(s string) map[rune]int {
	m := map[rune]int{}
	for _, val := range s {
		// val here is a rune.
	}
	return m
}
```

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLetters(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    string
+		expected map[rune]int
+	}{
+		{"Scenario One", "eaa!baabq23123!", map[int32]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1}},
```
rokane: I like the way you have structured your test cases here, but we can do a couple of things to improve this further.

1. You have stated the expected variable is of type `map[rune]int` however when assigning values for this you are using a `map[int32]int`. This makes things a lot harder to interpret as you are assigning the int representation of the rune. Can you change this so your keys to the expected map are actual rune characters?

2. Your `Scenario` name could be a bit more descriptive. The names `"Scenario One"` (and two) don't give much information about what it is testing. These should be something more like, "Empty String" etc which describes the input under test.

3. You could add a few more test cases as well as the empty and non empty cases. Since we are using runes, why don't you add a case which involves special characters, and emoji unicode characters?

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+		{"Scenario two", "", map[int32]int{}},
+	}
+	for _, td := range testData {
```
rokane: Another alternative to using the `[]struct {Scenario string, input string, expected map[rune]int}` is to use:
`map[string]struct{input string, expected map[rune]int}`. Then you will be able to access the name as the key of the map when iterating and not have to extract it from the struct variable.
```go
for scenario, td := range testData {
    t.Run(scenario, func(t *testing.T){ ... })
}
```

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+	assert.Equal(expected, actual)
+}
+
+func TestSortLetters(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    map[rune]int
+		expected []string
+	}{
+		{"Scenario One", map[int32]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1},
```
rokane: Same things apply here as the above `TestLetters` feedback

03_letters/mohitnag/main_test.go
rokane: Agree with @alfredxiao, or alternatively use the '`' string definitions and do the input with actual new lines.

rokane (changes requested):
Looking good so far, I've left some feedback for your testing files. If you can look through and apply that would be great. Also I have taken a look at your commit history for this PR and it looks as through there are a few commits which have carried across from a previous lab. Can you please rebase these and clean them up. You should have 1-2 commits for a lab this big, and include descriptive information in the body of the commit message to assist the reader.

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func TestSortLetters(t *testing.T) {
+	type testCases struct {
+		input    map[rune]int
+		expected []string
+	}
+	testData := map[string]testCases{
+
+		"Happy Case":                          {map[rune]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1}, []string{"!:2", "1:1", "2:2", "3:2", "a:4", "b:2", "e:1", "q:1"}},
```
golangcibot: line is 180 characters (from `lll`)

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func TestSortLetters(t *testing.T) {
+	type testCases struct {
+		input    map[rune]int
+		expected []string
+	}
+	testData := map[string]testCases{
+
+		"Happy Case":
```
golangcibot: File is not `goimports`-ed (from `goimports`)

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+func TestSortLetters(t *testing.T) {
+	type testCases struct {
+		input    map[rune]int
+		expected []string
+	}
+	testData := map[string]testCases{
+
+		"Happy Case":                          
+			{map[rune]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1}, []string{"!:2", "1:1", "2:2", "3:2", "a:4", "b:2", "e:1", "q:1"}},
```
golangcibot: line is 142 characters (from `lll`)

03_letters/mohitnag/main.go
mohitnag: Done

03_letters/mohitnag/main_test.go
mohitnag: Agree. Done the changes

03_letters/mohitnag/main_test.go
mohitnag: Done

03_letters/mohitnag/main_test.go
mohitnag: Done

03_letters/mohitnag/main_test.go
mohitnag: Done

03_letters/mohitnag/main_test.go
mohitnag: Done

03_letters/mohitnag/main_test.go
rokane: With number one this is still not fully complete. You have updated this to be of type rune, but you should be using the actual characters which it refers to rather than the int representation. This improves readability.

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func TestLetters(t *testing.T) {
+	type testCases struct {
+		input    string
+		expected map[rune]int
+	}
+	testData := map[string]testCases{
+		"Happy Case": {"eaa!baabq23123!",
+			map[rune]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1}},
```
rokane: Here this isn't complete. You are still assigning the values for the keys in the maps as int representations. Use the actual characters, this improves readability by a mile.

If you know you have an input string of "aaaabbbb" then I would expect an output of `map[rune]int{'a':3, 'b':3}` that is so much easier to read than the int representation of the rune.

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+		"Empty string should return empty map": {"", map[rune]int{}},
+		"Unicode string test":                  {"\u00C0\u00C0\u00C0\u00C0\u00C1\u00C1", map[rune]int{192: 4, 193: 2}},
```
rokane: You can also do the same here. If you are on a mac, you can access the emoji keyboard by clicking "command, control, space" which allows you to enter all kinds of special unicode values. Use this and it will be a lot clearer.
Eg.

```go
"Unicode string test":                  {"😀😀😀😛😛😛", map[rune]int{😀: 3, 😛: 3}}
```

03_letters/mohitnag/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func TestSortLetters(t *testing.T) {
+	type testCases struct {
+		input    map[rune]int
+		expected []string
+	}
+	testData := map[string]testCases{
+
+		"Happy Case": {map[rune]int{33: 2, 49: 1, 50: 2, 51: 2, 97: 4, 98: 2, 101: 1, 113: 1},
```
rokane: And also here.

rokane (changes requested):
Almost there. Just finish off the last piece and this should be good.

03_letters/mohitnag/main_test.go
mohitnag: Agree. Done

03_letters/mohitnag/main_test.go
mohitnag: Agree. Done

03_letters/mohitnag/main_test.go
mohitnag: Done

rokane (approved):
LGTM. Well done Mohit

PR #386, opened 2019-06-15
Lab1 fibonacci - alfredxiao

alfredxiao:
- Fixes #413
- Review of colleague's PR #412
Implemented fibonacci and tests.

PR #410, opened 2019-06-17
Lab4- Numeronym

### LINKED ISSUE #409: Lab 4 - Numeronym
- Create an executable go program in directory `04_numeronym/USERNAME`
- Write a function that returns a slice of numeronyms for its input strings:

```
func numeronyms(vals ...string) []string
```

Call `fmt.Println(numeronyms("accessibility", "Kubernetes", "abc"))` in `main` to print:

```
[a11y K8s abc]
```

mohitnag:
Fixes #409 

Review of colleague's PR #386

add numeronyms and its tests

### DIFF
--- a/04_numeronym/mohitnag/main.go
+++ b/04_numeronym/mohitnag/main.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
+	ns := make([]string, len(vals))
+	for i, val := range vals {
+		ns[i] = numeronym(val)
+	}
+	return ns
+}
+
+func numeronym(v string) string {
+	rs := []rune(v)
+	length := len(rs)
+	if length <= 3 {
+		return v
+	}
+	return fmt.Sprintf("%c%d%c", rs[0], length-2, rs[length-1])
+}
--- a/04_numeronym/mohitnag/main_test.go
+++ b/04_numeronym/mohitnag/main_test.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNumeronyms(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    []string
+		expected []string
+	}{
+		{"Happy Day", []string{"ehi!@  45%!"}, []string{"e9!"}},
+		{"Empty slice should return empty", []string{""}, []string{""}},
+		{"String with 2 char", []string{"we"}, []string{"we"}},
+		{"String with 3 char", []string{"win"}, []string{"win"}},
+	}
+	for _, td := range testData {
+		td := td
+		t.Run(td.Scenario, func(t *testing.T) {
+			numeronym := numeronyms(td.input...)
+			assert.Equal(t, td.expected, numeronym)
+		})
+	}
+}
+func TestMain(t *testing.T) {
+	assert := assert.New(t)
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := string("[a11y K8s abc]")
+	actual := buf.String()
+	actual = strings.Replace(actual, "\n", ",", -1)
+	actual = strings.TrimRight(actual, ",")
+	assert.Equal(expected, actual)
+}
+
+func TestNumeronym(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    string
+		expected string
+	}{
+		{"Three letters", "abc", "abc"},
+		{"Four letters", "test", "t2t"},
+		{"Test Unicode", "🤓😀🤣🥳", "🤓2🥳"},
+	}
+	for _, td := range testData {
+		td := td
+		t.Run(td.Scenario, func(t *testing.T) {
+			numeronym := numeronym(td.input)
+			assert.Equal(t, td.expected, numeronym)
+		})
+	}
+}


04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+		})
+	}
+}
+func TestMain(t *testing.T) {
+	assert := assert.New(t)
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := string("[a11y K8s abc]")
+	actual := buf.String()
```
alfredxiao, line 36: Since you have tested the logic in another Test function, main test can be made simpler to test a single one like `abc` or `abcd`

04_numeronym/mohitnag/main.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	return ns
+}
+
+func numeronym(v string) string {
+	length := len(v)
+	if length <= 3 {
+		return v
+	}
+	return fmt.Sprintf("%c%d%c", v[0], length-2, v[length-1])
```
alfredxiao: Nice! But there is a unicode issue, using v[0] you don't the the first unicode character, probably should be using rune slice like `rs := []rune(val)` and then use `rs[0]`

04_numeronym/mohitnag/main.go
```diff
@@ -0,0 +1,29 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
+	ns := []string{}
+	for _, val := range vals {
```
alfredxiao: size of `ns` is known and could potentially be preallocated

04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func TestNumeronym(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    string
+		expected string
+	}{
+		{"Scenario One", "abc", "abc"},
+		{"Scenario two", "test", "t2t"},
+	}
```
alfredxiao, line 51: Better add a test case for empty string

alfredxiao (changes requested):
need to support unicode character

04_numeronym/mohitnag/main_test.go
mohitnag: Good point. However this is in line with the requirements : ```Call `fmt.Println(numeronyms("accessibility", "Kubernetes", "abc"))` in `main` to print:

```
[a11y K8s abc]
```
So i would like to keep it as it is .

04_numeronym/mohitnag/main_test.go
mohitnag: yes i already have the empty string test ```{"Scenario two", []string{""}, []string{""}},```

04_numeronym/mohitnag/main.go
mohitnag: Done

alfredxiao (changes requested):
Please add test to cover unicode character

mohitnag:
added tests for unicode characters

alfredxiao (dismissed):
Looks good to me.

rokane:
First things first @mohitnag, can you please address the number of commits in this pull request. I can see that you have been merging master into this branch again. Please rebase onto master instead of merging in future and you will avoid ending up with a large number of commits. Please clean this up and aim for 1-2 commits for a pull request of this size

04_numeronym/mohitnag/main.go
```diff
@@ -0,0 +1,30 @@
...
+func numeronyms(vals ...string) []string {
+	ns := make([]string, len(vals))
+	for i, val := range vals {
+		ns[i] = numeronym(val)
+	}
+	return ns
+}
+
+func numeronym(v string) string {
+	rs := []rune(v)
```
rokane, line 24: You shouldn't need to create another `rs` variable so that you have a `[]rune`. You can access the individual runes from the `[]string`. Please remove this and directly access from `v`

04_numeronym/mohitnag/main.go
```diff
@@ -0,0 +1,30 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
+
+func numeronyms(vals ...string) []string {
```
rokane, line 15: Looking really good. Super clean!

04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNumeronyms(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    []string
+		expected []string
+	}{
+		{"Scenario One", []string{"ehi!@  45%!"}, []string{"e9!"}},
```
rokane: I think I have mentioned this on an earlier pull request of yours as well, but it is best to use descriptive explanations for your test scenario. Scenario One and Scenario Two are not descriptive and won't give much to anyone else who runs your tests and sees this on the output.

Please reword these so they are explaining what the test is doing.

Eg. Empty String. String with Length greater than 3.

On top of this I would like to see more test cases. Think about the edge cases here, what could possibly go wrong. The function is doing some different input on a string of length 3 vs a string of length >3. Perhaps you want to test around this line of len(3)?

04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+		{"Scenario two", []string{""}, []string{""}},
+	}
+	for _, td := range testData {
+		td := td
+		t.Run(td.Scenario, func(t *testing.T) {
+			numeronym := numeronyms(td.input...)
+			assert.Equal(t, td.expected, numeronym)
+			fmt.Println(numeronym)
```
rokane: Can you please remove this print statement, this is not required

04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+func TestNumeronym(t *testing.T) {
+	testData := []struct {
+		Scenario string
+		input    string
+		expected string
+	}{
+		{"Three letters", "abc", "abc"},
+		{"Four letters", "test", "t2t"},
+		{"Test Unicode", "\u00C0\u00C0\u00C0\u00C0", "\u00C02\u00C0"},
```
rokane: Here again, I think this was mentioned in another pull request, but it is easier to read if you use the actual character, not the unicode value.

Eg. `À` instead of `\u00C0`. Please fix these

04_numeronym/mohitnag/main_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	for _, td := range testData {
+		td := td
+		t.Run(td.Scenario, func(t *testing.T) {
+			numeronym := numeronym(td.input)
+			assert.Equal(t, td.expected, numeronym)
+			fmt.Println(numeronym)
```
rokane: Please remove this print statement as well

rokane (changes requested):
Well done, code is looking really good. I have a couple of suggestions throughout for cleaning up and improving readability. If you can address these and then slack me this should be good to be merged.

04_numeronym/mohitnag/main_test.go
mohitnag: Done

04_numeronym/mohitnag/main.go
mohitnag: 👍

04_numeronym/mohitnag/main_test.go
mohitnag: Done

04_numeronym/mohitnag/main_test.go
mohitnag: Done

04_numeronym/mohitnag/main_test.go
mohitnag: Done

04_numeronym/mohitnag/main.go
mohitnag: Ryan, i think i need to keep the `[]rune`  variable to support unicode.

rokane: True. I just tested this, I was under the impression that this still supported unicode. Thanks for pointing this out!

rokane (dismissed):
Well done mate. Great catch on the unicode capability as well!

anzboi (approved):
LGTM

PR #394, opened 2019-06-16
Lab 5 - Created IPAddr data type

joel00wood:
Overwrote `.String()` implementation to output underlying byte slice `{0,0,0,0}` as \'0.0.0.0\'

Fixes  #393

Review of colleague's PR #452

#### Changes proposed in this PR:
- IPAddr data type
- Overwrites underlying `.String()` to format output as an IP v4 Address

PR #422, opened 2019-06-17
complete lab08

### LINKED ISSUE #421: lab08 :new_moon:
solve lab08

n0npax:
Refactor lab07 to fit project layout

Fixes #421 

Review of colleague's PR #394

## Complete lab08
move lab07 and refactor to proper layout


n0npax:
due to GolangCI

Julia Ogris Commented on 18 Jun [12:04]
ignore golangci-bot if travis passes (uese go 1.12). ... feel free to update `.golanci.yaml` if you want in a separate pr.

08_project/n0npax/pkg/n0npax-puppy/types_test.go
```diff
@@ -0,0 +1,33 @@
...
+		"empty": {p: Puppy{}, s: "{\"id\":0,\"value\":0,\"breed\":\"\",\"colour\":\"\"}"},
+	}
+	for name, test := range testCases {
+		input, expected := test.p, test.s
+		t.Run(name, func(t *testing.T) {
+			assert := assertion.New(t)
+			puppyByte, _ := json.Marshal(&input)
+			assert.Equalf(expected, string(puppyByte), "Marshaling test failed")
+			var recoveredPuppy Puppy
+			json.Unmarshal(puppyByte, &recoveredPuppy)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

08_project/n0npax/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,20 @@
...
+
+	puppy "github.com/anz-bank/go-course/08_project/n0npax/pkg/puppy"
+	store "github.com/anz-bank/go-course/08_project/n0npax/pkg/puppy/store"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	store := store.MemStore{}
+	p := puppy.Puppy{ID: 1, Breed: "Mix", Colour: "White", Value: 100}
```
juliaogris: Don't set ID - it should be filled in Create function IMO.

08_project/n0npax/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+func TestMain(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+	expected := "\"Puppy id: 0\""
+	actual := strconv.Quote(buf.String())
+	if expected != actual {
+		t.Errorf("Unexpected output\nexpected: %v\nactual: %v", expected, actual)
```
juliaogris: I'd start using testify - it collapses all these three lines into a single line with `assert.Equal(expected, actual, optionalMessage)`

08_project/n0npax/pkg/puppy/errors.go
```diff
@@ -0,0 +1,54 @@
+package puppy
+
+import (
+	"fmt"
+)
+
+// Error codes
+const (
+	ErrInvalidInputCode  = 422
```
juliaogris: If you stick with HTTP codes, I'd use 400 - Bad Request here

This is how I usually like to do it:
```go
// ErrCode defines the kind of error
type ErrCode uint8

// Error codes
const (
	// ErrInvalidInput for invalid user input
	ErrInvalidInput ErrCode = iota
	// ErrNotFound is used when trying to access a non-existing entry
	ErrNotFound 
        // etc.
)
```

08_project/n0npax/pkg/puppy/store/type.go
```diff
@@ -0,0 +1,26 @@
+package store
```
juliaogris: IMO this file should not exist. The types should be declared in the files where their methods are declared.
I think it's preferable to not split type and method definitions over several files.

08_project/n0npax/pkg/puppy/store/syncStore.go
```diff
@@ -0,0 +1,62 @@
...
+		p := puppyData.(puppy.Puppy)
+		return &p, nil
+	}
+	return nil, puppy.ErrNotFound(fmt.Sprintf(puppy.PuppyNotFoundMsg, id))
+}
+
+// UpdatePuppy updates puppy
+func (s *SyncStore) UpdatePuppy(id int, p *puppy.Puppy) error {
+	if p.Value < 0 {
+		return puppy.ErrInvalidInput(puppy.InvalidInputMsg)
```
juliaogris: This strikes me as a bit strange - I'd provide the error message here and not go with constants, but I can't really argue well why. My usual approach is to create function akin to this:
```go
func Errorf(code ErrCode, format string, args ...interface{}) *Error
```
I usually declare this one Error method and provide code on creation. It's also useful to mention calling context in the message IMO.
So I'd do something like:

```go
return Errorf(InvalidInput, "UpdatePuppy: puppy value (%d) must not be negative", p.Val)  
```

08_project/n0npax/pkg/puppy/errors.go
```diff
@@ -0,0 +1,54 @@
+package puppy
```
juliaogris, line 1: IMO I see people "over-structuring" go projects quite often (i.e. too many, too fine grained packages).
I'd keep everything in a single package.
Saves you prefixing errors with `puppy.`.

08_project/n0npax/pkg/puppy/store/type.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+// MemStore map based type for storing puppies data
+type MemStore map[int]puppy.Puppy
+
+// LevelDBStore provides sync for leveldb
+type LevelDBStore struct {
+	ldb *leveldb.DB
+	sync.Mutex
+	total int
```
juliaogris: This is a bit of a race condition. What if somebody else adds a puppy to your DB in another way?
I've had a bit of a look at the LevelDB API - not sure if there is a better solution - this might just be the best we can do....
Aside: total is a bit of a misleading name as it is not the total of all entries once you start deleting.
Maybe call it nextId or similar?

juliaogris (changes requested):
Super solid, as before.
I have only a bunch of suggestions, no strong opinions on things needing changing here :)
(Please make your commit message start with a capital letter.)

juliaogris (commented):
oops, looks like I started a review in a separate tab. Apologies....

08_project/n0npax/cmd/puppy-server/main.go
n0npax: Fixed. Thanks!

08_project/n0npax/cmd/puppy-server/main_test.go
n0npax: fixed.

08_project/n0npax/pkg/puppy/errors.go
n0npax: Replaced `422` with `400`.
I'd prefer to stay with `http` codes. Thanks for example with `iota`

08_project/n0npax/pkg/puppy/store/type.go
n0npax: Your totally right, thanks for this. fixed.

08_project/n0npax/pkg/puppy/store/type.go
n0npax: `total` renamed to `nextID`.

I believe 1st part of the comment is overkill.
If someone is adding a puppy to any store in another way than API, there will be always race condition.
Let's say we're working with `mysql`. Someone can be connected directly to the InnoDB engine and skip most of the `mysql` synchronization. Besides this, most of the client libraries assume nobody modifies persistent layer in a hacky way in the midtime.

08_project/n0npax/pkg/puppy/errors.go
n0npax: I may be misunderstanding sth here.

* Due to: https://github.com/anz-bank/go-course/tree/master/08_project I should keep storer implementations in `store` directory.
* Due to https://golang.org/ref/spec#Package_clause `go` may require keeping package files in a flat directory.
`A set of files sharing the same PackageName form the implementation of a package. An implementation may require that all source files for a package inhabit the same directory.`

Should I:
* leave packages as is?
* modify package structure?
* read about something I'm missing?

08_project/n0npax/pkg/puppy/store/syncStore.go
n0npax: I moved error messages to the requested format.
I provided them as const to avoid running `sed` on the source code.
`Errorf` function refactored. `args ...interface{}` looks awesome, Thanks!

n0npax:
@juliaogris almost everything fixed. I'm not sure about 2 comments, so they're not marked as resolved.

08_project/n0npax/pkg/puppy/errors.go
juliaogris: No you are right - who came up with this lab (oh yeah, that was me....) Sorry!
Carry on - ie leave package as is.
I guess the point was to make use of different packages, I should have double checked myself first.

08_project/n0npax/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+	expected := "\"Puppy id: 0\""
```
juliaogris, line 16: Prefer raw strings when character escaping is needed for readability.
Sorry I missed that in the first round feel free to leave it:
```go
expected := `"Puppy id: 0"`
```

juliaogris (approved):
Great Work!!
LGTM

n0npax:
adding label `lab11` to test if leaderboard can be tricked

PR #423, opened 2019-06-17
Add Fibonacci sequence generator

### LINKED ISSUE #416: Lab 01 - Fibonacci
# Lab 1 - Fibonacci

- Create an executable go program in directory `01_fib/USERNAME`
- Write a function that prints the first n Fibonacci numbers

```
func fib(n int)
```

Call `fib(7)` in `main` to print

```
1
1
2
3
5
8
13
```

Bonus points: For negative n print Negafibonacci numbers.

patrickmarabeas:
Implements a function which creates a slice of n ints of the Fibonacci sequence given an input of int n. A fib() function is also supplied which handles the formatting of output - whether it is an error or a new line separated output of the generated sequence.

Features:
- Starts from 0
- Generates both positive and negative Fibonacci sequences
- Returns an error if the generated sequence would result in ints
outside the lower and upper bounds

Fixes #416 

Review of colleague's PR #451

### DIFF
--- a/01_fib/patrickmarabeas/main.go
+++ b/01_fib/patrickmarabeas/main.go
@@ -1,0 +1,63 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// calculateFibSequence creates a slice of Fibonacci numbers
+// Starts from 0
+// Negative integer inputs will create a Negafibonacci sequence
+// Returns nil for input of 0
+// Returns an error if the input would generate numbers outside of an int lower/upper values
+func calculateFibSequence(n int) ([]int, error) {
+	// Early bailout returning zero value of a slice
+	if n == 0 {
+		return nil, nil
+	}
+	// Early bailout and error for return values outside of int limits
+	if n > 92 || n < -92 {
+		return nil, fmt.Errorf("fib: %d is outside of limits of -92..92", n)
+	}
+	// Handle negative inputs
+	// Multiplied against the initial Fibonacci value (1) to spawn either a positive or negative sequence
+	sign := 1
+	if n < 0 {
+		n *= -1
+		sign = -1
+	}
+
+	result := make([]int, n)
+	// Seed initial value for the sequence
+	result[0] = 0
+	if n > 1 {
+		// Seed second value for the sequence; accommodating for Negafibonacci
+		result[1] = 1 * sign
+		// Calculate remaining sequence
+		for i := 2; i < n; i++ {
+			result[i] = result[i-2] + result[i-1]*sign
+		}
+	}
+	return result, nil
+}
+
+// fib prints either a loop of Fibonacci numbers or an error
+func fib(n int) {
+	seq, err := calculateFibSequence(n)
+
+	if err != nil {
+		fmt.Fprint(out, err)
+		return
+	}
+
+	for _, num := range seq {
+		fmt.Fprintln(out, num)
+	}
+}
+
+func main() {
+	fib(7)
+}
--- a/01_fib/patrickmarabeas/main_test.go
+++ b/01_fib/patrickmarabeas/main_test.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"reflect"
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := fmt.Sprintf("%s\n", "0\n1\n1\n2\n3\n5\n8")
+	got := buf.String()
+
+	t.Run("Main function", func(t *testing.T) {
+		if expected != got {
+			t.Errorf("\nExpected: %s\nGot:      %s", expected, got)
+		}
+	})
+}
+
+func TestFibError(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	fib(100)
+
+	expected := "fib: 100 is outside of limits of -92..92"
+	got := buf.String()
+
+	t.Run("Main function", func(t *testing.T) {
+		if expected != got {
+			t.Errorf("\nExpected: %s\nGot:      %s", expected, got)
+		}
+	})
+}
+
+func TestCalculateFibSequence(t *testing.T) {
+	var cases = map[string]struct {
+		input     int
+		expected1 []int
+		expected2 error
+	}{
+		"Zero": {
+			input:     0,
+			expected1: nil,
+			expected2: nil,
+		},
+		"Input of 1": {
+			input:     1,
+			expected1: []int{0},
+			expected2: nil,
+		},
+		"Input of negative 1": {
+			input:     -1,
+			expected1: []int{0},
+			expected2: nil,
+		},
+		"Input of 3": {
+			input:     3,
+			expected1: []int{0, 1, 1},
+			expected2: nil,
+		},
+		"Input of 7": {
+			input:     7,
+			expected1: []int{0, 1, 1, 2, 3, 5, 8},
+			expected2: nil,
+		},
+		"Input of negative 3": {
+			input:     -3,
+			expected1: []int{0, -1, 1},
+			expected2: nil,
+		},
+		"Input of negative 7": {
+			input:     -7,
+			expected1: []int{0, -1, 1, -2, 3, -5, 8},
+			expected2: nil,
+		},
+		"Out of range - positive": {
+			input:     100,
+			expected1: nil,
+			expected2: errors.New("fib: 100 is outside of limits of -92..92"),
+		},
+		"Out of range - negative": {
+			input:     -100,
+			expected1: nil,
+			expected2: errors.New("fib: -100 is outside of limits of -92..92"),
+		},
+	}
+
+	for name, c := range cases {
+		c := c
+		t.Run(name, func(t *testing.T) {
+			got1, got2 := calculateFibSequence(c.input)
+			expected1 := c.expected1
+			expected2 := c.expected2
+			if !reflect.DeepEqual(got1, expected1) || !reflect.DeepEqual(got2, expected2) {
+				t.Errorf("\nExpected: %d, %s\nGot:      %d, %s", expected1, expected2, got1, got2)
+			}
+		})
+	}
+}


01_fib/patrickmarabeas/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := fmt.Sprintf("%s\n", "0\n1\n1\n2\n3\n5\n8\nERROR: 100 is outside of limits Please enter a number between -92 and 92")
```
golangcibot: line is 129 characters (from `lll`)

01_fib/patrickmarabeas/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+	got := fmt.Sprintf("%s", buf.String())
```
golangcibot: S1025: the argument is already a string, there's no need to use fmt.Sprintf (from `gosimple`)

juliaogris (dismissed):
LGTM.
Above and beyond :)
I like how you just decided: The fibonacci series starts at 0 (which really it does), even though the lab asked for it to start at 1.

juliaogris:
If you want to `rebase` rather than `merge` master in with "Update branch", feel free to force push. I'll reapprove right away.

patrickmarabeas:
Thanks, will definitely do some peer reviews ASAP as well.

juliaogris (dismissed):
lgtm

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// Multiplied against the initial Fibonacci value (1) to spawn either a positive or negative sequence
+	sign := 1
+	// Early bailout returning zero value of a slice
+	if n == 0 {
+		return nil, nil
+	}
+	// Early bailout and error for return values outside of int limits
+	if n > 92 || n < -92 {
+		return nil,
+			fmt.Errorf(fmt.Sprintf(`ERROR: %d is outside of limits Please enter a number between -92 and 92`, n*sign))
```
camscale: The `fmt.Sprintf` is unnecessary. `fmt.Errorf` does that already (the `f` suffix denotes a formatted string)

patrickmarabeas: Resolved.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// Multiplied against the initial Fibonacci value (1) to spawn either a positive or negative sequence
+	sign := 1
+	// Early bailout returning zero value of a slice
+	if n == 0 {
+		return nil, nil
+	}
+	// Early bailout and error for return values outside of int limits
+	if n > 92 || n < -92 {
+		return nil,
+			fmt.Errorf(`ERROR: %d is outside of limits Please enter a number between -92 and 92`, n*sign)
```
camscale: You should not start the error string with `ERROR:` as it is common for error messages to be wrapped in higher layers. This would put `ERROR:` in the middle of the string. Also, no need to say "Please" (also should not be capitalised). Just put the error message. I suggest: "fib: argument %d outside limits of -92..92" or something.

camscale: No need to use a raw string here. Just normal double quotes are fine.

camscale: The argument `n*sign` kind of pointless, as `sign` at this point is `1`.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+var out io.Writer = os.Stdout
+
+// Fibonacci slice generator
+// - Starts from 0
+// - Handles both positive and negative numbers
+// - Returns nil for input of 0
+// - Returns an error if the input would generate numbers outside of an int lower/upper values
+func calculateFibSequence(n int) ([]int, error) {
+	// Multiplied against the initial Fibonacci value (1) to spawn either a positive or negative sequence
+	sign := 1
```
camscale: I would move this to where you check negative values and set the sign to -1. As it stands, setting the value of `sign` is split over multiple lines.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil,
+			fmt.Errorf(`ERROR: %d is outside of limits Please enter a number between -92 and 92`, n*sign)
+	}
+	// Handle negative inputs
+	if n < 0 {
+		n *= -1
+		sign = -1
+	}
+	// Return a single length slice of zero value int
+	if n <= 1 {
```
camscale: This case is really `n == 1` because you've already handled the zero case earlier. You should make that your condition.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+		return make([]int, 1), nil
+	}
+	// Make a splice of required length
+	result := make([]int, n)
```
camscale, line 33: Move this above the `n == 1` condition and you don't need two `make`s. The `n == 1` case can just be `return result, nil`

01_fib/patrickmarabeas/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+			expected2: nil,
+		},
+		"Input of 7": {
+			input:     3,
+			expected1: []int{0, 1, 1, 2, 3, 5, 8},
+			expected2: nil,
+		},
+		"Input of negative 3": {
+			input:     -3,
+			expected1: []int{0, -1, -1},
```
camscale: This is not the correct sequence for negative fibonacci numbers. See https://en.wikipedia.org/wiki/Fibonacci_number#Negafibonacci

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// Seed initial two values of the Fibonacci sequence; accommodating for the negative sequence
+	result[0] = 0
+	result[1] = 1 * sign
+	// Calculate remaining sequence
+	for i := 2; i < n; i++ {
+		result[i] = result[i-2] + result[i-1]
```
camscale: This doesn't correctly calculate the negative sequence. See https://en.wikipedia.org/wiki/Fibonacci_number#Negafibonacci for details. Same comment made in the test harness.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+	// Handle negative inputs
+	if n < 0 {
+		n *= -1
+		sign = -1
+	}
+	// Return a single length slice of zero value int
+	if n <= 1 {
+		return make([]int, 1), nil
+	}
+	// Make a splice of required length
```
camscale: This comment adds no value. the code says that on its own

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+	return result, nil
+}
+
+// Fibonacci IO
+func fib(n int) {
+	seq, err := calculateFibSequence(n)
+	if err != nil {
+		fmt.Fprintln(out, err)
+	} else {
```
camscale: Add a return to the previous case and get rid of the else: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+		for _, num := range seq {
+			fmt.Fprintln(out, num)
+		}
+	}
+}
+
+func main() {
+	fib(7)
+	fib(100)
```
camscale: Technically the lab asks for the output of `fib(7)` only. I don't care all that much though, but if you are going to deviate from a spec, best to add a comment as to why so we know it's deliberate and not an accidental deviation.

01_fib/patrickmarabeas/main_test.go
```diff
@@ -0,0 +1,91 @@
...
+			input:     -100,
+			expected1: nil,
+			expected2: errors.New("ERROR: -100 is outside of limits Please enter a number between -92 and 92"),
+		},
+	}
+
+	for name, c := range cases {
+		got1, got2 := calculateFibSequence(c.input)
+		expected1 := c.expected1
+		expected2 := c.expected2
```
camscale: These three lines belong in the test that is run by the runner. It keeps the test case together. You'll probably need a `c := c` in the loop before the `t.Run` call, due to the loop variable being re-used and `t.Run` running concurrently.

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// Fibonacci slice generator
```
camscale: This is not the correct form for a doc string: https://blog.golang.org/godoc-documenting-go-code

camscale (changes requested):
Just a couple more things to fix up.

01_fib/patrickmarabeas/main.go
patrickmarabeas: Resolved

patrickmarabeas: Resolved

patrickmarabeas: Good point

01_fib/patrickmarabeas/main.go
patrickmarabeas: Resolved

01_fib/patrickmarabeas/main.go
```diff
@@ -0,0 +1,63 @@
...
+
+	result := make([]int, n)
+	// Seed initial value for the sequence
+	result[0] = 0
+	if n > 1 {
+		// Seed second value for the sequence; accommodating for Negafibonacci
+		result[1] = 1 * sign
+		// Calculate remaining sequence
+		for i := 2; i < n; i++ {
+			result[i] = result[i-2] + result[i-1] * sign
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
			result[i] = result[i-2] + result[i-1]*sign
```

01_fib/patrickmarabeas/main.go
patrickmarabeas: Reworked the conditions in a similar manner

01_fib/patrickmarabeas/main.go
patrickmarabeas: More for personal documentation while learning; but it does add no value now.

01_fib/patrickmarabeas/main.go
patrickmarabeas: Resolved

01_fib/patrickmarabeas/main_test.go
patrickmarabeas: I must admit, I assumed what that meant rather than reading. Corrected.

01_fib/patrickmarabeas/main.go
patrickmarabeas: Resolved

01_fib/patrickmarabeas/main.go
patrickmarabeas: Agreed; thanks for link - wasn't aware it had a name

01_fib/patrickmarabeas/main.go
patrickmarabeas: Resolved; added test case instead.

01_fib/patrickmarabeas/main_test.go
patrickmarabeas: Resolved

01_fib/patrickmarabeas/main.go
patrickmarabeas: Hopefully this has been corrected...

camscale (approved):
LGTM. That's looking great now.

PR #424, opened 2019-06-18
Add bubble and insertion sort algorithms

### LINKED ISSUE #417: Lab 02 - Bubble Sort
# Lab 2 - Bubble sort

- Create an executable go program in directory `02_bubble/USERNAME`
- Write a function that returns a sorted copy of `int` slice `s` using [bubble sort](https://en.wikipedia.org/wiki/Bubble_sort):

```
func bubble(s []int) []int
```

```
[1 2 3 5]
```

- Bonus points: implement [Insertion sort](https://en.wikipedia.org/wiki/Insertion_sort)
- Extra bonus points: implement an _O(n_ _log(n))_ sorting algorithm

patrickmarabeas:
Implements the bubble and insertion sort algorithms. Both functions take a slice of n ints and return a sorted slice of n ints using specified sorting algorithm.

- Adds bubble sort algorithm
- Adds insertion sort algorithm

Fixes #417 

Review of colleague's PR #501

### DIFF
--- a/02_bubble/patrickmarabeas/main.go
+++ b/02_bubble/patrickmarabeas/main.go
@@ -31,6 +31,29 @@ func bubble(s []int) []int {
 	return result
 }
 
+// Insertion implements the Insertion sort algorithm.
+func insertion(s []int) []int {
+	length := len(s)
+	// Create copy
+	result := make([]int, length)
+	copy(result, s)
+	// Increment through the slice
+	// All but first number need to be checked
+	for i := 1; i < length; i++ {
+		// Increment backwards through slice
+		// Start with first unsorted index
+		for j := i; j > 0; j-- {
+			// Swap unsorted values
+			if result[j] < result[j-1] {
+				result[j], result[j-1] = result[j-1], result[j]
+			}
+		}
+	}
+
+	return result
+}
+
 func main() {
 	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+	fmt.Fprint(out, insertion([]int{3, 2, 1, 5}))
 }
--- a/02_bubble/patrickmarabeas/main.go
+++ b/02_bubble/patrickmarabeas/main.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// Bubble implements the Bubble sort algorithm.
+func bubble(s []int) []int {
+	length := len(s)
+	// Create copy
+	result := make([]int, length)
+	copy(result, s)
+	// Increment through the slice
+	// All but last number need to be checked
+	for i := 1; i < length; i++ {
+		// Increment through the slice
+		// Compare current index with next
+		// Don't check previous sorted indices
+		for j := 0; j < length-i; j++ {
+			// Swap unsorted values
+			if result[j] > result[j+1] {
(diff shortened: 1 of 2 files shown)


Marchitect:
Looks all good to me, I think your solutions look neat, good work. One suggestion I can give would be that you might consider adding negative integers into test cases.

patrickmarabeas:
@Marchitect - Added negative integer tests.

undrewb (dismissed):
Nice neat implementation. Plenty of test cases. Good work!

02_bubble/patrickmarabeas/main.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+// Bubble implements the Bubble sort algorithm
+func bubble(s []int) []int {
+	length := len(s)
+	// Create copy
+	result := make([]int, length)
+	copy(result, s)
```
anzdaddy, line 16: This forces a copy that the caller may not require. Making a copy should be left up to the caller.

patrickmarabeas: This is a good point, thanks. I'll make sure I'm not being opinionated re immutability in the future.

02_bubble/patrickmarabeas/main.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// Bubble implements the Bubble sort algorithm.
+func bubble(s []int) []int {
```
anzdaddy, line 12: nit: With the change in mutability behaviour, it would be clearer to not return anything.

anzdaddy (dismissed):

02_bubble/patrickmarabeas/main.go
patrickmarabeas: ~Would this result in the following signature `func bubble(s *[]int)`?~

I see Slices are passed by reference already.

patrickmarabeas: Added alternative implementation and tests

02_bubble/patrickmarabeas/main_test.go
```diff
@@ -0,0 +1,92 @@
...
+				t.Errorf("\nExpected: %d\nGot:      %d", expected, got)
+			}
+		})
+	}
+}
+
+func TestBubble2(t *testing.T) {
+	for name, c := range cases {
+		got := c.input
+		bubble2(got)
```
anzdaddy: Don't forget that you're now mutating the original and thus might require a copy.

juliaogris (commented):
Please address [failing build]().  
From the [lab assignment doc](https://github.com/anz-bank/go-course/blob/master/02_bubble/README.md):
> Write a function that returns a sorted *copy* 
(my emphasis).  

You are mutating the input.

02_bubble/patrickmarabeas/main_test.go
patrickmarabeas: Reverted to returning a copy, as per spec

patrickmarabeas:
Reverted to original implementation

juliaogris (approved):
LGTM.

Just a minor nit on style, don't comment the obvious, that just creates noise:

```go
	// Create copy
	result := make([]int, length)
	copy(result, s)
```

PR #427, opened 2019-06-18
Add IPAddr type

patrickmarabeas:
Implements fmt.Stringer

Fixes #420

Review of colleague's PR #534

PR #429, opened 2019-06-18
Feature/lab 1

### LINKED ISSUE #428: Lab 1 - Implement Fib Function
Need to implement Fib function that accepts both positive and negative input and output the right fib sequence.

starfallz:
Fixes #428 

Review of colleague's PR #427 

- My very first PR for go yay! Fib function that takes both positive / negative values with tests included.

### DIFF
--- a/01_fib/starfallz/main.go
+++ b/01_fib/starfallz/main.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, strings.Trim(strings.Join(strings.Split(fmt.Sprint(fib(7)), " "), "\n"), "[]"))
+}
+
+func fib(n int) []int {
+	if n == 0 {
+		return []int{0}
+	} else {
+		isNegativeFib := false
+
+		if n < 0 {
+			isNegativeFib = true
+			n = n * -1
+		}
+
+		result := []int{1}
+		for i := 1; i < n; i++ {
+			var previousValue int
+			if len(result) == 1 {
+				previousValue = 0
+			} else {
+				previousValue = result[len(result)-2]
+			}
+
+			if isNegativeFib {
+				result = append(result, previousValue-result[i-1])
+			} else {
+				result = append(result, result[i-1]+previousValue)
+			}
+		}
+		return result
+	}
+}
--- a/01_fib/starfallz/main.go
+++ b/01_fib/starfallz/main.go
@@ -16,29 +16,29 @@ func main() {
 func fib(n int) []int {
 	if n == 0 {
 		return []int{0}
-	} else {
-		isNegativeFib := false
-
-		if n < 0 {
-			isNegativeFib = true
-			n = n * -1
-		}
-
-		result := []int{1}
-		for i := 1; i < n; i++ {
-			var previousValue int
-			if len(result) == 1 {
-				previousValue = 0
-			} else {
-				previousValue = result[len(result)-2]
-			}
-
-			if isNegativeFib {
-				result = append(result, previousValue-result[i-1])
-			} else {
-				result = append(result, result[i-1]+previousValue)
-			}
-		}
-		return result
 	}
+
+	isNegativeFib := false
+
+	if n < 0 {
+		isNegativeFib = true
+		n *= -1
+	}
+
+	result := []int{1}
+	for i := 1; i < n; i++ {
+		var previousValue int
+		if len(result) == 1 {
+			previousValue = 0
+		} else {
+			previousValue = result[len(result)-2]
+		}
+
+		if isNegativeFib {
+			result = append(result, previousValue-result[i-1])
+		} else {
+			result = append(result, result[i-1]+previousValue)
+		}
+	}
+	return result
 }
--- a/01_fib/starfallz/main.go
+++ b/01_fib/starfallz/main.go
@@ -4,41 +4,31 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"strings"
 )
 
 var out io.Writer = os.Stdout
 
 func main() {
-	fmt.Fprint(out, strings.Trim(strings.Join(strings.Split(fmt.Sprint(fib(7)), " "), "\n"), "[]"))
+	printIntSlice(fib(7))
+}
+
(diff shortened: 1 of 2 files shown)


01_fib/starfallz/main.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, strings.Trim(strings.Join(strings.Split(fmt.Sprint(fib(7)), " "), "\n"), "[]"))
```
juliaogris: That's a novel approach :)

However, the lab requirement was to write a function with the following signature:  
`func fib(n int)` 
happy to let this go.

You could pull the `Split` and `Join` together with a `strings.ReplaceAll`.
I agree it's nice that you don't print inside `fib`, however, I'd wrap the printing in a `printIntSlice` function or similar, that iterates through the `fib`'s result array `Println`s each element. 
In order to understand your approach the reader of your code has to have intimate knowledge with the way golang prints slices.

01_fib/starfallz/main.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	isNegativeFib := false
+
+	if n < 0 {
+		isNegativeFib = true
+		n *= -1
+	}
+
+	result := []int{1}
```
juliaogris: `result := make([]int, n)` - you already know how big you need your slice to be.
slightly shorter:
```go
	sign := 1
	if n < 0 {
		sign = -1
	}

	n1, n2 := 0, 1
	result := make([]int, n*sign)
	for i := 0; i < n*sign; i++ {
		n1, n2 = n2, n1+sign*n2
		result[i] = n1
	}
	return result
}
```

if you print inside the function as required you should get a shortened:

```go
	sign := 1
	if n < 0 {
		sign = -1
	}

	n1, n2 := 0, 1
	for j := 0; j < n*sign; j++ {
		n1, n2 = n2, n1+sign*n2
		fmt.Fprintln(out, n1)
	}
```

01_fib/starfallz/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+	})
+}
+
+func TestFibFunction(t *testing.T) {
+	testCases := []struct {
+		description string
+		input       int
+		expected    []int
+	}{
+		{"Test zero fib to return zero", 0, []int{0}},
```
juliaogris: Arguably this shouldn't print anything. Up to you.

01_fib/starfallz/main_test.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	for _, testCase := range testCases {
+		input := testCase.input
+		expected := testCase.expected
+
+		t.Run(testCase.description, func(t *testing.T) {
+			actual := fib(input)
+
+			if !reflect.DeepEqual(actual, expected) {
```
juliaogris, line 90: Might be worth while looking into `testify/assert` and `assert.Equal`, it pulls these three lines into one and prints nice diff in case of failure.

juliaogris (changes requested):
Please [fix commit messages](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages) with `git rebase -i`

01_fib/starfallz/main.go
starfallz: Yeah initially I did that however it felt kind of wierd to do many things in one function. I will implement the pritntIntSlice func instead since I want to be able to test them separately and not do too much in one func

01_fib/starfallz/main.go
starfallz: Haha you are literally giving me the solution Julia 😂😂😂😂 I agree it's much shorter but should I be copy pasting your code? Lol feels like it's a completely new implementation haha

01_fib/starfallz/main_test.go
starfallz: Ok i will return empty array for this instead of 0

01_fib/starfallz/main_test.go
starfallz: Initially I wanted to explore testify but I looked at lab6 and that's where it specifically ask us to use the assert functions. So I kept it plain vanilla go so I know how to do it without external dependency. But I'm happy to modify it to assert equal if it's a requirement

juliaogris: ok.

01_fib/starfallz/main.go
juliaogris: I know - can't help myself ;)

**TL;DR: You don't need to update anything (but you certainly can if you want ;) ).**

Some of us running this course have spent a bit of time going back and forth between us on what we think the canonical solution should be, and I always find it interesting to see how different the same short piece of functionality can look.  

Personally, I love it when others look at my code and help me make it shorter, nicer and/or more readable. Everything that makes it better makes _me_  happy: removing little stumbling blocks like typos, bad variable names, going against conventions and idioms or poor naming. Even more important I really appreciate it when medium size weirdnesses on a block/logic level are clarified or even bigger poor design choices are brought to light. 

To me it's not just about getting the job/ticket done but to have fun. To achieve this I see my job as a craft that wants to express itself as beautifully as it can and it almost always becomes better with many eyes looking over it. That's why I posted it :)

I guess that was the long answer.

01_fib/starfallz/main.go
juliaogris: Sounds good!

juliaogris (approved):
Code LGTM.
Usually I'd ask for a git commit clean up, but I don't want to block you any more.
Please have a read through the [git commit message standards](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages) we are asking for.
It's invaluable to learn the usage of `git rebase -i` to re-write and clean up you commit history. If you do want to fix up your commits before merge, please reach out on Slack and I will re-approve right away.

starfallz:
@juliaogris i tried rebasing via sourcetree, (kinda screwed things up) then i tried manually, kinda worked, but my biggest issue is my initial commit, that cant be changed so i gave up hahaha, if you are in the office can u show me how to reset the initial commit msg? i heard i can use --root but i havent actually tried it.

juliaogris:

PR #432, opened 2019-06-18
Pin golanci-lint-version to 1.17.1

n0npax:
pin golang-lint-version to 1.17.1 to keep tools up to date.

### DIFF
--- a/03_letters/n0npax/main_test.go
+++ b/03_letters/n0npax/main_test.go
@@ -103,7 +103,7 @@ func helperTestSortLetters(name string, input map[rune]int, expected []string) f
 				t.Errorf("Test: %s - Expect: %s got: %s", name, expected[i], v)
 			}
 		}
-		assert.Equalf(t, len(expected), len(result), "Lenght missmatch")
+		assert.Equalf(t, len(expected), len(result), "Length mismatch")
 	}
 }
 
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,5 +1,5 @@
 service:
-  golangci-lint-version: 1.16.0
+  golangci-lint-version: 1.17.1
 linters:
   enable-all: true
   disable:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ env:
 go:
   - "1.12.x"
 before_script:
-  - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -x -- -b $(go env GOPATH)/bin v1.16.0
+  - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -x -- -b $(go env GOPATH)/bin v1.17.1
 script:
   - golangci-lint run
   - go test -v -race -coverprofile=coverage.out -covermode=atomic ./...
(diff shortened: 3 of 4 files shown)


juliaogris (changes requested):
Thank you very much for doing this!

Could you please tell us in your commit message (and/or) PR description - most importantly in the merge commit _why_ you are doing it?

Is this an attempt to fix the golangci.com service?  
Is something else broken?
Is this just to keep things up-to-date?

Could you please also update the following line:  
https://github.com/anz-bank/go-course/blob/master/.travis.yml#L7

Thanks again for doing this - you really are a tremendous help to our efforts!

03_letters/n0npax/main_test.go
```diff
@@ -103,7 +103,7 @@ func helperTestSortLetters(name string, input map[rune]int, expected []string) f
 				t.Errorf("Test: %s - Expect: %s got: %s", name, expected[i], v)
 			}
 		}
-		assert.Equalf(t, len(expected), len(result), "Lenght missmatch")
+		assert.Equalf(t, len(expected), len(result), "Length missmatch")
```
n0npax: @juliaogris this line was changed due to: https://travis-ci.com/anz-bank/go-course/builds/116097682
seems new golanci-lint checks misspellings in strings :open_mouth:
If you prefer I can split this into 2 PR

n0npax:
To tell the truth, I tried to fix the CI, but the problem is the `go` version in `golanci` as you mentioned earlier.
I believe keeping software up to date is good practice, so I updated PR description with this info.

03_letters/n0npax/main_test.go
anzdaddy: Did it catch "missmatch"?

n0npax: Yes. Link refers to previous build without modification of this particular file.

juliaogris (dismissed):
LGTM

n0npax:
Need to fix misspellings in master

n0npax:
@juliaogris  @camh-anz could you approve once again?

03_letters/n0npax/main_test.go
camscale: I think what @anzdaddy is saying is that the word "missmatch" is also misspelt (unless it is a reference to the Swedish electronica band: https://en.wikipedia.org/wiki/MissMatch). It should be "mismatch". Even though golangci-lint is not picking this up, since you are fixing that line anyway, you may as well fix that other misspelling.

camscale (changes requested):
You should reverse the order of the commits in this PR. Ideally the tests would run on every commit and if it did, it would fail after the first commit because the misspellings are still present when you update the version.

Also please expand on the commit message and say why you're upgrading it. If it is just to get the latest version, then say that. If it is to resolve a specific problem, please describe that problem.

03_letters/n0npax/main_test.go
n0npax: awww I see now. Will be fixed soon

n0npax:
@juliaogris could you re-approve please?

camscale (approved):
LGTM. Sorry for not noticing this was ready for re-review. The notification got lost in the 200 or so other ones

PR #433, opened 2019-06-19
Implement 03_letters

### LINKED ISSUE #353: Lab 3 - 03_letters - Letter frequency
Build lab 3 assignment code for finding letter frequency in a string and output as a map with sorted strings.

saurabhvyas9:
Fixes #353 

Review of colleague's PR #212 

Changes proposed in this PR:
Lab3 - 03_letters assignment committed as part of the 3rd assignment for go-course.

### DIFF
--- a/03_letters/saurabhvyas9/main.go
+++ b/03_letters/saurabhvyas9/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	frequencyLetters := make(map[rune]int)
+	for _, str := range s {
+		frequencyLetters[str]++
+	}
+	return frequencyLetters
+}
+
+func sortLetters(m map[rune]int) []string {
+	sortedLetters := make([]string, len(m))
+	tempKey := make([]int, len(m))
+	i := 0
+	for key := range m {
+		tempKey[i] = int(key)
+		i++
+	}
+	sort.Ints(tempKey)
+
+	for index, runeKey := range tempKey {
+		sortedLetters[index] = fmt.Sprintf("%c:%d", runeKey, m[rune(runeKey)])
+	}
+	return sortedLetters
+}
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
--- a/03_letters/saurabhvyas9/main_test.go
+++ b/03_letters/saurabhvyas9/main_test.go
@@ -1,0 +1,93 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	// Given
+	var buf bytes.Buffer
+	out = &buf
+
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote("a:2\nb:1\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equalf(t, expected, actual, "Letter frequency failed.")
+}
+
+var testCases = map[string]struct {
+	input         string
+	outputLetters map[rune]int
+	outputSort    []string
+}{
+	"main": {
+		input:         "aba",
+		outputLetters: map[rune]int{'a': 2, 'b': 1},
+		outputSort:    []string{"a:2", "b:1"},
+	},
+	"no duplicates": {
+		input:         "abcd",
+		outputLetters: map[rune]int{'a': 1, 'b': 1, 'c': 1, 'd': 1},
+		outputSort:    []string{"a:1", "b:1", "c:1", "d:1"},
+	},
+	"duplicates with space": {
+		input:         " aaaxbyb bb c z cddsd sd sdsds",
+		outputLetters: map[rune]int{' ': 7, 'a': 3, 'b': 4, 'c': 2, 'd': 6, 's': 5, 'x': 1, 'y': 1, 'z': 1},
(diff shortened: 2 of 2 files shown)


03_letters/saurabhvyas9/main.go
```diff
@@ -0,0 +1,37 @@
...
+		frequencyLetters[str]++
+	}
+	return frequencyLetters
+}
+
+func sortLetters(m map[rune]int) []string {
+	sortedLetters := make([]string, len(m))
+	tempKey := make([]int, 0, len(m))
+	for key := range m {
+		tempKey = append(tempKey, int(key))
```
juliaogris: I appending is more expensive than inserting at known position.
I'd go:
```go
i := 0
keys := make([]int, len(m))
for key := range m {
     keys[i] = int(key)
     i++
}

File: 03_letters/saurabhvyas9/main_test.go [hunk 2]
```diff
@@ -0,0 +1,105 @@
...
+	},
+	"emoticons": {
+		input:         "😃🐻🙏🔥😂😂🤔😂🔥😂😂🔥",
+		outputLetters: map[rune]int{'🐻': 1, '🔥': 3, '😂': 5, '😃': 1, '🙏': 1, '🤔': 1},
+		outputSort:    []string{"🐻:1", "🔥:3", "😂:5", "😃:1", "🙏:1", "🤔:1"},
+	},
+}
+
+func TestSortLetters(t *testing.T) {
+
```
juliaogris: Unnecessary blank line.

03_letters/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,105 @@
...
+	for caseName, test := range testCases {
+		input := test.input
+		expected := test.outputLetters
+		actual := letters(input)
+		assert.Equalf(t, expected, actual, "Letters function failed for TC: %d", caseName)
+	}
+}
+
+// func TestInsertionSort(t *testing.T) {
```
juliaogris: Please remove commented code.

juliaogris (dismissed):
Looks very good, just a few little nit picks.

03_letters/saurabhvyas9/main.go
saurabhvyas9: Understood. Changes Done.

03_letters/saurabhvyas9/main_test.go
saurabhvyas9: Done.

03_letters/saurabhvyas9/main_test.go
saurabhvyas9: My bad. 
Done.

03_letters/saurabhvyas9/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letters(s string) map[rune]int {
+	frequencyLetters := make(map[rune]int)
+	for _, str := range s {
```
camscale, line 15: `str` is a poor choice of name for this - it is not a string and pretty much everyone would expect a variable called `str` to be a string. Just `r` would be sufficient (short for rune) and that makes a lot of sense for ranging over a variable called `s` (which I would expect to be a string).

03_letters/saurabhvyas9/main.go
```diff
@@ -0,0 +1,39 @@
...
+
+func sortLetters(m map[rune]int) []string {
+	sortedLetters := make([]string, len(m))
+	tempKey := make([]int, len(m))
+	i := 0
+	for key := range m {
+		tempKey[i] = int(key)
+		i++
+	}
+	sort.Ints(tempKey)
```
camscale, line 29: This is sorting on the rune value not the count. The lab was ambiguous in this regard so I'm fine with this solution. However I wonder, did you consider the other sort order? If so it would have been a good idea to document your assumption of the order to use, either as a comment in this function, or in the commit message.

No need to make any changes in response to this comment.

03_letters/saurabhvyas9/main_test.go
```diff
@@ -0,0 +1,93 @@
...
+	// When
+	main()
+
+	// Then
+	expected := strconv.Quote("a:2\nb:1\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equalf(t, expected, actual, "Letter frequency failed.")
+}
+
+var testCases = map[string]struct {
```
camscale, line 25: It is nice to see a lot of test cases, but be wary of adding too many. You need to consider whether some test cases are adding any additional testing. For example, the "foreign language" and "emoticons" one are really testing the same thing - it is just a sequence of non-ascii runes. Likewise, your code makes no distinction between alpha, numeric, spaces, etc, so adding tests for those different cases is not testing your code any more widely.

This matters when it comes to maintenance. As code grows, if something in the function changes, possibly all the test cases need to be updated. This could add unnecessary work which becomes tedious after a while.

Next time, try to consider the minimum number of test cases you can come up with that exercise the various branches and boundaries in your code that completely covers it, but no more.

camscale (approved):
LGTM. My comments are more about general software engineering not Go-specific, so I'll approve this for merging. Please consider those comments for later labs though.

03_letters/saurabhvyas9/main.go
saurabhvyas9: Understood - appropriate variable names.

03_letters/saurabhvyas9/main.go
saurabhvyas9: Yes, the sorting is on the rune value. 
I did think of it when I was writing up the special test cases for emoticons etc.

03_letters/saurabhvyas9/main_test.go
saurabhvyas9: Makes sense.

PR #425, opened 2019-06-18
Add letter frequency functionality

patrickmarabeas:
Implements two functions. One which creates a frequency map of given runes, the other sorts the map and returns a slice of "k:v" strings

Fixes #418 

Review of colleague's PR #527

PR #436, opened 2019-06-19
Puppy store

### LINKED ISSUE #88: Lab 6 - CRUD puppy with interface

runnerdave:
Fixes [#88](https://github.com/anz-bank/go-course/issues/88)

Review of colleague's PR [#425](https://github.com/anz-bank/go-course/pull/425#pullrequestreview-251498379)

 - Storer interface implementations as map and sync.map
 - Suite tests

### DIFF
--- a/06_puppy/runnerdave/map_store.go
+++ b/06_puppy/runnerdave/map_store.go
@@ -1,0 +1,47 @@
+package main
+
+import "fmt"
+
+// MapStore represents a simple map storage for the Puppy store
+type MapStore struct {
+	puppies map[uint16]*Puppy
+}
+
+// NewMapStore creates a new in-memory store with map intialised
+func NewMapStore() *MapStore {
+	var m MapStore
+	m.puppies = make(map[uint16]*Puppy)
+	return &m
+}
+
+// CreatePuppy saves new puppy if not in store, if it is already returns error
+func (m *MapStore) CreatePuppy(p *Puppy) error {
+	if _, ok := m.puppies[p.ID]; ok {
+		return fmt.Errorf("puppy with id %d already exists", p.ID)
+	}
+	m.puppies[p.ID] = p
+	return nil
+}
+
+// ReadPuppy reads store by Puppy ID
+func (m *MapStore) ReadPuppy(id uint16) (*Puppy, error) {
+	if puppy, ok := m.puppies[id]; ok {
+		return puppy, nil
+	}
+	return nil, fmt.Errorf("puppy not found")
+}
+
+// UpdatePuppy updates puppy with new value if ID present otherwise creates new Puppy
+func (m *MapStore) UpdatePuppy(id uint16, p *Puppy) error {
+	m.puppies[id] = p
+	return nil
+}
+
+// DeletePuppy deletes a puppy by id from the store
+func (m *MapStore) DeletePuppy(id uint16) (bool, error) {
+	if _, ok := m.puppies[id]; ok {
+		delete(m.puppies, id)
+		return true, nil
+	}
+	return false, fmt.Errorf("puppy with id %d not found", id)
+}
--- a/06_puppy/runnerdave/map_store.go
+++ b/06_puppy/runnerdave/map_store.go
@@ -4,13 +4,13 @@ import "fmt"
 
 // MapStore represents a simple map storage for the Puppy store
 type MapStore struct {
-	puppies map[uint16]*Puppy
+	puppies map[uint16]Puppy
 }
 
 // NewMapStore creates a new in-memory store with map intialised
 func NewMapStore() *MapStore {
 	var m MapStore
-	m.puppies = make(map[uint16]*Puppy)
+	m.puppies = make(map[uint16]Puppy)
 	return &m
 }
 
@@ -19,21 +19,21 @@ func (m *MapStore) CreatePuppy(p *Puppy) error {
 	if _, ok := m.puppies[p.ID]; ok {
 		return fmt.Errorf("puppy with id %d already exists", p.ID)
 	}
-	m.puppies[p.ID] = p
+	m.puppies[p.ID] = *p
 	return nil
 }
 
 // ReadPuppy reads store by Puppy ID
-func (m *MapStore) ReadPuppy(id uint16) (*Puppy, error) {
+func (m *MapStore) ReadPuppy(id uint16) (Puppy, error) {
 	if puppy, ok := m.puppies[id]; ok {
 		return puppy, nil
 	}
-	return nil, fmt.Errorf("puppy not found")
+	return Puppy{}, fmt.Errorf("puppy not found")
 }
 
 // UpdatePuppy updates puppy with new value if ID present otherwise creates new Puppy
 func (m *MapStore) UpdatePuppy(id uint16, p *Puppy) error {
-	m.puppies[id] = p
+	m.puppies[id] = *p
 	return nil
 }
 
--- a/06_puppy/runnerdave/storer_test.go
+++ b/06_puppy/runnerdave/storer_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	tassert "github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/stretchr/testify/suite"
 )
 
@@ -22,16 +23,16 @@ func (s *storerSuite) TestUpdatePuppyIDDoesNotExist() {
 	// given
 	assert := tassert.New(s.T())
 	testPuppy := puppy1()
+	r := require.New(s.T())
 
 	// when
-	err := s.store.UpdatePuppy(11, &testPuppy)
+	uerr := s.store.UpdatePuppy(11, &testPuppy)
 
 	// then
-	assert.NoError(err, "Updating to non-existing puppy ID is not an error")
+	r.NoError(uerr, "Updating to non-existing puppy ID is not an error")
 	newPuppy, err := s.store.ReadPuppy(11)
-	if assert.NoError(err, "Should be able to read a newly added puppy via update") {
-		assert.Equal(&testPuppy, newPuppy, "Newly added puppy should be equal to test puppy")
-	}
+	r.NoError(err, "Should be able to read a newly added puppy via update")
+	assert.Equal(testPuppy, newPuppy, "Newly added puppy should be equal to test puppy")
 }
 
 func (s *storerSuite) TestReadPuppy() {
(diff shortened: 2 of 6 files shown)


06_puppy/runnerdave/syncStore.go
```diff
@@ -0,0 +1,54 @@
+package main
```
alfredxiao: Seems to me most go package and filenames are in lowercase only, feels a bit weird with `syncStore.go`

runnerdave: Good point, updated

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+
+// NewMapStore creates a new in-memory store with map intialised
+func NewMapStore() *MapStore {
+	var m MapStore
+	m.puppies = make(map[uint16]Puppy)
+	return &m
+}
+
+// CreatePuppy saves new puppy if not in store, if it is already returns error
+func (m *MapStore) CreatePuppy(p *Puppy) error {
```
alfredxiao, line 18: Since your store stores `Puppy` rather than `*Puppy`, I think there is no need to accept `*Puppy` as argument here.

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+	if _, ok := m.puppies[p.ID]; ok {
```
alfredxiao, line 19: Some think you should provision ID here, up to you to decide in that regard.

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+		return fmt.Errorf("puppy with id %d already exists", p.ID)
+	}
+	m.puppies[p.ID] = *p
+	return nil
+}
+
+// ReadPuppy reads store by Puppy ID
+func (m *MapStore) ReadPuppy(id uint16) (*Puppy, error) {
+	if puppy, ok := m.puppies[id]; ok {
+		return &puppy, nil
```
alfredxiao: I am not sure if returning a pointer is acceptable here, that may lead to a situation that some code can use the pointer to update the puppy (struct) without going through `UpdatePuppy` function, which IMO should be avoided.

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	return nil, fmt.Errorf("puppy not found")
+}
+
+// UpdatePuppy updates puppy with new value if ID present otherwise creates new Puppy
+func (m *MapStore) UpdatePuppy(id uint16, p *Puppy) error {
```
alfredxiao, line 35: Same as above, should we use `*Puppy` argument here?

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+	m.puppies[id] = *p
+	return nil
+}
+
+// DeletePuppy deletes a puppy by id from the store
+func (m *MapStore) DeletePuppy(id uint16) (bool, error) {
```
alfredxiao, line 41: I am ok with either, but you could consider simplifying this API by returning either only error or bool.

06_puppy/runnerdave/types.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+// Puppy represents an item in the puppy store
+type Puppy struct {
+	Breed  string  `json:"name"`
+	Colour string  `json:"color"`
+	ID     uint16  `json:"id"`
+	Value  float32 `json:"value"`
```
alfredxiao, line 8: Not a big deal, but json is not required for this lab IMO.

06_puppy/runnerdave/sync_store.go
```diff
@@ -0,0 +1,54 @@
+package main
+
+import (
+	"fmt"
+	"sync"
+)
+
+// SyncStore implementation of the Storer interface
+type SyncStore struct {
+	sync.Mutex
```
alfredxiao, line 10: I am not experienced in using `sync.Mutex` but I think it is redundant in sync_store - otherwise, what's the point in using a `sync.Map` if we need to lock/unlock all the time?

06_puppy/runnerdave/storer_test.go
```diff
@@ -0,0 +1,119 @@
...
+		return Puppy{ID: 11, Breed: "Chihuahua", Colour: "Brown", Value: 12.30}
+	}
+)
+
+type storerImpl int
+
+type storerSuite struct {
+	suite.Suite
+	store Storer
+	impl  storerImpl
```
alfredxiao: I might have missed something, is `impl` field actually used here?

06_puppy/runnerdave/storer_test.go
```diff
@@ -0,0 +1,119 @@
...
+		store: NewSyncStore(),
+		impl:  mem,
+	}
+	mapSuite := storerSuite{
+		store: NewMapStore(),
+		impl:  mem,
+	}
+	suite.Run(t, &syncSuite)
+	suite.Run(t, &mapSuite)
+}
```
alfredxiao, line 112: You might want to consider whether all Test* functions actually use the same store instance (in the case of MemStore and SyncStore respectively), or different instances. If they use the same instance, then the outcome of one test could affect another (because they same the same store instance, and most tests mutate the store state), which makes your tests not independent of each other.

06_puppy/runnerdave/map_store.go
runnerdave: Good point I have made a change so my store stores *Puppy now

06_puppy/runnerdave/map_store.go
runnerdave: I just want to keep it simple for now

06_puppy/runnerdave/map_store.go
runnerdave: fixed, good observation

06_puppy/runnerdave/map_store.go
runnerdave: fixed

06_puppy/runnerdave/map_store.go
runnerdave: I am following the hint given, and they use the same: https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/memstore.go I think is useful to practice this aspect of the language.

06_puppy/runnerdave/types.go
runnerdave: you are right but it will be used later

06_puppy/runnerdave/storer_test.go
runnerdave: I perform a cleanup in each test when needed, making them independent that way

06_puppy/runnerdave/sync_store.go
runnerdave: I am not that experienced either, but I am following the hint given at: https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/memstore.go#L8

06_puppy/runnerdave/storer_test.go
runnerdave: good catch, cleaned it from the code

06_puppy/runnerdave/storer_test.go
alfredxiao: In this lab scale, no problem (small and you are the sole owner). Wouldn't recommend this in real scale project, that can be tricky to manage and maintain especially with many team devs.

06_puppy/runnerdave/map_store.go
alfredxiao: Interesting, @juliaogris pointed this out in my code, the the given sample does it as well.

06_puppy/runnerdave/map_store.go
alfredxiao: Absolute fine from my perspective.

06_puppy/runnerdave/storer_test.go
runnerdave: absolutely agree

runnerdave:
@alfredxiao could you please approve if you are happy with the changes?

06_puppy/runnerdave/map_store.go
anzboi: A consequence of storing `*Puppy` is now the value returned by `Read` method can be modified and will modify the stored puppy.
```go
p := store.Read(1)
p.Name = "Fido"
q := store.Read(1)
q.Name == "Fido" // true
```
This is not a huge issue for the lab, but can introduce weird bugs to someone using the mapstore. IMO to fix
```go
func ReadPuppy(id uint16) (*Puppy, error) {
  if puppy, ok := m.puppies[id]; ok {
    p := *puppy // Copy the value to a new variable
    return &p, nil // modifying this return will not affect stored value
  }
  ...
}
```

06_puppy/runnerdave/storer_test.go
```diff
@@ -0,0 +1,110 @@
...
+		assert.Equal(&testPuppy, newPuppy, "Newly added puppy should be equal to test puppy")
+	}
+}
+
+func (s *storerSuite) TestDeletePuppy() {
+	// given
+	assert := tassert.New(s.T())
+	testPuppy := puppy1()
+	createError := s.store.CreatePuppy(&testPuppy)
+	s.T().Log(createError)
```
anzboi: you can use testify require to terminate the test here if you get an error on create
```
r := require.New(t)
r.NoError(err, "Create should not produce an error")
```

06_puppy/runnerdave/storer_test.go
```diff
@@ -0,0 +1,110 @@
...
+
+	// when
+	isDeleted, err := s.store.DeletePuppy(11)
+
+	// then
+	if assert.NoError(err, "Should be able to delete a newly added puppy") {
```
anzboi: Similar here,
```
// Prevents the rest of the test running if the delete fails
r.NoError(err, "Should be able to delete an existing puppy") 

a.True(isDeleted, "A successful delete should return true")
_, err = s.store.ReadPuppy(11)
a.Error(err, "Should return error when attempting to read a deleted puppy")
```

anzboi (changes requested):
This looks really good, just a couple of things

1. Your mapstore allows modification by modifying the returned puppy from read, rather than going through update. I've suggested a fix to read.
2. I've suggested cleaner test code, don't feel the need to change this is it's fine as is, just using require and assert together can make for fairly nice test code.

06_puppy/runnerdave/storer_test.go
runnerdave: implemented

06_puppy/runnerdave/storer_test.go
runnerdave: implemented

06_puppy/runnerdave/map_store.go
```diff
@@ -0,0 +1,47 @@
...
+		return fmt.Errorf("puppy with id %d already exists", p.ID)
+	}
+	m.puppies[p.ID] = *p
+	return nil
+}
+
+// ReadPuppy reads store by Puppy ID
+func (m *MapStore) ReadPuppy(id uint16) (Puppy, error) {
+	if puppy, ok := m.puppies[id]; ok {
+		return puppy, nil
```
anzboi, line 29: That works too 👍

anzboi (approved):
LGTM

PR #443, opened 2019-06-20
Lab 6 CRUD Puppy

### LINKED ISSUE #442: Lab 6 - Basic CRUD Puppy
- [ ] Create an executable go program in directory `06_puppy/USERNAME`
- [ ] Implement a `Puppy` struct containing `ID`, `Breed`, `Colour`, `Value`
- [ ] Create `Storer` interface with C.R.U.D. methods for `Puppy`
- [ ] Write a `MapStore` implementation of `Storer` backed by a `map`
- [ ] Write a `SyncStore` implementation of `Storer` backed by a `sync.Map`
- [ ] Keep all implementation files in the same folder and in package `main`
- [ ] Test against the `Storer` interface and run in suite with both implementations

marksomething:
Fixes #442 

Review of colleague's PR #476

Implement minimal CRUD Puppy Storage whilst
avoiding implementing the requirements of Lab 7
All error cases are silently ignored

- CreatePuppy stores a Puppy at the ID in the Puppy unless one is already present
- ReadPuppy returns the puppy at the ID given, or a nil Puppy if not found
- UpdatePuppy replaces the Puppy at the ID given if there is a Puppy there and
  the replacement Puppy shares the same ID
- DeletePuppy deletes the Puppy at the given ID if one exists

### DIFF
--- a/06_puppy/marksomething/main.go
+++ b/06_puppy/marksomething/main.go
@@ -1,0 +1,91 @@
+package main
+
+import "sync"
+
+// Puppy represents a small dog
+type Puppy struct {
+	ID     uint32
+	Breed  string
+	Colour string
+	Value  int
+}
+
+// PuppyStorer stores an object in a key-value store
+type PuppyStorer interface {
+	CreatePuppy(p Puppy)
+	ReadPuppy(k uint32) Puppy
+	UpdatePuppy(k uint32, p Puppy)
+	DeletePuppy(k uint32)
+}
+
+// MapStore is a storage mechanism for Puppy
+type MapStore map[uint32]Puppy
+
+// CreatePuppy adds a Puppy to storage unless a Puppy of the same ID is stored in which case nothing is done
+func (m MapStore) CreatePuppy(p Puppy) {
+	if _, found := m[p.ID]; found {
+		return
+	}
+	m[p.ID] = p
+}
+
+// ReadPuppy returns a Puppy from storage, or a blank Puppy where no Puppy of that id is found
+func (m MapStore) ReadPuppy(id uint32) Puppy {
+	return m[id]
+}
+
+// UpdatePuppy updates an existing puppy if id matches the Puppy and id is found in the storage, otherwise nothing
+func (m MapStore) UpdatePuppy(id uint32, p Puppy) {
+	if id != p.ID {
+		return
+	}
+	if _, ok := m[id]; !ok {
+		return
+	}
+	m[id] = p
+}
+
+// DeletePuppy removes a puppy from storage if it exists
+func (m MapStore) DeletePuppy(id uint32) {
+	delete(m, id)
+}
+
+// SyncStore is a storage mechanism for Puppy
+type SyncStore struct{ sync.Map }
+
+// CreatePuppy adds a Puppy to storage unless a Puppy of the same ID is stored in which case nothing is done
+func (s *SyncStore) CreatePuppy(p Puppy) {
+	if _, found := s.Load(p.ID); found {
+		return
+	}
+	s.Store(p.ID, p)
+}
+
+// ReadPuppy returns a Puppy from storage, or a blank Puppy where no Puppy of that id is found
+func (s *SyncStore) ReadPuppy(id uint32) Puppy {
+	p, ok := s.Load(id)
+	if !ok {
+		return Puppy{}
+	}
+	return p.(Puppy)
+}
+
+// UpdatePuppy updates an existing puppy if id matches the Puppy and id is found in the storage, otherwise nothing
+func (s *SyncStore) UpdatePuppy(id uint32, p Puppy) {
+	if id != p.ID {
+		return
+	}
+	if _, ok := s.Load(id); !ok {
+		return
+	}
+	s.Store(id, p)
+}
+
+// DeletePuppy removes a puppy from storage if it exists
+func (s *SyncStore) DeletePuppy(id uint32) {
+	s.Delete(id)
+}
+
+func main() {
+
+}
--- a/06_puppy/marksomething/main_test.go
+++ b/06_puppy/marksomething/main_test.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type PuppyStorerTest struct {
+	suite.Suite
+	store        PuppyStorer
+	storeFactory func() PuppyStorer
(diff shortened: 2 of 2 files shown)


06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+	Value  int
+}
+
+// PuppyStorer stores an object in a key-value store
+type PuppyStorer interface {
+	CreatePuppy(p Puppy)
+	ReadPuppy(k uint32) Puppy
+	UpdatePuppy(k uint32, p Puppy)
+	DeletePuppy(k uint32)
+}
```
alfredxiao, line 19: None of your interface functions return errors, so how does function caller know if anything goes wrong or transaction success? For example, ReadPuppy or DeletePuppy with non-existing ID?

juliaogris: Agreed, add some basic error handling please, that will make lab 6 a lot easier too.

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+// MapStore is a storage mechanism for Puppy
+type MapStore map[uint32]Puppy
+
+// CreatePuppy adds a Puppy to storage unless a Puppy of the same ID is stored in which case nothing is done
+func (m MapStore) CreatePuppy(p Puppy) {
+	if _, found := m[p.ID]; found {
```
juliaogris, line 26: `found` should be `ok` ([comma ok idiom](https://golang.org/doc/effective_go.html#maps))

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// ReadPuppy returns a Puppy from storage, or a blank Puppy where no Puppy of that id is found
+func (m MapStore) ReadPuppy(id uint32) Puppy {
+	return m[id]
+}
+
+// UpdatePuppy updates an existing puppy if id matches the Puppy and id is found in the storage, otherwise nothing
+func (m MapStore) UpdatePuppy(id uint32, p Puppy) {
+	if id != p.ID {
```
juliaogris, line 39: maybe no need to have a separate id arg?

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+// UpdatePuppy updates an existing puppy if id matches the Puppy and id is found in the storage, otherwise nothing
+func (s *SyncStore) UpdatePuppy(id uint32, p Puppy) {
+	if id != p.ID {
+		return
+	}
+	if _, ok := s.Load(id); !ok {
+		return
+	}
+	s.Store(id, p)
```
juliaogris, line 81: thins function needs to use a mutex (lock at before load, unlock after store) to ensure there are no race conditions (the reason for using a syncmap in the first place).

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+func (m MapStore) DeletePuppy(id uint32) {
+	delete(m, id)
+}
+
+// SyncStore is a storage mechanism for Puppy
+type SyncStore struct{ sync.Map }
+
+// CreatePuppy adds a Puppy to storage unless a Puppy of the same ID is stored in which case nothing is done
+func (s *SyncStore) CreatePuppy(p Puppy) {
+	if _, found := s.Load(p.ID); found {
```
juliaogris, line 58: thins function needs to use a mutex (lock at before load, unlock after store) to ensure there are no race conditions (the reason for using a syncmap in the first place).

juliaogris (changes requested):
Please add basic error handling and respond to comments inline.
I'd say it's also time to bring things up into different files: main.go, storer.go or types.go, mapstore.go, syncmapsotre.go or so....

06_puppy/marksomething/main.go
alextmz: Agreed

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
+package main
+
+import "sync"
+
+// Puppy represents a small dog
+type Puppy struct {
+	ID     uint32
+	Breed  string
+	Colour string
+	Value  int
```
alextmz, line 10: Values need to hold partial values (eg. $1.99) so a int is not fit for it.

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+// ReadPuppy returns a Puppy from storage, or a blank Puppy where no Puppy of that id is found
+func (m MapStore) ReadPuppy(id uint32) Puppy {
+	return m[id]
+}
+
+// UpdatePuppy updates an existing puppy if id matches the Puppy and id is found in the storage, otherwise nothing
+func (m MapStore) UpdatePuppy(id uint32, p Puppy) {
+	if id != p.ID {
+		return
```
alextmz, line 40: That would be a case to return an error - here the calling function won't know if puppy was updated or not

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	if _, ok := m[id]; !ok {
+		return
```
alextmz, line 43: same as above

06_puppy/marksomething/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	m[id] = p
+}
+
+// DeletePuppy removes a puppy from storage if it exists
+func (m MapStore) DeletePuppy(id uint32) {
+	delete(m, id)
```
alextmz, line 50: and if `id` doesn't exist?

06_puppy/marksomething/main.go
alextmz: > thins function needs to use a mutex (lock at before load, unlock after store) to ensure there are no race conditions (the reason for using a syncmap in the first place).

Although it would be nice, mutexes are a Lab 7 specification, not a requisite on 6.

alextmz (changes requested):
Tests were not reviewed - no point in reviewing them before basic error treatment is thoroughly addressed in main code, as the tests will need to change a lot.

Also, specification for lab 6 is "Keep all implementation files in the same folder and in package main" - not necessarily in the same *file* main - you could separate the Map and the Sync store in files for better readability and to make your life easier.

juliaogris:

PR #445, opened 2019-06-20
Lab5 - Implement String() and add tests

alfredxiao:
Fixes #444 

Review of colleague's PR #412

PR #454, opened 2019-06-20
Feature/lab 2

### LINKED ISSUE #453: Lab 02 - Bubble
# Lab 2 - Bubble sort

- Create an executable go program in directory `02_bubble/USERNAME`
- Write a function that returns a sorted copy of `int` slice `s` using [bubble sort](https://en.wikipedia.org/wiki/Bubble_sort):

```
func bubble(s []int) []int
```

```
[1 2 3 5]
```

- Bonus points: implement [Insertion sort](https://en.wikipedia.org/wiki/Insertion_sort)
- Extra bonus points: implement an _O(n_ _log(n))_ sorting algorithm

starfallz:
Fixes #453

Review of colleague's PR #445

- Write a function that returns a sorted copy of `int` slice `s` using [bubble sort]

### DIFF
--- a/02_bubble/starfallz/main.go
+++ b/02_bubble/starfallz/main.go
@@ -9,5 +9,29 @@ import (
 var out io.Writer = os.Stdout
 
 func main() {
-	fmt.Fprintln(out, "test")
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+	for i := 0; i < len(s); i++ {
+		isSwapped := false
+		for j := 0; j < len(s)-1; j++ {
+			swap(s, j, &isSwapped)
+		}
+
+		if !isSwapped {
+			break
+		}
+	}
+
+	return s
+}
+
+func swap(s []int, j int, isSwapped *bool) {
+	tempVal := s[j+1]
+	if s[j] > tempVal {
+		s[j+1] = s[j]
+		s[j] = tempVal
+		*isSwapped = true
+	}
 }
--- a/02_bubble/starfallz/main.go
+++ b/02_bubble/starfallz/main.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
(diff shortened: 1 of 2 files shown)


02_bubble/starfallz/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
```
n0npax, line 15: the function should not modify passed slice:
`Write a function that returns a sorted copy of int slice s using bubble sort:`
In your case, you working on `s` reference

02_bubble/starfallz/main.go
```diff
@@ -0,0 +1,37 @@
...
+
+		if !isSwapped {
+			break
+		}
+	}
+
+	return s
+}
+
+func swap(s []int, j int, isSwapped *bool) {
```
n0npax, line 30: the function name is confusing.
This function could be named `swapIfNextValIsGrater`, but this has no sense in my mind.
1) I would suggest moving this logic into to `bubble` function
2) get rid of `tempVal` and implement swapping by sth like `s[i],s[i+1] = s[i+1],s[i]`

n0npax (changes requested):

02_bubble/starfallz/main.go
starfallz: If I'm not wrong, slice by default would not be passed by reference, it would be by value so I'm not modifying the original slice passed in. Again if I'm not wrong, slice will always create a new copy whenever you do anything to it.

I did a quick check on it:

https://play.golang.org/p/r-apnquHx0n

02_bubble/starfallz/main.go
starfallz: Cool thanks for the shortcut for the tempVal thing.

I generally love separating concerns into smaller functions so I can unit test them in isolation. Although I do agree that the name could be better so I will rename the func if Ur cool with that.

juliaogris:

PR #438, opened 2019-06-20
Lab4 Implement numeronyms and tests

alfredxiao:
Fixes #437 

Review of colleague's PR #410 

#### Changes proposed in this PR:
- Implement numeronyms
- Test numeronyms and main

PR #458, opened 2019-06-20
lab 02 updated, issues raised by bot fixed

### LINKED ISSUE #441: Lab 02 Bubblesort and Sorting
Create an executable go program in directory 02_bubble/marchitect
Write a function that returns a sorted copy of int slice s using bubble sort
Bonus points: implement Insertion sort
Extra bonus points: implement an O(n log(n)) sorting algorithm

Marchitect:
Fixes #441

Review of colleague's PR #438 

Changes proposed in this PR:
- added sorting functions and test cases, fixed issues raised by bot

### DIFF
--- a/02_bubble/marchitect/main.go
+++ b/02_bubble/marchitect/main.go
@@ -1,0 +1,105 @@
+package main
+
+import (
+	"crypto/rand"
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	// n is the number of items in the int array
+	n := len(s)
+	// make a copy of the original int array s
+	result := make([]int, n)
+	copy(result, s)
+	// maybe necessary
+	if n < 2 {
+		return result
+	}
+	// set swapped to true
+	swapped := true
+	// in Go, use keyword for to act as a while loop
+	for swapped {
+		// set swapped to false
+		swapped = false
+		// iterate through all of the elements in our list
+		for i := 1; i < n; i++ {
+			// if the current element is greater than the next element, swap them
+			if result[i-1] > result[i] {
+				// swap values - Go tuple assignment
+				result[i], result[i-1] = result[i-1], result[i]
+				// set swap back to true
+				swapped = true
+			}
+		}
+	}
+	return result
+}
+
+func insertion(s []int) []int {
+	// n is the number of items in the int array
+	n := len(s)
+	// make a copy of the original int array s
+	result := make([]int, n)
+	copy(result, s)
+	// maybe necessary
+	if n < 2 {
+		return result
+	}
+	var i, j, key int
+	for i = 1; i < n; i++ {
+		// set the key, a value holder, to be the second item in the int array
+		key = result[i]
+		// set j to be the item before key in the int array
+		j = i - 1
+		// for elements of array result[0..i-1], that are greater than key
+		for j >= 0 && result[j] > key {
+			// sort those greater than key elements one position after current position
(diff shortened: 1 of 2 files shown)


02_bubble/marchitect/main.go
```diff
@@ -0,0 +1,99 @@
...
+			// if the current element is greater than the next element, swap them
+			if s[i-1] > s[i] {
+				// swap values - Go tuple assignment
+				s[i], s[i-1] = s[i-1], s[i]
+				// set swap back to true
+				swapped = true
+			}
+		}
+	}
+	return s
```
davidbroughsmyth: The lab asks for returning a sorted copy of int slice s. Returning s is not a copy.
You can make a copy by making a slice then copy:
```
copy := make([]int, len(s))
copy(copy, s)
```

davidbroughsmyth (changes requested):
Need to change sort functions to returns a sorted copy of int slice s

davidbroughsmyth:
You will also need to review a colleague's PR.

02_bubble/marchitect/main.go
Marchitect: Thank you for the feedback, will improve the code accordingly. And yes I aim to spare some time to review more PR too

Marchitect: > You will also need to review a colleague's PR.

Kindly please help me check if the updated commit is good to fix the issue :) Thank you David.

davidbroughsmyth: Nicely done. Looks good to me.

davidbroughsmyth:
And I like how you made use of `if !reflect.DeepEqual(actual, expected) {`in your tests.

02_bubble/marchitect/main.go
```diff
@@ -0,0 +1,105 @@
...
+	"fmt"
+	"io"
+	"math/big"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	// n is the number of items in the int array
```
juliaogris, line 14: don't comment the obvious (below as well).

02_bubble/marchitect/main.go
```diff
@@ -0,0 +1,105 @@
...
+		return result
+	}
+	// set swapped to true
+	swapped := true
+	// in Go, use keyword for to act as a while loop
+	for swapped {
+		// set swapped to false
+		swapped = false
+		// iterate through all of the elements in our list
+		for i := 1; i < n; i++ {
```
juliaogris, line 30: The inner loop usually goes from `n` down. Reason: after the first outer iteration the largest element has "bubbled up" to the slice position with the higher index.
Our reference solution looks like this - feel free to use, adjust or ignore:
```go
	for i := len(s); i > 0; i-- {
		swapped := false
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
				swapped = true
			}
		}
		if !swapped {
			return
		}
	}
```

02_bubble/marchitect/main.go
```diff
@@ -0,0 +1,105 @@
...
+func bubble(s []int) []int {
+	// n is the number of items in the int array
+	n := len(s)
+	// make a copy of the original int array s
+	result := make([]int, n)
+	copy(result, s)
+	// maybe necessary
+	if n < 2 {
+		return result
+	}
```
juliaogris, line 22: nope ;) not necessary.

juliaogris (changes requested):
Looks pretty good.
I think you have gone overboard a bit with the comments, which IMO hinders readability.

PR #459, opened 2019-06-20
Feature/lab 3 letters

### LINKED ISSUE #456: Lab 03 - Letters
# Lab 3 - Letter frequency

```
```

```
```

Call `fmt.Println(strings.Join(sortLetters(letters("aba"))), "\n")` in `main` to print:

```
a:2
b:1
```

Bonus points: comprehensive tests

starfallz:
Fixes #456

Review of colleague's PR #

- Write a function that returns a mapping of each letter to its frequency

### DIFF
--- a/03_letters/starfallz/main.go
+++ b/03_letters/starfallz/main.go
@@ -4,10 +4,46 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sort"
+	"strconv"
+	"strings"
 )
 
 var out io.Writer = os.Stdout
 
 func main() {
-	fmt.Fprintln(out, "test letters")
+	fmt.Fprint(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	letterCount := make(map[rune]int)
+
+	for _, currentLetter := range s {
+		letterCount[currentLetter]++
+	}
+
+	return letterCount
+}
+
+func sortLetters(m map[rune]int) []string {
+	allKeys := make([]rune, 0, len(m))
+	for k := range m {
+		allKeys = append(allKeys, k)
+	}
+
+	sortRune(allKeys)
+
+	result := make([]string, len(allKeys))
+
+	for i, currentKey := range allKeys {
+		result[i] = strings.Join([]string{string(currentKey), strconv.Itoa(m[currentKey])}, ":")
+	}
+
+	return result
+}
+
+func sortRune(r []rune) {
+	sort.Slice(r, func(i, j int) bool {
+		return r[i] < r[j]
+	})
 }
--- a/03_letters/starfallz/main.go
+++ b/03_letters/starfallz/main.go
@@ -1,0 +1,13 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, "test letters")
+}
--- a/03_letters/starfallz/main_test.go
+++ b/03_letters/starfallz/main_test.go
@@ -1,0 +1,1 @@
+package main
--- a/03_letters/starfallz/main_test.go
+++ b/03_letters/starfallz/main_test.go
@@ -1,1 +1,103 @@
 package main
+
+import (
+	"bytes"
+	"reflect"
+	"strconv"
+	"testing"
+)
+
+func TestMainFunction(t *testing.T) {
+	t.Run("Test main to return string values of letters(aba) with proper formatting", func(t *testing.T) {
+		var buf bytes.Buffer
+		out = &buf
+
+		main()
+
+		expected := strconv.Quote(`a:2
+b:1`)
+		actual := strconv.Quote(buf.String())
+
+		if expected != actual {
+			t.Errorf("Unexpected output, expected: %s, actual: %s", expected, actual)
+		}
+	})
+}
+
+func TestLettersFunction(t *testing.T) {
+	testCases := []struct {
+		description string
+		input       string
+		expected    map[rune]int
+	}{
+		{"Test letters to return sorted map of letters based on rune # with count", "abaabazzzzzz",
+			map[rune]int{97: 4, 98: 2, 122: 6}},
+		{"Test letters to handle uppercase letters", "aAAaA", map[rune]int{65: 3, 97: 2}},
+		{"Test letters to handle non-alphabetical characters", "    !!!*", map[rune]int{32: 4, 33: 3, 42: 1}},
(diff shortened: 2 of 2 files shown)


03_letters/starfallz/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func TestSortRuneFunction(t *testing.T) {
+	testCases := []struct {
+		description string
+		input       []rune
+		expected    []rune
+	}{
+		{"Test sort unsorted rune slice to sorted rune slice", []rune{97, 82}, []rune{82, 97}},
+		{"Test sort negative rune", []rune{-97, 82}, []rune{-97, 82}},
```
n0npax: I would prefer to keep rune as rune: i.e. `e` instead of 99

03_letters/starfallz/main.go
```diff
@@ -0,0 +1,49 @@
...
+
+	for _, currentLetter := range s {
+		letterCount[currentLetter]++
+	}
+
+	return letterCount
+}
+
+func sortLetters(m map[rune]int) []string {
+	allKeys := make([]rune, 0, len(m))
```
n0npax, line 29: I believe `len: 0` and `cap: len(m)` should be replaced with  `len: len(m)`. You don't plan to grow slice, you already know the size.

03_letters/starfallz/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+func TestSortLettersFunction(t *testing.T) {
+	testCases := []struct {
+		description string
+		input       map[rune]int
+		expected    []string
+	}{
+		{"Test sort letters using letters() as input and return count", letters("aba"), []string{"a:2", "b:1"}},
+		{"Test sort letters with uppercase using letters() as input and return count", letters("Aba"),
+			[]string{"A:1", "a:1", "b:1"}},
+		{"Test sort letters with uppercase and return count", map[rune]int{97: 2, 65: 3}, []string{"A:3", "a:2"}},
```
n0npax, line 62: i would prefer to keep rune as rune (i.e. `'a':2 `instead of `97:2`

n0npax (changes requested):
good work mate. Just 2 minor issues.

03_letters/starfallz/main_test.go
starfallz: Cool no worries will change them. It's easier to read too. Thanks

03_letters/starfallz/main.go
starfallz: Yeah you are right I am not sure why I did this. I will try it and see if that works or not. Thanks

03_letters/starfallz/main.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+func letters(s string) map[rune]int {
+	letterCount := make(map[rune]int)
+
+	for _, currentLetter := range s {
+		letterCount[currentLetter]++
+	}
+
+	return letterCount
```
alextmz, line 25: Minor nitpick: on small functions like these, and between blocks that have just 1 line (the return),  internal line spacing doesn't add legibility, to the contrary. What do you think?

03_letters/starfallz/main.go
```diff
@@ -0,0 +1,49 @@
...
+		allKeys = append(allKeys, k)
+	}
+
+	sortRune(allKeys)
+
+	result := make([]string, len(allKeys))
+
+	for i, currentKey := range allKeys {
+		result[i] = strings.Join([]string{string(currentKey), strconv.Itoa(m[currentKey])}, ":")
+	}
```
alextmz, line 40: As a suggestion I'd try to do it without creating an intermediate slice - that's how I did it, but not yet reviewed, so not sure if it's the right way to go but I do think so. Makes things a lot shorter too.

03_letters/starfallz/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+		main()
+
+		expected := strconv.Quote(`a:2
+b:1`)
+		actual := strconv.Quote(buf.String())
+
+		if expected != actual {
+			t.Errorf("Unexpected output, expected: %s, actual: %s", expected, actual)
+		}
+	})
```
alextmz, line 24: I'm not usually very big on getting things formatted properly but it definitely looks weird with the long line plus the extra spacing plus the line break for the literals. 

Also, don't capitalize error strings throughout (from https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments, see also https://github.com/golang/go/wiki/CodeReviewComments#error-strings

03_letters/starfallz/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+func TestLettersFunction(t *testing.T) {
+	testCases := []struct {
+		description string
+		input       string
+		expected    map[rune]int
+	}{
+		{"Test letters to return sorted map of letters based on rune # with count", "abaabazzzzzz",
+			map[rune]int{97: 4, 98: 2, 122: 6}},
+		{"Test letters to handle uppercase letters", "aAAaA", map[rune]int{65: 3, 97: 2}},
+		{"Test letters to handle non-alphabetical characters", "    !!!*", map[rune]int{32: 4, 33: 3, 42: 1}},
```
alextmz, line 36: Nice one testing punctuation and white spaces, I'd suggest also testing Unicode (eg. Emojis and whatanot). There's a whole bunch of weird characters out there.

03_letters/starfallz/main_test.go
alextmz: Would make it much easier to read the weird Unicode too. By definition Go source code is UTF-8 encoded, so you shouldn't be afraid of putting weird characters in, every tool used should either support it or be abandoned.

03_letters/starfallz/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+		expected := testCase.expected
+		t.Run(testCase.description, func(t *testing.T) {
+			sortRune(actual)
+
+			if !reflect.DeepEqual(actual, expected) {
+				t.Errorf("Unexpected output, expected: %d, actual: %d", expected, actual)
+			}
+		})
+	}
+}
```
alextmz, line 103: The TestSortLettersFunction already tests sortRune indirectly, so I'd suggest moving the tests from TestSortRuneFunction to the former to simplify things.

alextmz (changes requested):
The Uppercase error strings should be fixed, it is a nice habit to have that pays over time. I'd also suggest using characters directly instead of their code, as it makes things more legible (IMHO)

juliaogris:

PR #471, opened 2019-06-22
Lab2 Complete bubble sort algorithm and write test scripts

nicko-lee:
Fixes #52 
Review of colleague's PR #487 
Implement lab2 solution and add tests


02_bubble/nickolee/bubblesort_test.go
```diff
@@ -0,0 +1,40 @@
...
+	test1Expected := []int{}
+
+	// Test random slice
+	test2 := bubbleSort([]int{3, 2, 5, 1, 7, 2, 9, 1, 0})
+	test2Expected := []int{0, 1, 1, 2, 2, 3, 5, 7, 9}
+
+	// Check whether expected results equal actual results
+	if !reflect.DeepEqual(test1, test1Expected) || !reflect.DeepEqual(test2, test2Expected) {
+		t.Errorf("TestBubbleSortFunc: Unexpected output in main()")
+	}
```
davidbroughsmyth, line 35: Consider table driver test as referred to in the course https://dave.cheney.net/2019/05/07/prefer-table-driven-tests
Also a wider range of boundary tests like the worst case sorting using reverse numbers 9,8,7, ... 0, and what about negative numbers.

davidbroughsmyth (commented):
Looks good to me. My comments are for something to consider in future labs.
I've also added "Lab2" to your PR title. When you have time add your review of colleague's PR.

02_bubble/nickolee/bubblesort_test.go
davidbroughsmyth: Review of colleague's PR noted. I'm happy to approve.

02_bubble/nickolee/bubblesort.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+// general sorting algorithm that sorts by "bubbling" smaller numbers to the left
+func bubbleSort(s []int) []int {
+	result := make([]int, len(s))
+	copy(result, s)
+
+	unsorted := true
```
anzboi: Usually you don't want negators in names for booleans. The reason is it makes it harder to read statements
eg:
`!unsorted` is the same as `sorted`
`for unsorted` is the same as `for !sorted`
`unsorted = false` is the same as `sorted = true`

In general, it just reads better and makes it easier to understand, even when theres more symbols (`!sorted` reads just as well as `unsorted`).

02_bubble/nickolee/bubblesort.go
```diff
@@ -0,0 +1,33 @@
...
+	for unsorted {
+		// perform the check to see if sorting is complete to exit for loop
```
anzboi: Comments like this usually go on the same line or before the check.

02_bubble/nickolee/bubblesort_test.go
anzboi: I must insist on table driven tests here.

anzboi (changes requested):
Great work so far, functionality looks perfect, three things I would like improved
1. Rename `unsorted` to `sorted` and invert all uses.
1. Use table driven tests.
1. A test to check the original slice is not sorted
I would also like to see test cases for 1 element and an already sorted slice.

02_bubble/nickolee/bubblesort.go
nicko-lee: you mean negators in the variable name itself right?

nicko-lee: so sorted = false

02_bubble/nickolee/bubblesort_test.go
nicko-lee: Yep have implemented this now

02_bubble/nickolee/bubblesort.go
nicko-lee: Fixed this

02_bubble/nickolee/bubblesort.go
nicko-lee: Refactored to reflect this

02_bubble/nickolee/bubblesort_test.go
```diff
@@ -0,0 +1,49 @@
...
+		{name: "descending order", input: []int{10, 9, 8, 7, 6, 5, 4, 3, 2, 1},
+			expected: []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}},
+		{name: "same number", input: []int{1, 1, 1, 1}, expected: []int{1, 1, 1, 1}},
+		{name: "empty array", input: []int{}, expected: []int{}},
+		{name: "average case", input: []int{59, 62, 48, 53, 100, 86, 10, 68, 50, 9, 2},
+			expected: []int{2, 9, 10, 48, 50, 53, 59, 62, 68, 86, 100}},
+	}
+
+	for _, testCase := range tests {
+		actual := bubbleSort(testCase.input)
```
anzboi, line 29: In future use the following
```
for name, test := range tests {
    // test code
  })
}
```

This will run each case as a sub test and automatically format the output such that the test name is associated with the output.

anzboi (approved):
LGTM
left a note on nicer table-driven tests for future.

PR #474, opened 2019-06-23
Lab06 - Puppy Storer 

### LINKED ISSUE #473: Lab06 - Puppy Storer
- Constructed puppy type using struct
- Constructed Storer type  interface
- Implemented  Storer interface using map
- Implemented  Storer interface using sync.Map
- Test cases for all CRUD methods both map and sync.Map implementations

willshen8:
Fixes #473 

Review of colleague's PR #463

## - CRUD Puppy

- Constructed puppy type using struct
- Constructed Storer type interface
- Implemented Storer interface using map
- Implemented Storer interface using sync.Map
- Test cases for all CRUD methods both map and sync.Map implementations

### DIFF
--- a/06_puppy/willshen8/mapStore.go
+++ b/06_puppy/willshen8/mapStore.go
@@ -1,0 +1,57 @@
+package main
+
+type MapStore struct {
+	ms      map[uint32]Puppy
+	counter uint32
+}
+
+// NewMapStore initialise a new MapStore
+func NewMapStore() *MapStore {
+	var newMapStore = MapStore{}
+	newMapStore.ms = map[uint32]Puppy{}
+	return &newMapStore
+}
+
+// IncrementCounter increase the ID counter everytime a new Puppy is created.
+func (m *MapStore) IncrementCounter() {
+	m.counter++
+}
+
+// CreatePuppy create a new puppy and store in mapStore.
+func (m *MapStore) CreatePuppy(p *Puppy) uint32 {
+	m.IncrementCounter()
+	p.ID = m.counter
+	m.ms[m.counter] = *p
+	return p.ID
+}
+
+// ReadPuppy read a puppy given its id.
+// It returns the pointer to that puppy.
+func (m *MapStore) ReadPuppy(id uint32) *Puppy {
+	if p, ok := m.ms[id]; ok {
+		return &p
+	}
+	return nil
+}
+
+// UpdatePuppy updates the store with key of id with the new puppy.
+// It returns a boolean whether the operation is successful or not.
+func (m *MapStore) UpdatePuppy(id uint32, puppy *Puppy) bool {
+	// if no existing puppy with this id, return false
+	if _, ok := m.ms[id]; !ok {
+		return false
+	}
+	puppy.ID = id
+	m.ms[id] = *puppy
+	return true
+}
+
+// DeletePuppy delete the puppy given the id.
+// It returns true/success or false/unsuccessful.
+func (m *MapStore) DeletePuppy(id uint32) bool {
+	if _, found := m.ms[id]; found {
+		delete(m.ms, id)
+		return true
+	}
+	return false
+}
--- a/06_puppy/willshen8/puppy.go
+++ b/06_puppy/willshen8/puppy.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type Puppy struct {
+	ID    uint32
+	Breed string
+	Color string
+	Value string
+}
+
+func main() {
+	cutePuppy := Puppy{1, "Jack Russell", "White and Brown", "500"}
+	naughtyPuppy := Puppy{1234, "Fox Terrier", "Black", "1300"}
+	//
+	var store = MapStore{}
+	store.ms = map[uint32]Puppy{}
+	createdID := store.CreatePuppy(&cutePuppy)
+	fmt.Fprintln(out, "First Created Puppy ID:", createdID)
+
+	readPuppy := store.ReadPuppy(createdID)
+	fmt.Fprintln(out, "ReadPuppy:", readPuppy)
+
+	store.UpdatePuppy(createdID, &naughtyPuppy)
+	fmt.Fprintln(out, "Updated Puppy:", store.ReadPuppy(createdID))
+
+	deleteSuccess := store.DeletePuppy(createdID)
+	fmt.Fprintln(out, "Result of deleting puppy:", deleteSuccess)
+}
--- a/06_puppy/willshen8/storer_test.go
+++ b/06_puppy/willshen8/storer_test.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type StorerSuite struct {
+	suite.Suite
+	storer Storer
+	Puppy1 Puppy
+	Puppy2 Puppy
+	Puppy3 Puppy
+	Puppy4 Puppy
+	Puppy5 Puppy
+	Puppy6 Puppy
+	Puppy7 Puppy
+}
+
+func (s *StorerSuite) SetupTest() {
+	s.Puppy1 = Puppy{1, "Jack Russell Terrier", "White and Brown", "1500"}
+	s.Puppy2 = Puppy{1234, "Fox Terrier", "Black", "1300"}
+	s.Puppy3 = Puppy{100, "German Shepperd", "Brown", "2000"}
+	s.Puppy4 = Puppy{120, "Golden Retriever", "Golden", "2500"}
+	s.Puppy5 = Puppy{200, "Chihuahua", "White", "500"}
+	s.Puppy6 = Puppy{300, "Husky", "White", "3500"}
+	s.Puppy7 = Puppy{700, "Pomeranian", "White", "700"}
+}
+
+func (s *StorerSuite) TestCreatePuppySuccessfully() {
+	m := assert.New(s.T())
+	createdID := s.storer.CreatePuppy(&s.Puppy1)
+
+	readPuppy1 := s.storer.ReadPuppy(createdID)
+	m.Equal(&s.Puppy1, readPuppy1)
+}
+
+func (s *StorerSuite) TestCreateNextPuppyWithCorrectID() {
+	m := assert.New(s.T())
+	s.storer.CreatePuppy(&s.Puppy2)
+	// create a second puppy should give an id of 2
+	createdID := s.storer.CreatePuppy(&s.Puppy3)
+	m.Equal(uint32(2), createdID)
+}
+
+func (s *StorerSuite) TestReadPuppySuccessfully() {
+	m := assert.New(s.T())
+	returnedPuppyID4 := s.storer.CreatePuppy(&s.Puppy4)
+	returnPuppy4 := s.storer.ReadPuppy(returnedPuppyID4)
+	m.Equal(&s.Puppy4, returnPuppy4)
+}
+
+func (s *StorerSuite) TestReadNonExistentPuppy() {
+	m := assert.New(s.T())
+	returnPuppy := s.storer.ReadPuppy(100)
+	m.Nil(returnPuppy)
+}
+
+func (s *StorerSuite) TestUpdatePuppySuccessfully() {
+	returnedPuppyID := s.storer.CreatePuppy(&s.Puppy5)
+	updateResult := s.storer.UpdatePuppy(returnedPuppyID, &s.Puppy6)
+	s.Equal(true, updateResult)
+}
+
+func (s *StorerSuite) TestUpdateNonExistentID() {
+	updateResult := s.storer.UpdatePuppy(888, &s.Puppy1)
+	s.Equal(false, updateResult)
+}
+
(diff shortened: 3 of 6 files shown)


06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+type MapStore map[uint32]*Puppy
```
n0npax: you are using pointer to puppy in a map. In my mind it is not secure. I believe I can modify puppy by the reference and skip the `API`.

n0npax: why are you using `uint32` instead of `int`?

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,26 @@
...
+
+func (m MapStore) CreatePuppy(p Puppy) {
```
n0npax: some `codeowners` are asking for documenting exported functions. Not everyone, but some of them.

06_puppy/willshen8/syncStore.go
```diff
@@ -0,0 +1,31 @@
...
+	}
+	return nil
+}
+
+func (s SyncStore) UpdatePuppy(id uint32, puppy *Puppy) {
+	s.syncStore.Store(id, *puppy)
+}
+
+func (s SyncStore) DeletePuppy(id uint32) bool {
+	if _, ok := s.syncStore.Load(id); ok {
```
n0npax, line 54: 👍

06_puppy/willshen8/mapStore_test.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (suite *MapStoreTestSuite) TestReadNonExistentPuppy() {
+	storer := make(MapStore)
+	readPuppy := storer.ReadPuppy(1)
+	suite.Nil(readPuppy)
+}
+
+func (suite *MapStoreTestSuite) TestUpdatePuppy() {
+	var store4 Storer = make(MapStore)
```
n0npax: in my mind name `store` would be bwtter than `store4`. please consider `s/var store[0-9]/var store/g`

06_puppy/willshen8/mapStore.go
n0npax: what if puppy with `id` already exist?

06_puppy/willshen8/syncStore_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type SyncStoreTestSuite struct {
+	suite.Suite
+}
+
+func (suite *SyncStoreTestSuite) SetupTest() {
```
n0npax: I may be missing sth, but i cannot find where this fixture was used?
Maybe you can use puppy created by `SetupTest` in `Delete` and `Read` tests

n0npax (changes requested):

06_puppy/willshen8/mapStore.go
willshen8: Hi, I referred to the [(See hints)](https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/types.go) from the goSlides for lab6. ID field should always be an unsigned int type, as it shouldn't be negative.

06_puppy/willshen8/mapStore.go
willshen8: It will override the old value based on the current implementation. Not sure what should be the ideal scenario here (insert or ignore duplicate key). Perhaps this can be built into lab7 when we deal with errors?

06_puppy/willshen8/syncStore_test.go
willshen8: I played with that method and didn't find any use-cases for it. So left it in there accidentally.

06_puppy/willshen8/mapStore.go
willshen8: I used the sample struct given by the [project hint.](https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/types.go). You have a good point, I'll have to think about the pros and cons of both approaches.

06_puppy/willshen8/mapStore.go
willshen8: Do you mean to document the CRUD methods as comments?

n0npax: yes. Sth like `//CreatePuppy creates puppy` should be fine

06_puppy/willshen8/mapStore.go
willshen8: Thanks, I think you have a valid point.

06_puppy/willshen8/mapStore.go
n0npax: sounds good. 👍

n0npax (dismissed):

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Function which create a new puppy and store in mapStore
+func (m MapStore) CreatePuppy(p Puppy) {
+	m[p.ID] = p
+}
+
+// Function that read a puppy id and return a pointer
+func (m MapStore) ReadPuppy(id uint32) *Puppy {
+	if _, found := m[id]; found {
+		var puppyPointer *Puppy
```
camscale: `puppyPointer` is not a good variable name. It is not idiomatic Go to use types as parts of variable names. But there is also no need for the variable at all. You are looking up the map twice. If you do it once and capture the result the first time, you can do something like this:
```
if p, found := m[id]; found {
        return &p
}
return nil
```

(you could also rename `found` to `ok` which is a lot more common in Go code for this pattern, but `found` is still fine IMHO. Effective Go calls this the "comma ok" idiom: https://golang.org/doc/effective_go.html#maps)

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+type MapStore map[uint32]Puppy
+
+// Function which create a new puppy and store in mapStore
+func (m MapStore) CreatePuppy(p Puppy) {
+	m[p.ID] = p
```
camscale: It does not make much sense for the caller to allocate the ID. They have no idea what are valid IDs. I suggest you maintain a counter and when a Puppy is added to the store, the store sets the ID in the Puppy. This means you may want to pass `Puppy` as a pointer so `Create` can modify the ID of the Puppy passed to it. Or perhaps return an ID instead.

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,34 @@
+package main
+
+type MapStore map[uint32]Puppy
+
+// Function which create a new puppy and store in mapStore
```
camscale: This is not the correct form for doc strings. Have a read of https://blog.golang.org/godoc-documenting-go-code for more detail, but doc strings should always start with the name of the function/method and be complete sentences that end with a full stop. This is real documentation so needs to be written as such.

Please update all the doc strings in your PR to follow the format described in the linked blog post above.

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,34 @@
...
+		puppyRead := m[id]
+		puppyPointer = &puppyRead
+		return puppyPointer
+	}
+	return nil
+}
+
+// Function store that updates the puppy given the id
+func (m MapStore) UpdatePuppy(id uint32, puppy Puppy) {
+	m[id] = puppy
```
camscale: what if the ID in `puppy` is different to `id`? Since `puppy` already has an ID, what is the purpose of the `id` parameter? is it meant to be possible to change the ID of a puppy?

06_puppy/willshen8/mapStore_test.go
```diff
@@ -0,0 +1,66 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
```
camscale: remove unnecessary blank line. annoyingly, `goimports` does not always fix this.

06_puppy/willshen8/syncStore.go
```diff
@@ -0,0 +1,31 @@
+package main
+
+import "sync"
+
+type SyncStore struct {
+	syncStore *sync.Map
```
camscale: You don't have any code that initialises this map, meaning the client must do it. This does not seem like it should be the responsibility of the client. You could add a `NewSyncStore` function that returns an initialised `SyncStore`.

camscale (changes requested):
Some specific comments inline, but also your tests need to be redone. The lab description says:

> Test against the `Storer` interface and run in [suite](https://godoc.org/github.com/stretchr/testify/suite) with both implementations

This means you should have just a single test harness called `storer_test.go`, have your `TestSuite` have a `Storer` field, and in the `TestStorer` function, you create two suites, one with a `MapStore` and another with a `SyncStore` and call `suite.Run()` on each. Your test methods should then use `s.store` to use the `Storer` in the suite to test against.

06_puppy/willshen8/mapStore.go
willshen8: Thanks Cam. So based on the create and update methods, there doesn't seem to be much use of the ID field in puppy then?

willshen8: Does modify the ID of the puppy passed in cause mutation?

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,57 @@
...
+type MapStore struct {
+	ms      map[uint32]Puppy
+	counter uint32
+}
+
+// NewMapStore initialise a new MapStore
+func NewMapStore() *MapStore {
+	var newMapStore = MapStore{}
+	newMapStore.ms = map[uint32]Puppy{}
+	return &newMapStore
```
camscale, line 12: This constructor can be done in one statement:
```
return &MapStore{ms: map[uint32]Puppy{}}
```

This is commonly how simple constructors are done in Go.

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// IncrementCounter increase the ID counter everytime a new Puppy is created.
+func (m *MapStore) IncrementCounter() {
```
camscale, line 16: I'd suggest removing this function. Incrementing an integer is built into the language, and adding this function does not really provide a useful abstraction. Just use `m.nextID++` directly instead of calling this.

A more complicated implementation (with locking perhaps or an external ID generator) would make such a function useful, but I would only add this function when that complicated implementation became necessary.

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,57 @@
...
+	m.counter++
+}
+
+// CreatePuppy create a new puppy and store in mapStore.
+func (m *MapStore) CreatePuppy(p *Puppy) uint32 {
+	m.IncrementCounter()
+	p.ID = m.counter
+	m.ms[m.counter] = *p
+	return p.ID
+}
```
camscale, line 26: Without `IncrementCounter`, this would look like:
```
p.ID = m.nextID++
m.ms[p.ID] = *p
return p.ID
```

Doing it this way will mean your IDs start at 0, not 1 like they did before. I don't see this as a problem, but if you would prefer they start at 1, just initialise `nextID` to ` in `NewMapStore`.

06_puppy/willshen8/mapStore.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+type MapStore struct {
+	ms      map[uint32]Puppy
+	counter uint32
```
camscale, line 5: `counter` could be better named I think. `nextID` makes it clear exactly what it is. `counter` is a little generic.

06_puppy/willshen8/puppy.go
```diff
@@ -0,0 +1,35 @@
...
+type Puppy struct {
+	ID    uint32
+	Breed string
+	Color string
+	Value string
+}
+
+func main() {
+	cutePuppy := Puppy{1, "Jack Russell", "White and Brown", "500"}
+	naughtyPuppy := Puppy{1234, "Fox Terrier", "Black", "1300"}
```
camscale, line 20: :+1: for variable names

06_puppy/willshen8/puppy.go
```diff
@@ -0,0 +1,35 @@
...
+	//
```
camscale, line 21: Remove empty comment. Just use a blank line if you want to separate the sections.

06_puppy/willshen8/storer_test.go
```diff
@@ -0,0 +1,86 @@
...
+	Puppy1 Puppy
+	Puppy2 Puppy
+	Puppy3 Puppy
+	Puppy4 Puppy
+	Puppy5 Puppy
+	Puppy6 Puppy
+	Puppy7 Puppy
+}
+
+func (s *StorerSuite) SetupTest() {
```
camscale, line 22: Good to see the use of SetupTest() to define a test fixture that is consistent across all tests and not accidentally carry across changes from one test to another.
:+1:

camscale (approved):
A couple of comments on mapStore also apply to syncStore, but overall this is a very good implementation of the lab.

I've approved this because it is a quality implementation even though I have some comments. I think my suggestions make for more idiomatic Go, but up to you if you want to continue with this lab or move on. You can make the changes in a later lab if you like.

Sorry for the time taken to review this and thanks for hanging in there.

PR #476, opened 2019-06-24
Lab6 - Puppy Storer - alfredxiao

### LINKED ISSUE #475: Lab06 - Puppy Store

alfredxiao:
Fixes #475 

Review of colleague's PR #436 

 - Implement `Storer` using map
 - Implement `Storer` using sync.map
 - Test using suite


06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+)
+
+type mapStore struct {
```
marksomething, line 9: There's an option here to use promoted fields / embedded types that means you don't need to create constructor functions but there are pros and cons so keep this.

06_puppy/alfredxiao/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+type storerSuite struct {
+	suite.Suite
+	store Storer
+}
+
+func (s *storerSuite) TestCreatePuppyHappyCase() {
+	t := s.T()
+	err := s.store.CreatePuppy(Puppy{ID: "1", Colour: "Black"})
+	assert.NoError(s.T(), err, "Happy case puppy creation")
```
marksomething: You've already grabbed a reference to the `testing.T` on the first line of the function.

marksomething: [suite.Suite](https://godoc.org/github.com/stretchr/testify/suite#Suite) also embeds `assert.Assertions` so you can  call `s.NoError(...)` if you would like

06_puppy/alfredxiao/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+	_ = s.store.CreatePuppy(Puppy{ID: "2"})
+	err := s.store.CreatePuppy(Puppy{ID: "2", Colour: "Red"})
+	assert.Error(s.T(), err, "Puppy creation fails if ID already exists")
+}
+
+func (s *storerSuite) TestReadPuppyHappyCase() {
+	t := s.T()
+	_ = s.store.CreatePuppy(Puppy{ID: "3", Colour: "Blue"})
+	p, err := s.store.ReadPuppy("3")
+	require.NoError(t, err)
```
marksomething: Why the mix of require and assert?

06_puppy/alfredxiao/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (s *storerSuite) TestDeletePuppyNonExisting() {
+	t := s.T()
+	deleted, err := s.store.DeletePuppy("8")
+	require.Error(t, err)
+	assert.Equal(t, false, deleted)
+}
+
+func TestRest(t *testing.T) {
```
marksomething: Is this the right name?`TestRest`

06_puppy/alfredxiao/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+	suite.Run(t, &storerSuite{store: NewMapStore()})
```
marksomething: Will this reuse the same store for every test or create a new store?

06_puppy/alfredxiao/syncstore.go
```diff
@@ -0,0 +1,51 @@
+package main
+
+import (
+	"fmt"
+	"sync"
+)
+
+type syncStore struct {
```
marksomething, line 10: Interesting choice to make this private.

marksomething (dismissed):
Looks really good, nothing i can see that would make me not approve.

06_puppy/alfredxiao/mapstore.go
alfredxiao: surely correct, depending on context

06_puppy/alfredxiao/store_test.go
alfredxiao: Nice, didn't know that

06_puppy/alfredxiao/store_test.go
alfredxiao: require is for not continuing with other assertions when it fails, assertions do not have this attribute. In this case, my thought was that if `err` is `Error` then there is no point in continuing.

06_puppy/alfredxiao/store_test.go
alfredxiao: Oh, copied typo, thanks mate

06_puppy/alfredxiao/store_test.go
alfredxiao: Good point, I better use modify it such that each test uses a new store

alfredxiao: Had a look in your branch, not sure if it creates new store each time there. I couldn't find an easy solution in go yet for this.

alfredxiao: I think I was wrong. SetupTest is the way to go

06_puppy/alfredxiao/main.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	store1 := NewSyncStore()
```
juliaogris: why not `store` ?

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,52 @@
...
+	data map[string]Puppy
+}
+
+func (s *mapStore) CreatePuppy(p Puppy) error {
+	_, ok := s.data[p.ID]
```
juliaogris: Create should probably provision the ID - usually it can't be expected of the clien to know all existing (internal?) IDs.

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,52 @@
...
+func (s *mapStore) ReadPuppy(id string) (Puppy, error) {
+	p, ok := s.data[id]
+	if !ok {
+		return Puppy{}, fmt.Errorf("puppy with ID[%s] does not exists", p.ID)
+	}
+
+	return p, nil
+}
+
+func (s *mapStore) UpdatePuppy(p Puppy) error {
```
juliaogris: imo it's nicer to pull out the ID explicitly and make sure it is the same as p.ID (the REST API in later labs will align with this too):
```go
func (s *mapStore) UpdatePuppy(id string, p Puppy) error {
    if id != p.ID { // return error
```

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,52 @@
...
+	return nil
+}
+
+func (s *mapStore) DeletePuppy(id string) (bool, error) {
+	p, ok := s.data[id]
+	if !ok {
+		return false, fmt.Errorf("puppy with ID[%s] does not exists", p.ID)
+	}
+	delete(s.data, id)
+	return true, nil
```
juliaogris: what's the point of returning `true` and `false` here? `error == nil` and `error != nil` encapsulate the same information.

06_puppy/alfredxiao/store_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (s *storerSuite) SetupTest() {
+	s.store = s.storeFactory()
+}
+
+func (s *storerSuite) TestCreatePuppyHappyCase() {
+	err := s.store.CreatePuppy(Puppy{ID: "1", Colour: "Black"})
+	s.NoError(err, "Happy case puppy creation")
+
+	p, _ := s.store.ReadPuppy("1")
```
juliaogris:
```go
p, err := s.store.ReadPuppy("1")
s.NoError(err)
//....
```
also for all below - don't swallow errors in tests especially when they are against different implementations.

juliaogris (changes requested):
Gotta run now, but rushed, but should give you a start :)

06_puppy/alfredxiao/main.go
alfredxiao: Thought to run both mapstore and syncstore in main, but later decided no need to test both in main.

06_puppy/alfredxiao/mapstore.go
alfredxiao: Depending on how realistic we want to be. If `CreatePuppy` provisions an ID it should probably then return the ID, or the function consumer is not able query the Puppy again that has just been created. I'll update the code to do it this way.

06_puppy/alfredxiao/mapstore.go
alfredxiao: Not fully convinced in this case. The consumer will always call it like `s.UpdatePuppy(puppy.ID, puppy)` then, is it a bit redundant like the `true,nil` and `false,err` case below?

06_puppy/alfredxiao/mapstore.go
alfredxiao: Good point! Copied this from reference without too much consideration. Thought it is easier to find out success or not by bool, rather than check error.

06_puppy/alfredxiao/store_test.go
alfredxiao: This is testing puppy creation, hence swallowing errors in puppy read. Otherwise, the test code can become more lengthy. Not sure which way is better.

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+type mapStore struct {
+	data map[string]Puppy
+}
+
+func (s *mapStore) CreatePuppy(p Puppy) string {
```
juliaogris, line 13: Use [godoc](https://blog.golang.org/godoc-documenting-go-code) for all public functions.

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,58 @@
...
+	p, ok := s.data[id]
+	if !ok {
+		return Puppy{}, fmt.Errorf("puppy with ID[%s] does not exists", p.ID)
+	}
+
+	return p, nil
+}
+
+func (s *mapStore) UpdatePuppy(id string, p Puppy) error {
+	if id != p.ID {
```
juliaogris, line 31: not sure why you'd have separate id field?

06_puppy/alfredxiao/mapstore.go
```diff
@@ -0,0 +1,58 @@
...
+		return fmt.Errorf("bad update request, two IDs (%s, %s) do not match",
+			id, p.ID)
+	}
+
+	_, ok := s.data[id]
+	if !ok {
```
juliaogris, line 37: `if _, ok := s.data[id]; !ok { `

same below.

06_puppy/alfredxiao/syncstore.go
juliaogris: Don't make it private :)

06_puppy/alfredxiao/syncstore.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/google/uuid"
+)
+
+type syncStore struct {
+	data sync.Map
+}
+
+func (s *syncStore) CreatePuppy(p Puppy) string {
+	id := uuid.New().String()
```
juliaogris, line 15: 👍

06_puppy/alfredxiao/syncstore.go
```diff
@@ -0,0 +1,57 @@
...
+		return fmt.Errorf("bad update request, two IDs (%s, %s) do not match",
+			id, p.ID)
+	}
+
+	_, ok := s.data.Load(id)
+	if !ok {
+		return fmt.Errorf("puppy with ID[%s] does not exists", p.ID)
+	}
+
+	s.data.Store(id, p)
```
juliaogris, line 42: load and store need to be guarded with [Mutex](https://golang.org/pkg/sync/#Mutex) or [RWMutex](https://golang.org/pkg/sync/#RWMutex) to avoid race conditions.
dame for delete.

juliaogris (changes requested):
- [Limit the number of commits to one or two](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#limit-the-number-of-commits-to-one-or-two)
- [Fix commit messages](https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments#commit-messages) - Capitalise first word in commit message.

If you need help with this, please reach out and/or come over.

You'll also have to add a mutex to add agains race condition in the syncmap implementation (the point of using sync.Map is to protect your code from race conditions - make it "thread/goroutine safe").

Other than that looks pretty solid, well done.

PR #480, opened 2019-06-24
modules: Change module name to go-course from go-training

### LINKED ISSUE #479: go.mod has old go-training name in the module
The `go.mod`file still has `go-training` as the module name and not `go-course`. This should have been changed when the repository was renamed.

This has caused problems with the later labs that define their own packages - people are rightly using `go-course` in the name of the package, but that is not found at the top-level because of the incorrect name in the top-level `go.mod` file.

camscale:
When the repository was renamed from go-training to go-course, the
go.mod file was not updated to reflect the new name.

Fixes #479

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
-module github.com/anz-bank/go-training
+module github.com/anz-bank/go-course
 
 require github.com/stretchr/testify v1.3.0


juliaogris (approved):
lgtm

PR #478, opened 2019-06-24
Lab3

nicko-lee:
Fixes #477

Review of colleague's PR #457 

Changes proposed in this PR:
* Add letter sort functionality
* Add unit tests based on table driven testing

PR #485, opened 2019-06-24
Puppy Errors

### LINKED ISSUE #484: Lab 7 - Errors
- [x] Create an executable go program
 - [x] Copy the CRUD puppy from upstream master
 - [x] Add a custom error type `Error` with fields `Message` and `Code`
 - [x] Extend the `Storer` interface for all methods to also return `error`
 - [x] Create errors for:
     - [x] Value < 0
     - [x] ID not found in Read, Update and Delete
 - [x] Add locking for proper use of sync.Map

nicholascross:
Fixes #484

Review of colleague's PR #478

 - Create an executable go program
 - Copy the CRUD puppy from upstream master
 - Create errors for:
     - Value < 0

### DIFF
--- a/07_errors/nicholascross/main.go
+++ b/07_errors/nicholascross/main.go
@@ -1,0 +1,169 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+type Puppy struct {
+	ID     int64
+	Breed  string
+	Colour string
+	Value  float64
+}
+
+type Storer interface {
+	CreatePuppy(p Puppy) (bool, error)
+	RetrievePuppy(int64) (*Puppy, error)
+	UpdatePuppy(int64, Puppy) (bool, error)
+	DeletePuppy(int64) (bool, error)
+}
+
+type MapStore map[int64]Puppy
+
+const (
+	missingPup = iota
+	invalidPupValue
+	pupAlreadyExists
+)
+
+type PupError struct {
+	Message string
+	Code    int
+}
+
+func (e *PupError) Error() string {
+	return fmt.Sprintf("[%d] %s", e.Code, e.Message)
+}
+
+func (e *PupError) IsMissingPup() bool {
+	return e.Code == missingPup
+}
+
+func (e *PupError) PupAlreadyExists() bool {
+	return e.Code == pupAlreadyExists
+}
+
+func (e *PupError) InvalidPupValue() bool {
+	return e.Code == invalidPupValue
+}
+
+func missingPuppy(id int64) error {
+	return &PupError{fmt.Sprintf("Puppy not found: %d", id), missingPup}
+}
+
+func puppyAlreadyExists(id int64) error {
+	return &PupError{fmt.Sprintf("Puppy already exists: %d", id), pupAlreadyExists}
+}
+
+func invalidPuppyValue(value float64) error {
+	return &PupError{fmt.Sprintf("Puppy value must be non negative: %f", value), invalidPupValue}
+}
+
+var out io.Writer = os.Stdout
+
+func main() {
+	p := Puppy{1, "Jack Russel Terrier", "white and brown", 550.0}
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	if _, ok := ms[p.ID]; ok {
+		return false, puppyAlreadyExists(p.ID)
+	}
+
+	ms[p.ID] = p
+	return true, nil
+}
+
+func (ms MapStore) RetrievePuppy(id int64) (*Puppy, error) {
+	if _, ok := ms[id]; ok {
+		pup := ms[id]
+		return &pup, nil
+	}
+	return nil, missingPuppy(id)
+}
+
+func (ms MapStore) UpdatePuppy(id int64, p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	if _, ok := ms[id]; ok {
+		p.ID = id
+		ms[p.ID] = p
+		return true, nil
+	}
+	return false, missingPuppy(id)
+}
+
+func (ms MapStore) DeletePuppy(id int64) (bool, error) {
+	if _, ok := ms[id]; ok {
+		delete(ms, id)
+		return true, nil
+	}
+	return false, missingPuppy(id)
+}
+
+type SyncStore struct {
+	store *sync.Map
+	lock  *sync.Mutex
+}
+
+func (s SyncStore) CreatePuppy(p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(p.ID); ok {
+		return false, puppyAlreadyExists(p.ID)
+	}
+
+	s.store.Store(p.ID, p)
+	return true, nil
+}
+
+func (s SyncStore) RetrievePuppy(id int64) (*Puppy, error) {
+	if pup, ok := s.store.Load(id); ok {
+		p := pup.(Puppy)
+		return &p, nil
+	}
+	return nil, missingPuppy(id)
+}
+
+func (s SyncStore) UpdatePuppy(id int64, p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(id); ok {
+		p.ID = id
+		s.store.Store(p.ID, p)
+		return true, nil
+	}
+	return false, missingPuppy(id)
+}
+
+func (s SyncStore) DeletePuppy(id int64) (bool, error) {
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(id); ok {
+		s.store.Delete(id)
+		return true, nil
+	}
+	return false, missingPuppy(id)
+}
--- a/07_errors/nicholascross/main_test.go
+++ b/07_errors/nicholascross/main_test.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"sync"
+	"testing"
+
+	"github.com/stretchr/testify/suite"
+)
+
+type StorerTestSuite struct {
+	suite.Suite
+	storer Storer
+}
+
+func (suite *StorerTestSuite) SetupTest() {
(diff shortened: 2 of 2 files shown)


07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,157 @@
...
+	CreatePuppy(p Puppy) (bool, error)
+	RetrievePuppy(int64) (*Puppy, error)
+	UpdatePuppy(int64, Puppy) (bool, error)
+	DeletePuppy(int64) (bool, error)
+}
+
+type MapStore map[int64]Puppy
+
+const (
+	missingPup      = 0
```
n0npax: explicit numbers 0,1 can be replaced with `iota` here

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,157 @@
...
+func main() {
+	p := Puppy{1, "Jack Russel Terrier", "white and brown", 550.0}
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
```
n0npax, line 76: what if puppy with id `44` exist and p.ID == `44`? I believe you should return `false, AlreadyExist(id)` or sth simillar

n0npax: Anyway, I don't like situation when you are using `id` from request. Someone can accidentally modify some data and it can be pain on the neck in case of requests from distributed sources.

No action Required - decision is up to you

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,157 @@
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	s.store.LoadOrStore(p.ID, p)
+	return true, nil
+}
+
+func (s SyncStore) RetrievePuppy(id int64) (*Puppy, error) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
```
n0npax: I may be wrong here, but seems like operations on puppy are atomic. You don't risk inconsistent state here, so I don't see any reason for locking `ReadPuppy`.

No action required in my mind.

07_errors/nicholascross/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+type StorerTestSuite struct {
+	suite.Suite
+	storer Storer
+}
+
+func (suite *StorerTestSuite) SetupTest() {
+	ok, _ := suite.storer.CreatePuppy(Puppy{1, "Jack Russel", "white", 500.0})
+	if !ok {
```
n0npax: 👍

n0npax (changes requested):
Good work mate!

1 blocking issue.
1 suggestion
1 architectural issue to be decided on what you want to do

07_errors/nicholascross/main.go
nicholascross: Good point, I was treating it as no-op.  I have made this an error instead.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,172 @@
...
+	CreatePuppy(p Puppy) (bool, error)
+	RetrievePuppy(int64) (*Puppy, error)
+	UpdatePuppy(int64, Puppy) (bool, error)
+	DeletePuppy(int64) (bool, error)
+}
+
+type MapStore map[int64]Puppy
+
+const (
+	missingPup      = iota
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	missingPup = iota
```

07_errors/nicholascross/main.go
nicholascross: I think you are right, the reason we need to lock for create, update and delete is because they involve more than one interaction with the underlying data store, so if it was mutated on another thread it could be in an inconsistent state. 👍 

I have updated.

n0npax (approved):
Lgtm

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+func (e *PupError) PupAlreadyExists() bool {
+	return e.Code == pupAlreadyExists
+}
+
+func (e *PupError) InvalidPupValue() bool {
+	return e.Code == invalidPupValue
+}
+
+func missingPuppy(id int64) error {
+	return &PupError{fmt.Sprintf("Puppy not found: %d", id), missingPup}
```
camscale, line 54: Error strings should not start with a capital, as they are often wrapped and this leads to capitals in the middle of the message. See https://github.com/golang/go/wiki/Errors for more details.

This applies to the other two errors too.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+	pupAlreadyExists
+)
+
+type PupError struct {
+	Message string
+	Code    int
+}
+
+func (e *PupError) Error() string {
+	return fmt.Sprintf("[%d] %s", e.Code, e.Message)
```
camscale, line 38: Putting the code into the formatted string doesn't add any useful information because the code is meaningless to humans. The error string contains all the useful information. I'd suggest leaving it out.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	p := Puppy{1, "Jack Russel Terrier", "white and brown", 550.0}
+	fmt.Fprintf(out, "%d - %s [%s]: %f", p.ID, p.Breed, p.Colour, p.Value)
+}
+
+func (ms MapStore) CreatePuppy(p Puppy) (bool, error) {
+	if p.Value < 0 {
```
camscale, line 73: I don't know about dogs, but I'm pretty sure my cat has negative value.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+	Breed  string
+	Colour string
+	Value  float64
+}
+
+type Storer interface {
+	CreatePuppy(p Puppy) (bool, error)
+	RetrievePuppy(int64) (*Puppy, error)
+	UpdatePuppy(int64, Puppy) (bool, error)
+	DeletePuppy(int64) (bool, error)
```
camscale, line 21: Having both a `bool` and `error` return value seems redundant. Whenever the `bool` is `true`, `error` is `nil` and vice versa. You can just return an error, and all is good if it is `nil`.

This applies to `CreatePuppy`, `UpdatePuppy` and `DeletePuppy`

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (s SyncStore) CreatePuppy(p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(p.ID); ok {
```
camscale, line 127: You should be able to use `LoadOrStore` (https://golang.org/pkg/sync/#Map.LoadOrStore) without needing the mutex (and also simpler code):
```
if _, ok := s.LoadOrStore(p.ID, p); ok {
        return false, puppyAlreadyExists(p.ID)
}
return true, nil
```

Even if this is run concurrently with one of the methods that takes the mutex, they do not require the puppy not to exist between the two `sync.Map` operations so `CreatePuppy` done this way does not interfere.

Although as I think about it, perhaps keeping the mutex in this method is better as it is clearer. However, `LoadOrStore` is still a shorter way to implement this.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (s SyncStore) UpdatePuppy(id int64, p Puppy) (bool, error) {
+	if p.Value < 0 {
+		return false, invalidPuppyValue(p.Value)
+	}
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(id); ok {
```
camscale, line 151: Align the happy path to the left edge (https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) by inverting the sense of the conditional:

So,
```
if _, ok := s.store.Load(id); !ok {
        return false, missingPuppy(id)
}
p.ID = id
s.store.Store(p.ID, p)
return true, nil
```

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+	return false, missingPuppy(id)
+}
+
+func (s SyncStore) DeletePuppy(id int64) (bool, error) {
+
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.store.Load(id); ok {
```
camscale, line 164: Similarly, align the happy path to the left:

```
if _, ok := s.store.Load(id); !ok {
        return false, missingPuppy(id)
}
s.store.Delete(id)
return true, nil
```

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+	CreatePuppy(p Puppy) (bool, error)
+	RetrievePuppy(int64) (*Puppy, error)
+	UpdatePuppy(int64, Puppy) (bool, error)
+	DeletePuppy(int64) (bool, error)
+}
+
+type MapStore map[int64]Puppy
+
+const (
+	missingPup = iota
```
camscale, line 27: Using Pup in these error codes and in the error type, but using Puppy everywhere is inconsistent. It would help if you used Puppy everywhere instead of Pup.

07_errors/nicholascross/main.go
```diff
@@ -0,0 +1,169 @@
...
+	invalidPupValue
+	pupAlreadyExists
+)
+
+type PupError struct {
+	Message string
+	Code    int
```
camscale, line 34: Rather than one error type with a code to differentiate the errors, Go tends towards defining different error types for different errors named `FooError` for parameterised errors, or package-level variables that are error values created with `errors.New()` named `ErrFoo` for static errors.

See https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L230 for an example of the former and https://github.com/golang/go/blob/master/src/encoding/gob/decode.go#L19 for an example of the latter.

Or just create them on the fly as needed: https://github.com/golang/go/blob/master/src/database/sql/sql.go#L542

https://github.com/golang/go/wiki/Errors also has a bit more to say about this too.

I think the main thing is that your errors don't look like errors (from a Go perspective). `return nil, missingPuppy(id)` does not say err(or) anywhere which is what I would usually expect to see in Go code.

camscale (changes requested):
I think some things can be simplified a bit and made a bit more idiomatic, but overall I like it. You may also have a couple of changes to incorporate from my lab 6 feedback.

07_errors/nicholascross/main.go
camscale: Ok, so it looks like I didn't read the lab description where it says "Add a custom error type `Error` with fields `Message` and `Code`" - so ignore my comment about using `Code` (or not, as you please). I think my last paragraph in this comment still stands though - the errors don't look like errors, so if you can put "[Ee]rr" in the names somewhere I think it would read better.

juliaogris:

PR #379, opened 2019-06-14
Complete lab 1 fibonnaci

leonarj4:
Fixes #

Review of colleague's PR #

PR #487, opened 2019-06-24
Lab2 Added Bubble and Insertion sort functions

### LINKED ISSUE #486: Lab2 - Bubble Sort
Implements sorting functions using bubble and insertion sort algorithms.
Each function returns a sorted int slice copy of the passed in array.

davidbroughsmyth:
Fixes #486 

Review of colleague's PR #379 

**Changes proposed in this PR:**
- Add Bubble sort and Insertion sort functions
- Include common unit tests for both functions


02_bubble/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,39 @@
...
+		for i := 1; i < n; i++ {
+			if cpy[i-1] > cpy[i] {
+				cpy[i-1], cpy[i] = cpy[i], cpy[i-1]
+			}
+		}
+	}
+	return cpy
+}
+
+func insertionSort(s []int) []int {
```
nicko-lee, line 24: nice work implementing insertion sort as well!

nicko-lee (approved):

02_bubble/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+var tests = map[string]struct {
+	input []int
+	want  []int
+}{
+	"Reverse Sort":            {input: []int{5, 4, 3, 2, 1}, want: []int{1, 2, 3, 4, 5}},
+	"Already Sorted":          {input: []int{1, 2, 3, 4, 5}, want: []int{1, 2, 3, 4, 5}},
+	"All same":                {input: []int{8, 8, 8, 8, 8}, want: []int{8, 8, 8, 8, 8}},
+	"All zeros":               {input: []int{0, 0, 0, 0, 0}, want: []int{0, 0, 0, 0, 0}},
```
camscale, line 27: This test is not really needed since zero is not any special value in the sort functions. It ends up being the same test as "All same". Extra tests that do not test something extra just become a maintenance burden, so best to avoid.

02_bubble/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	"Empty":                   {input: []int{}, want: []int{}},
+	"Negative":                {input: []int{-1, -3, -5, -2, -4}, want: []int{-5, -4, -3, -2, -1}},
+	"Mixed Neg & Pos numbers": {input: []int{4, -1, -3, 0, -2, 3}, want: []int{-3, -2, -1, 0, 3, 4}},
+}
```
camscale, line 31: I think a test with 1 and 2 elements would both be useful. The 1-element test obviously is always sorted but can be a good test that your logic for working with a pair of values is still correct in the presence of a single value. Since 1 element is a boundary condition, a 2-element test is the other side of the boundary so it's worth testing that too - you'll of course need the two elements to be in the wrong order so you can detect if the code mistakenly skips doing the swap.

02_bubble/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,39 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	cpy := make([]int, len(s))
+	copy(cpy, s)
+	for n := len(cpy); n > 1; n-- {
+		for i := 1; i < n; i++ {
```
camscale, line 15: I'd suggest a better name for the loop indicies would be `i` and `j` as you have done in the insertion sort. `n` is often used as a count of some sort and is fixed for the duration of the loop. I was initially confused by the inner loop always counting up to `n` as I would typically expect something called `n` to be the length of the slice your iterating over. My confusion subsided rather quickly once I inspected the code more closely, but that would have been avoided by using `i` and `j`.

camscale (approved):
I've added some comments, but also approved this PR as my comments more related to general software engineering practices and not anything Go-specific, and since the focus of the course is Go, you're good to merge. If you want to revise the PR to address the comments, that's fine - just ping me on slack (camh) when you've re-pushed and I'll re-review right away (assuming I'm not at the pub tonight).

02_bubble/davidbroughsmyth/main_test.go
davidbroughsmyth: Yes I see that now.

davidbroughsmyth:
Thankyou @camh-anz for the review. I will these as learnings for the other labs which require changes and just proceed with the merge.

PR #451, opened 2019-06-20
lab 04 updated, first commit

Marchitect:
Fixes #448 

Review of colleague's PR #424 

#### Changes proposed in this PR:

- added function numeronyms() and corresponding test cases

PR #494, opened 2019-06-25
Implement numeronyms function and write test cases

### LINKED ISSUE #493: Lab 4 - Implement numeronyms function and write test cases

nicko-lee:
Fixes #493 

Review of colleague's PR #451  

Changes proposed in this PR:

* Add numeronym functionality (using variadic function)
* Add unit tests


04_numeronym/nickolee/numeronym_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{name: "big string", input: "deoxyribonucleicacid", expected: "d18d"},
+		{name: "four char string", input: "food", expected: "f2d"},
+		{name: "empty string", input: "", expected: ""},
+		{name: "three char string", input: "sky", expected: "sky"},
+		{name: "two char string", input: "be", expected: "be"},
+		{name: "strings with non alphabets", input: "spider-man", expected: "s8n"},
+		{name: "starting upper case", input: "Batman", expected: "B4n"},
+		{name: "strings with emojis", input: "a👍👍z", expected: "a2z"},
+		{name: "strings with emojis", input: "👍👍👍", expected: "👍👍👍"},
+		{name: "unicode string with empty space", input: "👍👍👍$€₡ 👍", expected: "👍6👍"},
```
willshen8, line 61: Great to see so many test cases. Though as per wikipedia: "a numeronym is a word where a number is used to form an abbreviation" [https://en.wikipedia.org/wiki/Numeronym](url). So I believe your function needs to check whether the input is a word or not. As per wiki again, this word doesn't have to be in English, but can't be emojis. So I would change the implementation a bit.

04_numeronym/nickolee/numeronym_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	tests := []test{
+		{name: "big string", input: "deoxyribonucleicacid", expected: "d18d"},
+		{name: "four char string", input: "food", expected: "f2d"},
+		{name: "empty string", input: "", expected: ""},
+		{name: "three char string", input: "sky", expected: "sky"},
+		{name: "two char string", input: "be", expected: "be"},
+		{name: "strings with non alphabets", input: "spider-man", expected: "s8n"},
+		{name: "starting upper case", input: "Batman", expected: "B4n"},
+		{name: "strings with emojis", input: "a👍👍z", expected: "a2z"},
```
willshen8, line 59: Also consider whether special characters and maybe numbers should be considered, e.g. if we convert "123456789" into "179"? Numeronyms should technically convert a word into a three character string.

04_numeronym/nickolee/numeronym_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{name: "strings with emojis", input: "👍👍👍", expected: "👍👍👍"},
+		{name: "unicode string with empty space", input: "👍👍👍$€₡ 👍", expected: "👍6👍"},
+	}
+
+	for _, testCase := range tests {
+		actual := numeronym(testCase.input)
+		fmt.Println(testCase.name, "expected: ", testCase.expected)
+		fmt.Println(testCase.name, "actual: ", actual)
```
willshen8: This is minor and no need to change if you don't feel like it, and you could print both onto the same line. And perhaps you don't need to know the expected result, if the test cases fail, it will print it to show you which one failed.

04_numeronym/nickolee/numeronym.go
```diff
@@ -0,0 +1,36 @@
...
+
+	firstChar := string(runes[0]) // casting a byte (one character) into a string
+	lastChar := string(runes[len(runes)-1])
+	middleChars := strconv.Itoa(len(runes) - 2)
+	return firstChar + middleChars + lastChar
+}
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+
```
willshen8: Remove empty line please.

04_numeronym/nickolee/numeronym.go
```diff
@@ -0,0 +1,36 @@
...
+	result := make([]string, len(vals))
+	for i, val := range vals {
+		result[i] = numeronym(val) // build results slice
+	}
+	return result
+}
+
+func numeronym(s string) string {
+	// convert string to slice of runes (Unicode int32) to avoid some kind of "protocol error"
+	runes := []rune(s)
```
willshen8, line 22: I am not sure what's happening here, the expected input is string, not runes. Perhaps I am missing something here.

04_numeronym/nickolee/numeronym_test.go
```diff
@@ -0,0 +1,76 @@
...
+		if actual != testCase.expected {
+			t.Fatalf("test case: %s failed. expected: %v, got: %v", testCase.name, testCase.expected, actual)
+		}
+	}
+}
```
willshen8, line 71: This is a suggestion only as some people say they don't like to use external test library. But I think you could essentially shorter the 3 lines of code into one `assert.Equal(t, test.expected, actual)` using test library `"github.com/stretchr/testify/assert"`.

willshen8 (dismissed):
Well done, just some minor adjustments and some considerations for the test cases. Might need to think about the implementations as per proper use of numeronyms.

04_numeronym/nickolee/numeronym.go
nicko-lee: Removed

04_numeronym/nickolee/numeronym_test.go
nicko-lee: Removed print statements

04_numeronym/nickolee/numeronym.go
nicko-lee: Without doing this - the code will not work as expected for Unicode characters for example emojis 😅😌😡😶🤤

04_numeronym/nickolee/numeronym_test.go
nicko-lee: ok added thanks for pointing out

anzboi (approved):
LGTM

PR #497, opened 2019-06-25
Allow duplicated code in golangci conf

### LINKED ISSUE #496: golangci should allow deduplicated code
some source code will be exavtly the same in lab08,09,10 and 11.
`golangci` should allow this.

n0npax:
Fixes #496 

Review of colleague's PR # N/A

##
Allow deduplicated code in golangci conf


camscale (changes requested):
I'm going to be strict on this because you're at the top of the leaderboard :)

The commit message is too short and really doesn't say anything the change itself is not already saying. It is important to add the "why" to the commit message. Six months down the track when someone else is running the course has a look at this and wants duplicate checking turned on for their own reason, they do a `git blame` to see why it was turned off. When they read your commit message, they are none-the-wiser so they just remove it.... and then hit the same problem later.

You want to say something like:

-----
Turn off duplicate checking because labs 06 - 11 can have the same files in them because they progressively build on the previous lab. With duplicate checking enabled, travis checks fail and the later lab PRs cannot be merged.

The following error triggered this change:
```
08_project/n0npax/pkg/puppy/store/store_test.go:1: 1-216 lines are duplicate of `09_json/n0npax/pkg/puppy/store/store_test.go:1-216` (dupl)
```
-----

Do not worry about being too verbose. You can see I even added the error message that prompted you to make this change. That can sometimes be useful info for your future self too, so stick it in.

All of this info I got from what you posted to the channel. If you needed to explain it there, you need to explain it in the commit message.

You'll need to do a `git commit --amend` on your branch to edit the commit message and then force push.

n0npax:
new better commit msg:

```
Author: marcin.niemira <marcin.niemira@anz.com>
Date:   Wed Jun 26 09:10:24 2019 +1000

    Allow duplicated code in golangci conf

    Turn off duplicate checking because labs 06 - 11 can have the same files,
    because they progressively build on the previous lab.
    With duplicate checking enabled, travis checks fail.
```

camscale (approved):
LGTM

PR #515, opened 2019-06-30
Implement Lab3 Letter frequency

### LINKED ISSUE #516: Lab 3:Letter frequency
Lab3 letter frequency

- Create an executable go program in directory `03_letters/sirigithub'
- Implement a function letters(s string)  that returns a map[rune]int of each letter
- and its frequency
- Implement a function sortLetters(m map[rune]int) that returns sorted slice of strings in the format "letter:frequency"
- Include tests for different string inputs.
- Implemented fixes as per PR review

sirigithub:
Fixes #516 

Review of colleague's PR #

- Implemented a function letters(s string)  that returns a map[rune]int of each letter
- and its frequency
- Implemented a function sortLetters(m map[rune]int) that returns sorted slice of strings in the format "letter:frequency"
- Implemented main method that calls fmt.Println(strings.Join(sortLetters(letters("aba")), "\n")) to print 
- each of the "letter:frequency" in a new line
- Included tests for all functions with different string inputs.
- Implemented fixes as per PR review
- Pushed missing fixes from earlier PR review suggestions.


03_letters/sirigithub/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLetters(t *testing.T) {
+	tests := []struct {
+		description string
+		input       string
+		expected    []string
+	}{
+		{description: "Simple string", input: "thisisasimplestring", expected: []string{"a:1", "e:1", "g:1", "h:1", "i:4", "l:1", "m:1", "n:1", "p:1", "r:1", "s:4", "t:2"}},
```
golangcibot: line is 167 characters (from `lll`)

undrewb (dismissed):
Consider using short variable names - e.g. characterFreq could be cFreq or even cf. http://doc.cat-v.org/bell_labs/pikestyle and https://www.reddit.com/r/golang/comments/8wxwgv/why_does_go_encourage_short_variable_names/

Consider having a seperate test for letters and renaming TestLetterFrequency to TestSortLetterFrequency

03_letters/sirigithub/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{description: "German Umlauts", input: "äöß€’üüüöäßß", expected: []string{"ß:3", "ä:2", "ö:2", "ü:3", "’:1", "€:1"}},
+
+		{description: "String with Back slashes", input: "a\\c\\b", expected: []string{"\\:2", "a:1", "b:1", "c:1"}},
+
+		{description: "String with Emojis", input: "😄🐷🙈🐷🏃😄", expected: []string{"🏃:1", "🐷:2", "😄:2", "🙈:1"}},
+	}
+
+	for _, test := range tests {
+		actual := sortLetters(letters(test.input))
+		fmt.Println(actual)
```
camscale: It's best not to print things out in tests. On a large program, if tests were to print out stuff you'd end up with a mess of output, particularly as tests can run concurrently - all the output would be interleaved.

Also, in the case of this repo, once this is merged to master, this test is going to run every time and this output will be in the common travis output. So best to remove this before merging.

It can be handy having these prints there when writing/growing your tests, but please be sure to remove it when you push your commits.

03_letters/sirigithub/main.go
```diff
@@ -0,0 +1,37 @@
...
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+// letters returns a map of rune literals and its frequency in the given input string s
+func letters(s string) map[rune]int {
+	charecterFreq := make(map[rune]int)
```
camscale: A minor comment, but "character" is spelt incorrectly here. However I'd suggest not using the word "character" at all. Go consciously avoided using it - using the word "rune" instead - because character has a history of mixed meaning - is it one character in a string, or an 8-bit value? Rune is the former, but not the latter. Go uses `byte` for the latter.

I'd suggest just using `r` for the range variable: `for _, r := range s` reads quite naturally - `s` is a string, `r` is a rune and the scope is very small so no problem with the short variable names.

Using `runeFreq` for the map name would also make sense here. The `Freq` naming is good - it makes it quite clear what is going on.

03_letters/sirigithub/main.go
```diff
@@ -0,0 +1,37 @@
...
+	// increment the counter if the charecter exists
```
camscale: I'd suggest removing this comment, as it doesn't say any more than the code already says.

camscale (changes requested):
This is pretty close to perfect, just a couple of minor comments that I think are worth addressing. Ping me when you've pushed the changes and I'll approve.

03_letters/sirigithub/main.go
sirigithub: That was an oversight, thanks Cam . Done

03_letters/sirigithub/main.go
sirigithub: Great points ! I suppose in my head  I was still thinking character with runes.
Done.

03_letters/sirigithub/main.go
sirigithub: done.

camscale: hmmm, the comment is still there. forgot to push?

camscale (dismissed):
Just one small change from the previous review that was missed, and you're good to go.

03_letters/sirigithub/main.go
sirigithub: oops again yes. Definitely fixed this time.

sirigithub: oops again . Definitely fixed this time.

davidbroughsmyth:
Just notice that you have not added in your PR comments:

> Review of colleague's PR #

But otherwise LGTM

juliaogris (approved):
LGTM

PR #520, opened 2019-06-30
Lab3: Implement letter frequency solution and tests

### LINKED ISSUE #519: Lab3 - Letter frequency (alanhillgemann)

alanhillgemann:
Fixes #519 

Review of colleague's PR #tba

### DIFF
--- a/03_letters/alanhillgemann/main.go
+++ b/03_letters/alanhillgemann/main.go
@@ -15,25 +15,18 @@ func main() {
 }
 
 func letters(s string) map[rune]int {
-	runeMap := make(map[rune]int)
-	runeSlice := []rune(s)
-	for i := 0; i <= len(runeSlice)-1; i++ {
-		runeMap[runeSlice[i]]++
+	letterTotalMap := map[rune]int{}
+	for _, r := range s {
+		letterTotalMap[r]++
 	}
-	return runeMap
+	return letterTotalMap
 }
 
 func sortLetters(m map[rune]int) []string {
-	stringSlice := make([]string, 0)
-	mapKeys := make([]string, 0)
-	mapValues := make([]int, 0)
+	letterTotalStrings := []string{}
 	for k, v := range m {
-		mapKeys = append(mapKeys, string(k))
-		mapValues = append(mapValues, v)
+		letterTotalStrings = append(letterTotalStrings, fmt.Sprintf("%s:%d", string(k), v))
 	}
-	for i := 0; i <= len(m)-1; i++ {
-		stringSlice = append(stringSlice, fmt.Sprintf("%s:%d", mapKeys[i], mapValues[i]))
-	}
-	sort.Strings(stringSlice)
-	return stringSlice
+	sort.Strings(letterTotalStrings)
+	return letterTotalStrings
 }
--- a/03_letters/alanhillgemann/main.go
+++ b/03_letters/alanhillgemann/main.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	runeMap := make(map[rune]int)
+	runeSlice := []rune(s)
+	for i := 0; i <= len(runeSlice)-1; i++ {
+		runeMap[runeSlice[i]]++
+	}
+	return runeMap
+}
+
+func sortLetters(m map[rune]int) []string {
+	stringSlice := make([]string, 0)
+	mapKeys := make([]string, 0)
+	mapValues := make([]int, 0)
+	for k, v := range m {
+		mapKeys = append(mapKeys, string(k))
+		mapValues = append(mapValues, v)
+	}
+	for i := 0; i <= len(m)-1; i++ {
+		stringSlice = append(stringSlice, fmt.Sprintf("%s:%d", mapKeys[i], mapValues[i]))
+	}
+	sort.Strings(stringSlice)
+	return stringSlice
+}
--- a/03_letters/alanhillgemann/main_test.go
+++ b/03_letters/alanhillgemann/main_test.go
@@ -1,0 +1,62 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


03_letters/alanhillgemann/main.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	runeMap := make(map[rune]int)
```
anzdaddy: I prefer to avoid `make(…)` except in cases where it's required. In this case you can write:
```go
runeMap := map[rune]int{}
```

03_letters/alanhillgemann/main.go
```diff
@@ -0,0 +1,39 @@
...
+	runeSlice := []rune(s)
+	for i := 0; i <= len(runeSlice)-1; i++ {
```
anzdaddy: You can simplify the end-test: `i < len(runeSlice)`. However, you can go further and use range-loop to simplify and avoid the slice copy:
```go
for _, r := range s {
	runeMap[r]++
}
```
This works because Go's range-loop has a special case that iterates over the decoded runes of a string instead of its bytes. See note 2 under https://golang.org/ref/spec#For_range.

03_letters/alanhillgemann/main.go
```diff
@@ -0,0 +1,39 @@
...
+		runeMap[runeSlice[i]]++
+	}
+	return runeMap
+}
+
+func sortLetters(m map[rune]int) []string {
+	stringSlice := make([]string, 0)
```
anzdaddy:
```go
stringSlice := []string{}
```
Same for `mapKeys` and `mapValues`.

03_letters/alanhillgemann/main.go
```diff
@@ -0,0 +1,39 @@
...
+	mapKeys := make([]string, 0)
+	mapValues := make([]int, 0)
+	for k, v := range m {
+		mapKeys = append(mapKeys, string(k))
+		mapValues = append(mapValues, v)
+	}
+	for i := 0; i <= len(m)-1; i++ {
```
anzdaddy: Use `<` to avoid the `-1`.

03_letters/alanhillgemann/main.go
anzdaddy: Name things by what they're used for, not what they are. In this case, `keyValueStrings` might suffice.

(**Hot tip:** In VSCode, you can press <kbd>F2</kbd> to edit a name and all of its references.)

anzdaddy: Move `stringSlice` (or whatever you decide to call it) to just before the loop that fills it.

anzdaddy (changes requested):

03_letters/alanhillgemann/main.go
alanhillgemann: Fixed

03_letters/alanhillgemann/main.go
alanhillgemann: Fixed

03_letters/alanhillgemann/main.go
alanhillgemann: Fixed. All variables renamed to be more descriptive

alanhillgemann: Fixed

alanhillgemann: Fixed

03_letters/alanhillgemann/main.go
alanhillgemann: Fixed. For statement consolidated into parents for range loop to make the function simpler.

alanhillgemann:
@anzdaddy thank you for reviewing my PR. I have made all changes requested.

03_letters/alanhillgemann/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := strconv.Quote("a:2\nb:1\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equalf(t, expected, actual, "expected %v, actual %v", expected, actual)
+}
+
+func TestLetters(t *testing.T) {
+	testCases := map[string]struct {
```
anzdaddy, line 21: Use of a map makes the order of tests non-deterministic. Not a problem, _per se_, just something to be aware of.

03_letters/alanhillgemann/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+		"2 same":    {input: "aa", expected: map[rune]int{97: 2}},
+		"utf8 char": {input: "£a£0", expected: map[rune]int{48: 1, 97: 1, 163: 2}},
+	}
+
+	for name, test := range testCases {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			actual := letters(test.input)
+			expected := test.expected
+			assert.Equalf(t, expected, actual, "expected %v, actual %v", expected, actual)
```
anzdaddy, line 37: Don't bother with an output message unless it's required to narrow down the context of the assertion. For example:
```go
for i := 0; i < 1000; i++ {
	assert.Equal(t, 42*i, f(i), "%d", i)
}
```

03_letters/alanhillgemann/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+		})
+	}
```
anzdaddy, line 39: (Just FYI) I've noticed this is a common pattern: define a bunch of test cases in a local struct, then loop over them. I much prefer doing the reverse, as it results in significantly less cruft:
```go
test := func(name string, expected map[rune]int, input string) {
		assert.Equal(t, expected, letters(input))
	})
}
test("No params", map[rune]int{}), "")
test("1 char", map[rune]int{97: 1}), "a")
test("2 chars", map[rune]int{48: 1, 97: 1}), "a0")
test("2 same", map[rune]int{97: 2}), "aa")
test("utf8 char", map[rune]int{48: 1, 97: 1, 163: 2}), "£a£0")
```
(Also note the convention of putting the expected value first. I'm personally not a fan of this direction, but it's fairly well established throughout the testing world, not just `github.com/stretchr/testify`.)

anzdaddy (approved):
All good, with some optional suggestions.

PR #522, opened 2019-07-01
Lab6- Add PuppyStore and tests

### LINKED ISSUE #521: Lab 6 - CRUD puppy with interface
- Create an executable go program in directory `06_puppy/USERNAME` (see [hints](https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/types.go))
- Create `Storer` interface with [crud](https://en.wikipedia.org/wiki/Create,\_read,\_update_and_delete) methods for `Puppy`
- Write a `MapStore` implementation of `Storer` backed by a `map`
- Write a `SyncStore` implementation of `Storer` backed by a [sync.Map](https://golang.org/pkg/sync/#Map)
- Keep all implementation files in the same folder and in package `main`
- Test against the `Storer` interface and run in [suite](https://godoc.org/github.com/stretchr/testify/suite) with both implementations

mohitnag:
Fixes #521

Review of colleague's PR #386 

Add CRUD puppy with interface


06_puppy/mohitnag/main.go
```diff
@@ -0,0 +1,17 @@
...
+
+var out io.Writer = os.Stdout
+
+func main() {
+	m := MemStore{}
+	puppy := Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "$2"}
+	_ = m.CreatePuppy(puppy)
+	puppy, _ = m.ReadPuppy(1)
+	fmt.Fprint(out, puppy.Colour)
+}
```
pfannerj, line 34: I don't think it's essential according to the lab requirements but I think it would be good to execute both the sync and the map store functionality from the main... also might be worth initiating the updates and deletes too.

06_puppy/mohitnag/syncstore.go
```diff
@@ -0,0 +1,40 @@
+package main
+
+import (
+	"fmt"
+)
+
+// CreatePuppy creates a Puppy in memstore
```
pfannerj: The comments in syncstore should be updated (ie. they all say 'memstore')... I also saw a comment from Cam which suggested looking at the following link for some details on godoc comments (https://blog.golang.org/godoc-documenting-go-code)

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,126 @@
...
+	assert := assert.New(s.T())
+	puppy := Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "$2"}
+	testCases := []struct {
+		Scenario      string
+		Input         Puppy
+		ExpectedError string
+	}{
+		{"Create Puppy", puppy, "<nil>"},
+		{"Creating already existing Puppy should fail", puppy, "puppy with Id 1 already exists"},
+	}
```
pfannerj, line 37: I like this way of setting up the tests where there is one test function for each function being tested with multiple 'Scenario's setup.  I have separate test functions for success and fail... hopefully the other reviewers will like it too...

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,126 @@
...
+			err := s.store.CreatePuppy(tc.Input)
+			assert.Equal(tc.ExpectedError, fmt.Sprint(err))
+
+		})
+	}
+}
+
+func (s *storer) TestReadPuppy() {
+	assert := assert.New(s.T())
+	puppy := Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "$2"}
```
pfannerj: I think the puppy objects that u create within each test function could be created once and reused...

pfannerj:
All my comments are minor and optional if u want to address them or not...

06_puppy/mohitnag/types.go
```diff
@@ -0,0 +1,27 @@
...
+// Puppy defines the data structure corresponding to a Puppy
+type Puppy struct {
+	ID     uint32
+	Breed  string
+	Colour string
+	Value  string
+}
+
+// MemStore stores Puppy details with Puppy Id as Key and Puppy  as value
+type MemStore map[uint32]Puppy
```
juliaogris: IMO this belongs in the memstore.go file.

06_puppy/mohitnag/memstore.go
```diff
@@ -0,0 +1,38 @@
+package main
+
+import "fmt"
+
+// CreatePuppy creates a Puppy in memstore
+func (m *MemStore) CreatePuppy(p Puppy) error {
+	if _, ok := (*m)[p.ID]; !ok {
```
juliaogris: MemStore is a map, so it already is an implicit pointer.
I think you should be able to write all of these functions as:
```go
func (m MemStore) CreatePuppy(p Puppy) error {
```
etc. (notice the value receiver instead of the pointer receiver in the method?
I'm not 100% sure if this still satisfies the interface though - gotta try ;)
If it does, then you can get rid of all the `(*m)[id]` and us `m[id]` instead which to me is more readable.

06_puppy/mohitnag/memstore.go
```diff
@@ -0,0 +1,38 @@
...
+		(*m)[p.ID] = p
+		return nil
+	}
+	return fmt.Errorf("puppy with Id %d already exists", p.ID)
+}
+
+// ReadPuppy reads a Puppy from memstore
+func (m *MemStore) ReadPuppy(id uint32) (Puppy, error) {
+	if _, ok := (*m)[id]; !ok {
```
juliaogris: No need to read twice:
```go
puppy, ok := (*m)[id]
if !ok { 
   // return err
}
return puppy, nil
```

06_puppy/mohitnag/memstore.go
```diff
@@ -0,0 +1,38 @@
...
+func (m *MemStore) UpdatePuppy(p Puppy) error {
+	if _, ok := (*m)[p.ID]; !ok {
+		return fmt.Errorf("puppy with Id %d does not exists", p.ID)
+	}
+	(*m)[p.ID] = p
+	return nil
+}
+
+// DeletePuppy deletes a Puppy
+func (m *MemStore) DeletePuppy(id uint32) bool {
```
juliaogris: better use error as return val than bool. if your datastore is backed by a database or similar any operation can error (connection timeout etc.).
needs to be fixed in interface as well.

06_puppy/mohitnag/syncstore.go
juliaogris: Please fix.

juliaogris (changes requested):

06_puppy/mohitnag/syncstore.go
```diff
@@ -0,0 +1,40 @@
...
+func (s *SyncStore) ReadPuppy(id uint32) (Puppy, error) {
+	if puppy, ok := s.Load(id); ok {
+		return puppy.(Puppy), nil
+	}
+	return Puppy{}, fmt.Errorf("puppy with Id %d does not exists", id)
+}
+
+// UpdatePuppy updates a Puppy
+func (s *SyncStore) UpdatePuppy(p Puppy) error {
+	if _, ok := s.Load(p.ID); !ok {
```
juliaogris, line 26: oops I missed this - this function should be guarded by a Mutex to avoid race conditions as you are accessing your store twice in an atomic fashion. Same for the delete.

06_puppy/mohitnag/main.go
mohitnag: Done. initiated all the CRUD operations  in main.

06_puppy/mohitnag/syncstore.go
mohitnag: Thanks for pointing out. Fixed

06_puppy/mohitnag/store_test.go
mohitnag: 👍

06_puppy/mohitnag/store_test.go
mohitnag: Done. creating puppy object in test setup.

06_puppy/mohitnag/types.go
mohitnag: I think all the types are under one file i.e 'types.go' , which could be more readable for some. However not sure which is the best way.

06_puppy/mohitnag/memstore.go
mohitnag: Yes. No need of pointer receiver. Removed from everywhere

06_puppy/mohitnag/memstore.go
mohitnag: Done

06_puppy/mohitnag/syncstore.go
mohitnag: I think the Locks are part of lab 7. and i have put locks in each operation in there. Can do the same in lab6 as well if you want me to do

06_puppy/mohitnag/mapstore.go
```diff
@@ -0,0 +1,42 @@
+package main
+
+import "fmt"
+
+// CreatePuppy creates a Puppy in mapstore
+func (m MapStore) CreatePuppy(p Puppy) error {
+	_, ok := m[p.ID]
+	if !ok {
```
juliaogris: Idiomatic go `if _, ok := m[p.ID]; !ok {` - below as well.

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+		ExpectedError string
+	}{
+		{"Delete already existing Puppy", 1, "<nil>"},
+		{"Delete a non-existing Puppy should fail", 32, "puppy with Id 32 does not exists"},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		s.T().Run(tc.Scenario, func(t *testing.T) {
+			err := s.store.DeletePuppy(tc.ID)
+			assert.Equal(tc.ExpectedError, fmt.Sprint(err))
```
juliaogris: err.Error() instead of fmt.Sprint(err) - also, bit dangerous to do text comparison.
consider using assert.Contains(err.Error(), "Delete")
below as well.

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type storer struct {
+	suite.Suite
+	store      Storer
+	storerType func() Storer
```
juliaogris: better name imo: `makeStorer` or ` newStorer`

06_puppy/mohitnag/syncstore.go
```diff
@@ -0,0 +1,41 @@
...
+func (s *SyncStore) ReadPuppy(id uint32) (Puppy, error) {
+	if puppy, ok := s.Load(id); ok {
+		return puppy.(Puppy), nil
+	}
+	return Puppy{}, fmt.Errorf("puppy with Id %d does not exists", id)
+}
+
+// UpdatePuppy updates a Puppy in syncstore
+func (s *SyncStore) UpdatePuppy(p Puppy) error {
+
```
juliaogris: unnecessary blank line.

06_puppy/mohitnag/syncstore.go
juliaogris: ok. it was even me who set up this requirement :( (doesn't make sense to me any more).
leave as is.

juliaogris (changes requested):
a few more little nits. Nearly there.

06_puppy/mohitnag/store_test.go
mohitnag: Well, I tried to use `err.Error()` but the it panics in case of `nil` error. I can do some changes to avoid panic like this 
`err := s.store.DeletePuppy(tc.ID)
			if tc.ExpectedError == nil{
				assert.NoError(err)
			}else{
				assert.Contains(tc.ExpectedError.Error(), err.Error())
			}`
What do you think?

06_puppy/mohitnag/store_test.go
mohitnag: Done

06_puppy/mohitnag/syncstore.go
mohitnag: Lint dont allows me to delete. :|

rokane: I don't see why you can't delete this line when all the other functions have this removed?

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type storer struct {
```
rokane: I would call this a different name. It can make it a little confusing if you already have a `Storer` type, and then a `storer` type. 

Since this is encapsulating details about the test suite, you could call this `storerSuite`?

06_puppy/mohitnag/store_test.go
```diff
@@ -0,0 +1,122 @@
...
+			assert.Equal(tc.Expected, puppy)
+			assert.Equal(tc.ExpectedError, fmt.Sprint(err))
+		})
+	}
+
+}
+
+func (s *storer) TestUpdatePuppy() {
+	assert := assert.New(s.T())
+	err := s.store.CreatePuppy(s.puppy)
```
rokane: See how on all but one of these tests your first step is to create a new puppy based off the puppy stored in the storer. What you could do here is move this logic into the `SetupTest` function. This will remove the need to duplicate the logic across these three tests, as well as remove the need to store the puppy inside the store.

rokane (dismissed):
Great work Mohit, I just have a couple of really minor suggestions. If you could apply them and let me know I will be happy with this :)

06_puppy/mohitnag/syncstore.go
mohitnag: Sorry i got confused with some other line. Removed now

06_puppy/mohitnag/store_test.go
mohitnag: done

06_puppy/mohitnag/store_test.go
mohitnag: Done

juliaogris (approved):
LGTM

PR #531, opened 2019-07-02
Lab07 - Errors

### LINKED ISSUE #530: Lab07 - Errors
**Create errors for:**
- Value < 0

willshen8:
Fixes #530

Review of colleague's PR #533

## Creating error conditions for the following two conditions
- Value < 0
Also adding locking for cater for race conditions

### DIFF
--- a/07_errors/willshen8/error.go
+++ b/07_errors/willshen8/error.go
@@ -1,0 +1,13 @@
+package main
+
+import "fmt"
+
+type Error struct {
+	Message string
+	Code    int
+}
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("Error code %d: %s",
+		e.Code, e.Message)
+}
--- a/07_errors/willshen8/error.go
+++ b/07_errors/willshen8/error.go
@@ -11,3 +11,8 @@ func (e *Error) Error() string {
 	return fmt.Sprintf("Error code %d: %s",
 		e.Code, e.Message)
 }
+
+const (
+	NegativeValue    = 1001
+	NonExistentPuppy = 1002
+)
--- a/07_errors/willshen8/puppy.go
+++ b/07_errors/willshen8/puppy.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type Puppy struct {
+	ID    uint32
+	Breed string
+	Color string
+	Value string
+}
+
+func main() {
+	cutePuppy := Puppy{1, "Jack Russell", "White and Brown", "500"}
+	naughtyPuppy := Puppy{1234, "Fox Terrier", "Black", "1300"}
+
+	var store = MapStore{}
+	store.ms = map[uint32]Puppy{}
+	createdID, _ := store.CreatePuppy(&cutePuppy)
+	fmt.Fprintln(out, "First Created Puppy ID:", createdID)
+
+	readPuppy, _ := store.ReadPuppy(createdID)
+	fmt.Fprintln(out, "ReadPuppy:", readPuppy)
+
+	updateResult, _ := store.UpdatePuppy(createdID, &naughtyPuppy)
+	fmt.Fprintln(out, "Update puppy result:", updateResult)
+
+	deleteSuccess, _ := store.DeletePuppy(createdID)
+	fmt.Fprintln(out, "Result of deleting puppy:", deleteSuccess)
+}
--- a/07_errors/willshen8/storer_test.go
+++ b/07_errors/willshen8/storer_test.go
@@ -9,107 +9,84 @@ import (
 	"github.com/stretchr/testify/suite"
 )
 
+var (
+	Puppy1        = Puppy{1, "Jack Russell Terrier", "White and Brown", "1500"}
+	Puppy2        = Puppy{1234, "Fox Terrier", "Black", "1300"}
+	Puppy3        = Puppy{100, "German Shepperd", "Brown", "2000"}
+	Puppy4        = Puppy{120, "Golden Retriever", "Golden", "2500"}
+	Puppy5        = Puppy{200, "Chihuahua", "White", "500"}
+	Puppy6        = Puppy{300, "Husky", "White", "3500"}
+	Puppy7        = Puppy{700, "Pomeranian", "White", "700"}
+	InvalidPuppy  = Puppy{22, "Pomeranian", "White", "-100"}
+	ReadError     = Error{Code: NonExistentPuppy, Message: "Puppy ID can not be found, read operation failed."}
+	UpdateError   = Error{Code: NonExistentPuppy, Message: "Puppy ID can not be found, update operation failed."}
+	DeleteError   = Error{Code: NonExistentPuppy, Message: "Puppy ID can not be found, delete operation failed."}
+	ErrorCreateID = uint32(0)
+)
+
 type StorerSuite struct {
 	suite.Suite
-	storer        Storer
-	Puppy1        Puppy
-	Puppy2        Puppy
-	Puppy3        Puppy
-	Puppy4        Puppy
-	Puppy5        Puppy
-	Puppy6        Puppy
-	Puppy7        Puppy
-	InvalidPuppy  Puppy
-	ErrorCreateID uint32
-	ErrorCreate   string
-	ErrorRead     string
-	ErrorUpdate   string
-	ErrorDelete   string
-}
-
-func (s *StorerSuite) SetupTest() {
-	s.Puppy1 = Puppy{1, "Jack Russell Terrier", "White and Brown", "1500"}
-	s.Puppy2 = Puppy{1234, "Fox Terrier", "Black", "1300"}
-	s.Puppy3 = Puppy{100, "German Shepperd", "Brown", "2000"}
-	s.Puppy4 = Puppy{120, "Golden Retriever", "Golden", "2500"}
-	s.Puppy5 = Puppy{200, "Chihuahua", "White", "500"}
-	s.Puppy6 = Puppy{300, "Husky", "White", "3500"}
-	s.Puppy7 = Puppy{700, "Pomeranian", "White", "700"}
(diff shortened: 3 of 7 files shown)


07_errors/willshen8/syncStore.go
```diff
@@ -0,0 +1,79 @@
...
+func (s *SyncStore) CreatePuppy(p *Puppy) (uint32, error) {
+	if i, err := strconv.Atoi(p.Value); err == nil {
+		if i < 0 {
+			return 0, &Error{
+				Message: "Puppy value can't be less than 0.",
+				Code:    1,
+			}
+		}
+	}
+	s.mu.Lock()
```
pfannerj: The example provided by Dan for lab6 uses `s.mu.Lock()` followed directly by `defer s.mu.Unlock()` at the beginning of the each applicable function.  I believe using defer is good practice because then it doesn't matter where the function exits, the defer means it will always unlock before the function exits.

07_errors/willshen8/syncStore.go
```diff
@@ -0,0 +1,79 @@
...
+		Message: "Puppy ID can not be found, read operation failed.",
+		Code:    2,
+	}
+}
+
+// UpdatePuppy updates the store with key of id with the new puppy.
+// It returns a boolean whether the operation is successful or not.
+func (s *SyncStore) UpdatePuppy(id uint32, puppy *Puppy) (bool, error) {
+	if _, ok := s.syncStore.Load(id); ok {
+		s.mu.Lock()
```
pfannerj: My question here is whether or not the Lock should be done before the Load... in the example that @anzboi  provided with lab6, it does the Lock as the first statement and then defer Unlock (as mentioned in my comment above)... I see ur logic in that u are only locking if u need to update but I think using the Lock/defer Unlock pattern is preferable.

07_errors/willshen8/error.go
```diff
@@ -0,0 +1,13 @@
...
+
+type Error struct {
+	Message string
+	Code    int
+}
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("Error code %d: %s",
+		e.Code, e.Message)
+}
```
pfannerj, line 13: I like the idea in the example where the error codes are defined as an enum within error.go and then the more meaningful name (eg. ErrNotFound) is passed into the error function rather than just passing a 1 or 2.

pfannerj:
Everything LGTM but I'll hold off labelling with "Approved by colleague" until I confirm that it's ok to Lock after the Load... or maybe you wanna just make that change and use defer Unlock?? 
Also noticed that your not looking for the bonus points for the build the new leveldb storer...

pfannerj:
I just had a chat with Dan and he says the update should definitely do the Lock and defer Unlock up front at the start of the function.. so I'll mark with 'Changes Requested' and just let me know when ur ready for me to have another look...

07_errors/willshen8/syncStore.go
willshen8: Thanks, changed added.

07_errors/willshen8/syncStore.go
willshen8: II have tried locking it before the  read operation and it goes to an infinite loop.

07_errors/willshen8/error.go
willshen8: I was thinking about this one and thought what they asking is an error code similar to network response code like 404, 200, etc. 

The one in the lecture slides `var ErrPermission = errors.New("permission denied")` actually returns a string. And my idea of code is a number, but I could be wrong.

So my change is to add the equivalent of enum:
```
const (
	NEGATIVEVALUE = 1001
	NONEXISTENTPUPPY = 1002
)
```

willshen8:
@pfannerj Thanks judy, changes added after your PR review. Please approve if you are happy with it,

07_errors/willshen8/error.go
pfannerj: yeah, that's good, numeric code is in the example too.

07_errors/willshen8/syncStore.go
pfannerj: hmmm, works fine for me... 
```
// UpdatePuppy modifies puppy data in the sync store, either creating a new one or overwriting an old one.
func (s *SyncStore) UpdatePuppy(puppyID uint32, p *Puppy) (uint32, error) {
	s.Lock()
	defer s.Unlock()
	if _, ok := s.Load(puppyID); !ok {
		s.currID++
		puppyID = s.currID
	}
	p.ID = puppyID //ensure the ID within p always matches the sync store key (puppyID)
	s.Store(puppyID, *p)
	return puppyID, nil
}
```
Do u want me to come and visit u?

pfannerj: Sorry... I don't know how to insert code so that it displays properly

camscale: @pfannerj Use triple backticks around the block of code instead of single backticks. I've edited your comment, so take a look

willshen8: @camh-anz  Thanks Cam, I told her the answer via slack yesterday.

07_errors/willshen8/error.go
```diff
@@ -0,0 +1,18 @@
...
+
+const (
+	NEGATIVEVALUE    = 1001
```
anzboi: Go convention is to name these the same as any other variable
`NegativeValue`
`NonExistentPuppy`

07_errors/willshen8/storer_test.go
```diff
@@ -0,0 +1,132 @@
...
+	readPuppy1, _ := s.storer.ReadPuppy(createdID)
+	m.Equal(&s.Puppy1, readPuppy1)
+}
+
+func (s *StorerSuite) TestCreatePuppyWithValueThanZero() {
+	m := assert.New(s.T())
+	var buf bytes.Buffer
+	out = &buf
+	createdID, createError := s.storer.CreatePuppy(&s.InvalidPuppy)
+	fmt.Fprintln(out, createError)
```
anzboi: You can use `assert.EqualValues(t, expected, actual)` to check the errors directly, rather than printing, capturing output and checking the error strings

Note that your interfaces return a generic `error` rather than your `*Error`. I actually prefer this way of doing it, but you may need to type cast the returned error first (though I think `EqualValues` handles this via reflection anyway). If you do need to, here's the code

```go
createID, createError := s.storer.CreatePuppy(&s.InvalidPuppy)

// You may not need these lines if EqualValues works on its own
// require.IsType(t, *Error{}, createError) // Check the type is correct, fail and stop test if not
// err := createError.(*Error)

assert.EqualValues(t, expected, createError)
```

07_errors/willshen8/storer_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type StorerSuite struct {
+	suite.Suite
+	storer        Storer
+	Puppy1        Puppy
```
anzboi: Store these as unexported variables
```
var (
  puppy1 = Puppy{1, "Jack Russell Terrier", "White and Brown", "1500"}
  ...

  createError = Error{Code: NEGATIVEVALUE, Message: "Puppy value can't be less than 0"}
  ...
)
```

This also avoids recreating them for every test (just don't modify them in any tests, copy first if you need to modify)

07_errors/willshen8/storer_test.go
```diff
@@ -0,0 +1,132 @@
...
+	InvalidPuppy  Puppy
+	ErrorCreateID uint32
+	ErrorCreate   string
+	ErrorRead     string
+	ErrorUpdate   string
+	ErrorDelete   string
+}
+
+func (s *StorerSuite) SetupTest() {
+	s.Puppy1 = Puppy{1, "Jack Russell Terrier", "White and Brown", "1500"}
```
anzboi: Move these to the variables at the top of the file.

07_errors/willshen8/puppy.go
```diff
@@ -0,0 +1,35 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+type Puppy struct {
+	ID    uint32
+	Breed string
+	Color string
+	Value string
```
anzboi, line 15: Not sure if value should be numeric or a string. Most days I would argue numeric, but there are non-trivial challenges storing monetary data, and for the json api you will need to do the integer conversion anyway so I will let this slide.

On a side note, A couple of people at anz have been working on an IEEE decimal implementation in go (which solves the problems with storing dollar values). Once that is in, there would be no reason to keep using strings for dollar values.

anzboi (changes requested):
Looking good, love the test coverage.
I would like to rework the suite tests though, move the test data out of the suite to unexported variables, and rework how you check your errors (check them directly rather than through output of `Error()`)

07_errors/willshen8/error.go
willshen8: Thanks agreed, was thinking Java when I wrote these constants.

07_errors/willshen8/storer_test.go
willshen8: Thanks, this is addressed. I simply used the testingSuite `Equal` function.

07_errors/willshen8/storer_test.go
willshen8: Addressed also.

07_errors/willshen8/storer_test.go
willshen8: Done!

07_errors/willshen8/puppy.go
willshen8: I think the original intention was that we could have currency signs like $.

07_errors/willshen8/storer_test.go
willshen8: Hi Dan, please note that I still used this old-style for one of the test cases for 100% test coverage purpose.

anzboi (approved):
LGTM, great work

PR #533, opened 2019-07-03

### LINKED ISSUE #532: Lab 6 - CRUD puppy with interface
Puppy's have to be stored somewhere!

Implement a Puppy struct containing ID, Breed, Colour, Value.
Create Storer interface with CRUD methods for Puppy
Write a MapStore implementation of Storer backed by a map
Write a SyncStore implementation of Storer backed by a sync.Map
Keep all implementation files in the same folder and in package main
Test against the Storer interface and run in suite with both implementations

undrewb:
Fixes #532 

Review of colleague's PR #424 

Implement a Puppy struct containing ID, Breed, Colour, Value.
Create Storer interface with CRUD methods for Puppy
Write a MapStore implementation of Storer backed by a map
Write a SyncStore implementation of Storer backed by a sync.Map
Keep all implementation files in the same folder and in package main
Test against the Storer interface and run in suite with both implementations


06_puppy/undrewb/types.go
```diff
@@ -0,0 +1,28 @@
+package main
+
+import(
```
golangcibot: File is not `goimports`-ed (from `goimports`)

.gitignore
```diff
@@ -11,3 +11,4 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 coverage.txt
```
willshen8: Oops, this file should have not been committed.

06_puppy/undrewb/types.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (p *Puppy) Clone() *Puppy {
+	puppy := new(Puppy)
+	puppy.ID = p.ID
+	puppy.Breed = p.Breed
+	puppy.Colour = p.Colour
+	puppy.Value = p.Value
+	return puppy
+}
+
```
willshen8, line 18: I like the idea of immutability.

06_puppy/undrewb/MapStore.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+	delete(*msp, p.ID)
+	return true, nil
+}
+
+func (msp *MapStore) UpdatePuppy(id uint32, puppy *Puppy) error {
+	if puppy == nil {
+		return fmt.Errorf("cant update to a nil puppy")
+	}
+	p, exist := (*msp)[id]
```
willshen8: This can be rewritten as follow (seem to be the way go is written mostly)
```
if p, exist := (*msp)[id]; !exist {
    return errors.New("no such puppy")
}
```

06_puppy/undrewb/store_test.go
```diff
@@ -0,0 +1,177 @@
...
+		wantErr bool
+	}{
+		{name: "New corgi",
+			puppy:   corgi,
+			wantErr: false},
+		{name: "Existing corgi",
+			puppy:   corgi,
+			wantErr: true},
+	}
+	for _, tt := range tests {
```
willshen8, line 60: This is just a comment. I have written my test cases as unit test cases, as each one test a specific condition. Rather than bundle all the create operations into one. Not saying it's right or wrong. But I think test cases should be atomic. And it will eliminate the need for struct, and making test files shorter.

06_puppy/undrewb/store_test.go
```diff
@@ -0,0 +1,177 @@
...
+		tt := tt
+		s.T().Run(tt.name, func(t *testing.T) {
+			got, err := s.store.ReadPuppy(tt.id)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("ReadPuppy() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
willshen8: You can try to use the 'assert.Equal' library, and reduce your code length:

e.g. ` assert.Equal(a, b, "The two words should be the same.")`

06_puppy/undrewb/store_test.go
```diff
@@ -0,0 +1,177 @@
...
+	suite.Run(t, &storerSuite{
+		storerType: func() Storer { return &SyncStore{} },
+	})
+}
+
+func (s *storerSuite) SetupTest() {
+	s.store = s.storerType()
+
+	corgi := &Puppy{1000, "corgi", "orange and white", "$$$$$$$"}
+	_ = s.store.CreatePuppy(corgi)
```
willshen8, line 29: I don't believe you should have `createPuppy` method inside `SetUpTest` func, as if your `createPuppy` method is wrong, then it would have failed here. Consider just have raw data in here.

06_puppy/undrewb/store_test.go
```diff
@@ -0,0 +1,177 @@
...
+		want    *Puppy
+	}{
+		{name: "lookup existing puppy",
+			id:      1000,
+			want:    corgi,
+			wantErr: false,
+		},
+		{name: "lookup non-existing puppy",
+			id:      1001,
+			want:    nil,
```
willshen8, line 90: I think this is bit verbose, you could have just run `suite.Nil(expected, actual)` in your test cases.

06_puppy/undrewb/store_test.go
```diff
@@ -0,0 +1,177 @@
...
+			want:    false,
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
+		s.T().Run(tt.name, func(t *testing.T) {
+			got, err := s.store.DeletePuppy(tt.id)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("DeletePuppy() error = %v, wantErr %v", err, tt.wantErr)
```
willshen8, line 134: Same as above. just use `assert.Equal` to test for differences.

willshen8 (changes requested):
It's mostly pretty good.

1. Uncommit gitignore file.
2. Fix up your test cases.

06_puppy/undrewb/MapStore.go
undrewb: That's different logic. I'm making sure the puppy passed in isn't nil.

06_puppy/undrewb/store_test.go
undrewb: Equal and DeepEqual are quite different.

06_puppy/undrewb/store_test.go
undrewb: Same as above.

06_puppy/undrewb/store_test.go
undrewb: If I tie the insert to implementation I cant use the same code for testing different implementations of the same interface. If the test fails because the implementation is wrong thats a good thing :)

06_puppy/undrewb/store_test.go
undrewb: I only want nil for a specific case, not for all tests. When I want corgi (line 79) suite.Nil isn't appropriate.

juliaogris:

PR #534, opened 2019-07-04
Lab 2: Bubble Sort

krnicolson:
Fixes #540 

Review of colleague's PR #554

#### Changes proposed in this PR:

- implement bubble() function
- implement main function to call bubble with required input
- test main and bubble

PR #536, opened 2019-07-04
Lab7- Errors

### LINKED ISSUE #535: Lab 7 - Errors
- Create errors for:

        - Value < 0

mohitnag:
Fixes #535

Review of colleague's PR #534 

- Add Errors in Puppy Store.
- Add mutex in sync store


patrickmarabeas:
Don't forget to apply the correct labels on your PR. Also look at the PR markdown formatting - it's a little all over the place.

You could be more descriptive in the error handling that's been applied.

You have also included your lab 06 commit within this PR. This should be removed.

It would also be handy to split the initial c+p from Lab 06 into a separate commit so Lab 07's additions can be seen explicitly.

06_puppy/mohitnag/types.go
```diff
@@ -0,0 +1,27 @@
...
+// Puppy defines the data structure corresponding to a Puppy
+type Puppy struct {
+	ID     uint32
+	Breed  string
+	Colour string
+	Value  string
+}
+
+// MemStore stores Puppy details with Puppy Id as Key and Puppy  as value
+type MemStore map[uint32]Puppy
```
patrickmarabeas: This should be called `MapStore`

patrickmarabeas: This will need to be updated in Lab 07 as well.

mohitnag:
I didnt add the label because i put the PR with 'WIP' tag. I will put the labels now. after doing the suggested changes

06_puppy/mohitnag/types.go
mohitnag: although naming is subjective. Still i will rename it as MapStore to make it more consistent with SyncStore

07_errors/mohitnag/error.go
```diff
@@ -0,0 +1,29 @@
+package main
+
+import "fmt"
+
+const (
+	// InvalidValue is used when the puppy value is below 0
+	InvalidValue = "invalid input"
```
juliaogris: Invalid value sounds odd.
How about just invalid or invalidinput?

07_errors/mohitnag/mapstore.go
```diff
@@ -0,0 +1,44 @@
...
+func (m MapStore) CreatePuppy(p Puppy) error {
+	if _, ok := m[p.ID]; !ok {
+		val, _ := strconv.Atoi(p.Value)
+		if val < 0 {
+			return ErrorF(InvalidValue, "puppy with value less than 0 not allowed")
+		}
+		m[p.ID] = p
+		return nil
+	}
+	return ErrorF(Duplicate, "puppy with Id %d already exists", p.ID)
```
juliaogris: Imo invert logic: if ok... Then error.
It's more consistent with the ifs usually containing the error case and "bail early on error"

07_errors/mohitnag/types.go
```diff
@@ -0,0 +1,28 @@
...
+// Puppy defines the data structure corresponding to a Puppy
+type Puppy struct {
+	ID     uint32
+	Breed  string
+	Colour string
+	Value  string
+}
+
+// MapStore stores Puppy details with Puppy Id as Key and Puppy  as value
+type MapStore map[uint32]Puppy
```
juliaogris: Move to mapstore.go 
Same below

juliaogris (changes requested):
Fix pr description markdown please.

07_errors/mohitnag/error.go
mohitnag: Done

07_errors/mohitnag/mapstore.go
mohitnag: Completely agree.

07_errors/mohitnag/types.go
mohitnag: Done

07_errors/mohitnag/error.go
```diff
@@ -0,0 +1,29 @@
...
+const (
+	// Invalid is used when the puppy value is below 0
+	Invalid = "invalid input"
+	// NotFound is used when the puppy with given id is not found
+	NotFound = "puppy not found"
+	//Duplicate is used when the puppy id already exists
+	Duplicate = "puppy already exists"
+)
+
+// Error is the authorisationservice package error with a code for comparison
```
rokane: I think this comment should be updated. Reference to the `authorisationservice package` is external to this project and should not be assumed knowledge. Just provide a description of what the struct is.

07_errors/mohitnag/error.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+// Error method implments the error interface for message retrieval
+func (e *Error) Error() string {
+	return fmt.Sprintf("%s :(code: %s)", e.Message, e.Code)
+}
+
+// ErrorF is a utility function creating an error with given code and message
+func ErrorF(code, format string, args ...interface{}) *Error {
+	message := fmt.Sprintf(format, args...)
```
rokane: No need to declare a separate `message` variable if we can return straight away. 

```go
return &Error{Code: code, Message: fmt.Sprintf(format, args...)}
```

07_errors/mohitnag/error_test.go
```diff
@@ -0,0 +1,15 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestErrors(t *testing.T) {
+	assert := assert.New(t)
+	err := ErrorF(NotFound, "test error message")
+	assert.Equal(NotFound, err.Code)
+	assert.Equal("test error message", err.Message)
+	assert.Equal("test error message :(code: puppy not found)", err.Error())
```
rokane, line 14: I find this error message to be a bit misleading, and I think it comes from the fact that a string type is being used to store the code of the error. In my opinion I think a message format like below would be more appropriate:

```go
assert.Equal("puppy not found (404)", err.Error())
```

07_errors/mohitnag/mapstore.go
rokane: Doesn't look like this has been applied yet. Can you please do so?

rokane (dismissed):
Looking really good @mohitnag. I have left a couple of comments throughout, the main one which I thing should be addressed is the format of the messages, I feel that the code should be an int code. But please let me know if you think otherwise.

07_errors/mohitnag/error.go
mohitnag: Done .

07_errors/mohitnag/error.go
mohitnag: Done

07_errors/mohitnag/error_test.go
mohitnag: I think error code in `string` type are more readable and easy to understand. However i  do understand that `int` code like 404 is conventional.  I would prefer keeping as it is, but let me know your thoughts and I will update accordingly.

07_errors/mohitnag/syncstore.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+// CreatePuppy creates a Puppy in syncstore
+func (s *SyncStore) CreatePuppy(p Puppy) error {
+	s.m.Lock()
+	defer s.m.Unlock()
+	if _, ok := s.Load(p.ID); ok {
+		return ErrorF(Duplicate, "puppy with Id %d already exists", p.ID)
+	}
+	val, _ := strconv.Atoi(p.Value)
```
juliaogris, line 21: nit: might be dangerous to ignore the error here.

juliaogris (approved):
one final nit, feel free to ignore.

PR #538, opened 2019-07-04
Implement CRUD puppy with interface

### LINKED ISSUE #537: Lab 6 - CRUD puppy with interface
# Lab 6 - CRUD puppy with interface

- Create an executable go program in directory `06_puppy/USERNAME` (see [hints](https://github.com/anz-bank/go-samplerest/blob/master/pkg/pet/types.go))
- Create `Storer` interface with [crud](https://en.wikipedia.org/wiki/Create,\_read,\_update_and_delete) methods for `Puppy`
- Write a `MapStore` implementation of `Storer` backed by a `map`
- Write a `SyncStore` implementation of `Storer` backed by a [sync.Map](https://golang.org/pkg/sync/#Map)
- Keep all implementation files in the same folder and in package `main`
- Test against the `Storer` interface and run in [suite](https://godoc.org/github.com/stretchr/testify/suite) with both implementations

patrickmarabeas:
Adds the Puppy, Storer, MapStore and SyncStore type definitions. 

- MapStore implements the Storer interface and supplies a NewMapStore factory function.
- SyncStore implements the Storer interface and supplies a NewSyncStore factory function.
- Uuid increment is handled internally by MapStore Create
- Read will return an empty Puppy struct if provided uuid doesn't exist
- Update will return the given Puppy struct if the uuid doesn't exist
- Destroy returns a bool dependent on uuid existing

Fixes #537

Review of colleague's PR #536

### DIFF
--- a/06_puppy/patrickmarabeas/main.go
+++ b/06_puppy/patrickmarabeas/main.go
@@ -1,0 +1,5 @@
+package main
+
+func main() {
+
+}
--- a/06_puppy/patrickmarabeas/mapStore.go
+++ b/06_puppy/patrickmarabeas/mapStore.go
@@ -1,0 +1,56 @@
+package main
+
+type MapStore struct {
+	uuid  uint
+	store map[uint]Puppy
+}
+
+// NewMapStore returns a pointer to a new instance of the MapStore struct which implements the Storer interface.
+func NewMapStore() Storer {
+	return &MapStore{
+		uuid:  0,
+		store: map[uint]Puppy{},
+	}
+}
+
+// Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
+func (store *MapStore) Create(puppy Puppy) uint {
+	puppy.ID = store.uuid
+	store.store[puppy.ID] = puppy
+	store.uuid++
+
+	return puppy.ID
+}
+
+// Read returns the puppy matching the provided uuid.
+// An empty Puppy struct is returned if the identifier does not exist.
+func (store *MapStore) Read(id uint) Puppy {
+	if _, ok := store.store[id]; ok {
+		return store.store[id]
+	}
+
+	return Puppy{}
+}
+
+// Update modifies the puppy matching the provided uuid in the store with the provided Puppy struct.
+// It returns a bool whether a matching identifier was modified or not.
+func (store *MapStore) Update(id uint, puppy Puppy) bool {
+	if _, ok := store.store[id]; !ok {
+		return false
+	}
+
+	puppy.ID = id
+	store.store[id] = puppy
+	return true
+}
+
+// Destroy removes the puppy matching the provided uuid from the store.
+// It returns a bool whether a matching identifier was deleted or not.
+func (store *MapStore) Destroy(id uint) bool {
+	if _, ok := store.store[id]; !ok {
+		return false
+	}
+
+	delete(store.store, id)
+	return true
+}
--- a/06_puppy/patrickmarabeas/store_test.go
+++ b/06_puppy/patrickmarabeas/store_test.go
@@ -77,4 +77,5 @@ func (suite *StoreSuite) TestIdIncrementOnDelete() {
 
 func TestStore(t *testing.T) {
 	suite.Run(t, &StoreSuite{store: NewMapStore()})
+	suite.Run(t, &StoreSuite{store: NewSyncStore()})
 }
--- a/06_puppy/patrickmarabeas/store_test.go
+++ b/06_puppy/patrickmarabeas/store_test.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+type StoreSuite struct {
+	suite.Suite
+	store Storer
+}
+
+func (suite *StoreSuite) TestCreate() {
+	a := assert.New(suite.T())
+	id := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: "450"})
+	id2 := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: "300"})
+
+	a.EqualValues(id+1, id2)
+}
+
+func (suite *StoreSuite) TestRead() {
+	a := assert.New(suite.T())
+	data := suite.store.Read(0)
+
+	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: "450"})
+}
+
+func (suite *StoreSuite) TestReadNonExistent() {
+	a := assert.New(suite.T())
+	success := suite.store.Read(100)
+
+	a.Equal(success, Puppy{})
+}
+
+func (suite *StoreSuite) TestUpdate() {
+	a := assert.New(suite.T())
+	success := suite.store.Update(0, Puppy{Breed: "Doberman", Color: "Black", Value: "500"})
+	data := suite.store.Read(0)
+
+	a.Equal(success, true)
+	a.Equal(data, Puppy{ID: 0, Breed: "Doberman", Color: "Black", Value: "500"})
+}
+
+func (suite *StoreSuite) TestUpdateNonExistent() {
+	a := assert.New(suite.T())
+	success := suite.store.Update(100, Puppy{Breed: "Doberman", Color: "Black", Value: "500"})
+
+	a.Equal(success, false)
+}
+
+func (suite *StoreSuite) TestDestroy() {
+	a := assert.New(suite.T())
+	success := suite.store.Destroy(1)
+	data := suite.store.Read(1)
+
+	a.Equal(success, true)
+	a.Equal(data, Puppy{})
+}
+
+func (suite *StoreSuite) TestDestroyNonExistent() {
+	a := assert.New(suite.T())
+	success := suite.store.Destroy(100)
+
+	a.Equal(success, false)
+}
+
+func (suite *StoreSuite) TestIdIncrementOnDelete() {
+	a := assert.New(suite.T())
+	id := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
+	suite.store.Destroy(id)
+	newID := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
+
+	a.EqualValues(newID, 3)
+}
+
+func TestStore(t *testing.T) {
+	suite.Run(t, &StoreSuite{store: NewMapStore()})
+}
--- a/06_puppy/patrickmarabeas/syncStore.go
+++ b/06_puppy/patrickmarabeas/syncStore.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"sync"
+)
+
+type SyncStore struct {
+	uuid uint
+	sync.Map
+}
+
+// NewSyncStore returns a pointer to a new instance of the SyncStore struct which implements the Storer interface.
+func NewSyncStore() Storer {
+	return &SyncStore{
+		uuid: 0,
+	}
+}
+
+// Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
+func (store *SyncStore) Create(puppy Puppy) uint {
+	puppy.ID = store.uuid
+	store.Store(puppy.ID, puppy)
+	store.uuid++
+
+	return puppy.ID
+}
+
+// Read returns the puppy matching the provided uuid.
+// An empty Puppy struct is returned if the identifier does not exist.
+func (store *SyncStore) Read(id uint) Puppy {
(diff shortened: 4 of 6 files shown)


pentaphobe (commented):
After a while staring at your code (and adding a few extra tests) I cannot work out what’s happening to your coverage.

_However_ one thing which _may_ help is if you conform to the go style of returning an `error` from functions which have an error condition and testing against that.

Still don't think that should matter, but never know...  (also worth mentioning just as a stylistic nitpick)

---

FWIW: I sometimes use the HTML view of coverage to help see what the tool's seeing, but in this case it told me very little.

Either way, this is the command I used:
```bash
go test -v -race -coverprofile /tmp/cover.out -covermode=atomic ./... && go tool cover -html /tmp/cover.out -o /tmp/profile.html && open /tmp/profile.html
```

---

FWIW 2: Tests I tried out were things like:

- doing non deep-equals checks of zero value returns
- calling the offending functions on a `nil` store (which obviously just crashes - alas precluding any coverage result.  Whilst it's arguable that one might need to check for such things, surely that would cause coverage issues on your other functions too)
- started adding a custom error to your code (for my own edification) but the task exploded - and was discarded - once I realised I'd have to update signatures and calls across most files and it's 2am :)

patrickmarabeas:
Hopefully it's not error related - as I removed all my error handling upon seeing Lab 07 outline!

06_puppy/patrickmarabeas/mapStore_test.go
```diff
@@ -0,0 +1,96 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestMapStore(t *testing.T) {
+
+	t.Run("NewMapStore", func(t *testing.T) {
```
anzdaddy: Seems like these should just be separate top-level `Test…` functions.

06_puppy/patrickmarabeas/mapStore_test.go
```diff
@@ -0,0 +1,96 @@
...
+			store: map[int]Puppy{},
+		}
+
+		if !reflect.DeepEqual(expected, got) {
+			t.Errorf("\nExpected: %#v\nGot:      %#v", expected, got)
+		}
+	})
+
+	t.Run("Create", func(t *testing.T) {
+		// Setup
```
anzdaddy: Avoid comments that state the obvious.

06_puppy/patrickmarabeas/syncStore.go
```diff
@@ -0,0 +1,49 @@
...
+	if value, ok := store.Load(id); ok {
+		return value.(Puppy)
+	}
+
+	return Puppy{}
+}
+
+// Update modifies the puppy matching the provided uuid in the store with the provided Puppy struct.
+// No action is performed if the identifier does not exist.
+func (store *SyncStore) Update(id int, puppy Puppy) Puppy {
```
anzdaddy: Returning the input `Puppy` is redundant, which results in extra testing requirements.

06_puppy/patrickmarabeas/mapStore.go
```diff
@@ -0,0 +1,50 @@
...
+	if _, ok := store.store[id]; ok {
+		return store.store[id]
+	}
+
+	return Puppy{}
+}
+
+// Update modifies the puppy matching the provided uuid in the store with the provided Puppy struct.
+// No action is performed if the identifier does not exist.
+func (store *MapStore) Update(id int, puppy Puppy) Puppy {
```
anzdaddy: Returning the input `Puppy` is redundant, which results in extra testing requirements.

06_puppy/patrickmarabeas/syncStore_test.go
```diff
@@ -0,0 +1,95 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestSyncStore(t *testing.T) {
+
+	t.Run("NewSyncStore", func(t *testing.T) {
```
anzdaddy: Seems like these should just be separate top-level `Test…` functions.

06_puppy/patrickmarabeas/syncStore_test.go
```diff
@@ -0,0 +1,95 @@
...
+			t.Errorf("\nExpected: %#v\nGot:      %#v", expected, got)
+		}
+	})
+
+	t.Run("Create", func(t *testing.T) {
+		// Setup
+		store := NewMapStore()
+		// Create a puppy
+		id := store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: "450"})
+		expectedID := 1
```
anzdaddy: Is it part of the API design that the first id is 1? If so, then what should the second puppy get after inserting and removing the first Puppy? If not, then you shouldn't test for it.

06_puppy/patrickmarabeas/syncStore_test.go
```diff
@@ -0,0 +1,95 @@
+package main
```
anzdaddy: Is this a duplicate of `mapStore_test.go` with a different backend? You should implement a single set of parameterised tests.

anzdaddy (changes requested):

06_puppy/patrickmarabeas/syncStore.go
pentaphobe: potential rationale for returning a value: the act of updating potentially has secondary side effects beyond the data in the passed struct (eg. modifying something like `UpdateTime`)

good point re. extra testing req's though

juliaogris:
Please fix broken build and address @anzdaddy 's comments

06_puppy/patrickmarabeas/mapStore_test.go
patrickmarabeas: Tests have been refactored.

06_puppy/patrickmarabeas/syncStore.go
patrickmarabeas: Simplified to bool

06_puppy/patrickmarabeas/mapStore.go
patrickmarabeas: Simplified to bool.

06_puppy/patrickmarabeas/syncStore_test.go
patrickmarabeas: Tests have been refactored.

06_puppy/patrickmarabeas/syncStore_test.go
patrickmarabeas: Tests have been refactored.

06_puppy/patrickmarabeas/store_test.go
```diff
@@ -0,0 +1,87 @@
...
+	success := suite.store.Destroy(100)
+
+	a.Equal(success, false)
+}
+
+func (suite *StoreSuite) TestIdIncrementOnDelete() {
+	a := assert.New(suite.T())
+	id := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
+	suite.store.Destroy(id)
+	newId := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
```
golangcibot: var `newId` should be `newID` (from `golint`)

06_puppy/patrickmarabeas/types.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type Storer interface {
+	Create(puppy Puppy) int
+	Read(ID int) Puppy
+	Update(ID int, puppy Puppy) bool
+	Destroy(ID int) bool
+}
+
+type MapStore struct {
```
anzboi: I would put `MapStore` and `SyncStore` in their respective implementation files `mapStore.go` and `syncStore.go`

06_puppy/patrickmarabeas/types.go
```diff
@@ -0,0 +1,29 @@
...
+	uuid  int
+	store map[int]Puppy
+}
+
+type SyncStore struct {
+	uuid int
```
anzboi: Interesting interpretation of `uuid` you have there.

06_puppy/patrickmarabeas/store_test.go
```diff
@@ -0,0 +1,87 @@
...
+	id := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: "450"})
+
+	a.Equal(id, 0)
+}
+
+func (suite *StoreSuite) TestCreateSecond() {
+	a := assert.New(suite.T())
+	id := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: "300"})
+
+	a.Equal(id, 1)
```
anzboi: I think what @anzdaddy was getting at here is that `TestCreate` should create 2 puppies, and check the second ones ID is incremented by 1 from the first, or at least different, think about what could be generating these id's if you don't know what the implementation is.
Otherwise, great tests.

anzboi (changes requested):
Great work, Just one think I would like to fix up.
From an interface level, we don't care about what ids are being created, only that they are different for any two `Create` calls. IMO the Create test should just create two puppies and check the generated ID's are different.
Feel free to disagree and I'll approve.

06_puppy/patrickmarabeas/types.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+type Storer interface {
+	Create(puppy Puppy) uint
+	Read(ID uint) Puppy
+	Update(ID uint, puppy Puppy) bool
+	Destroy(ID uint) bool
+}
+
+type SyncStore struct {
```
golangcibot: `SyncStore` redeclared in this block (from `typecheck`)

06_puppy/patrickmarabeas/syncStore.go
```diff
@@ -0,0 +1,59 @@
+package main
+
+import (
+	"sync"
+)
+
+type SyncStore struct {
```
golangcibot, line 7: other declaration of SyncStore (from `typecheck`)

06_puppy/patrickmarabeas/types.go
patrickmarabeas: Done

06_puppy/patrickmarabeas/types.go
patrickmarabeas: Changed to `uint`

06_puppy/patrickmarabeas/store_test.go
patrickmarabeas: Done

06_puppy/patrickmarabeas/main.go
```diff
@@ -0,0 +1,5 @@
+package main
+
+func main() {
+
```
rokane, line 4: I find it a little strange seeing a blank main program. I know the description of the lab doesn't state what you need to include in the main, so you have technically ticked all the boxes, but I feel it makes sense to maybe create a puppy or something to give context to the work being done. Happy to hear your thoughts otherwise.

06_puppy/patrickmarabeas/mapStore.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+type MapStore struct {
```
rokane, line 3: Would be nice to see some documentation on the MapStore struct as well, since it is a part of the exported package.

06_puppy/patrickmarabeas/syncStore.go
rokane, line 7: Again here with the documentation as well

06_puppy/patrickmarabeas/types.go
```diff
@@ -0,0 +1,15 @@
+package main
+
+type Puppy struct {
+	ID    uint
+	Breed string
+	Color string
+	Value string
+}
+
+type Storer interface {
```
rokane, line 10: Again here with the documentation.

06_puppy/patrickmarabeas/store_test.go
```diff
@@ -0,0 +1,81 @@
...
+func (suite *StoreSuite) TestRead() {
+	a := assert.New(suite.T())
+	data := suite.store.Read(0)
+
+	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: "450"})
+}
+
+func (suite *StoreSuite) TestReadNonExistent() {
+	a := assert.New(suite.T())
+	success := suite.store.Read(100)
```
rokane, line 32: success seems like a strange name for this variable. Maybe `notFound` or something which indicates that it didn't exist?

rokane (changes requested):
Looking really good! Great suite of tests and really clean, well documented code. I have left a couple of comments throughout which I think would be nice to have. Also there is the empty main function, if you feel against what I have suggested then let me know, otherwise it would be nice to see something in there (even if it's simple)

juliaogris:

PR #518, opened 2019-06-30
Implement lab 04 numeronym

undrewb:
Implement functions
Implement tests
Fixes #517 

Review of colleague's PR #515

#### Changes proposed in this PR:

- Implement lab 4 of go-course
- Decompose into numeronym and numeronyms for testing.
- Don't consider non-ascii or special chars ( like _ or white space ) different to regular chars.
Implement testing.

PR #542, opened 2019-07-07
Implemented Lab4 numeronym and test cases

sirigithub:
Fixes #

Review of colleague's PR #518

-  Implemented executable go program in directory `04_numeronym/sirigithub`
-  Implemented a function numeronyms that returns a slice of numeronyms for its input strings
- Included main method as per spec that calls the above function.
- Implemented tests for both methods

### DIFF
--- a/04_numeronym/sirigithub/main.go
+++ b/04_numeronym/sirigithub/main.go
@@ -1,0 +1,26 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	result := make([]string, len(vals))
+	for i, val := range vals {
+		r := []rune(val)
+		runeLength := len(r)
+		if runeLength < 4 {
+			result[i] = val
+		} else {
+			result[i] = fmt.Sprintf("%c%d%c", r[0], runeLength-2, r[runeLength-1])
+		}
+	}
+	return result
+}
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
--- a/04_numeronym/sirigithub/main.go
+++ b/04_numeronym/sirigithub/main.go
@@ -13,11 +13,10 @@ func numeronyms(vals ...string) []string {
 	for i, val := range vals {
 		r := []rune(val)
 		runeLength := len(r)
-		if runeLength < 4 {
-			result[i] = val
-		} else {
-			result[i] = fmt.Sprintf("%c%d%c", r[0], runeLength-2, r[runeLength-1])
+		if runeLength > 3 {
+			val = fmt.Sprintf("%c%d%c", r[0], runeLength-2, r[runeLength-1])
 		}
+		result[i] = val
 	}
 	return result
 }
--- a/04_numeronym/sirigithub/main_test.go
+++ b/04_numeronym/sirigithub/main_test.go
@@ -1,0 +1,45 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	actual := buf.String()
+	expected := "[a11y K8s abc]\n"
+	assert.Equal(t, expected, actual)
+
+}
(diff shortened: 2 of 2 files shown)


04_numeronym/sirigithub/main_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{description: "Empty string", input: []string{""}, expected: []string{""}},
+		{description: "Short strings", input: []string{"sss", "ab", "a"}, expected: []string{"sss", "ab", "a"}},
+		{description: "Alphanumeric string", input: []string{"sss1245sdfg"}, expected: []string{"s9g"}},
+		{description: "Back slashes and spaces", input: []string{"sss\\", "A long string with spaces"},
+			expected: []string{"s2\\", "A23s"}},
+		{description: "Emoji strings", input: []string{"😄🐷🙈🐷🏃😄", "😄🏃😄"},
+			expected: []string{"😄4😄", "😄🏃😄"}},
+		{description: "Unicode Strings", input: []string{"日本語日本語", "äöß€’üüüöäßß"},
+			expected: []string{"日4語", "ä10ß"}},
+		{description: "Strings with special charecters",
```
golangcibot: `charecters` is a misspelling of `characters` (from `misspell`)

sirigithub: Fixed.

n0npax (dismissed):
Looks good 👍 The only thing I could suggest is to split iteration over arguments and computing numeronym into separate functions, but it's no required in my mind.

04_numeronym/sirigithub/main.go
```diff
@@ -0,0 +1,26 @@
...
+func numeronyms(vals ...string) []string {
+	result := make([]string, len(vals))
+	for i, val := range vals {
+		r := []rune(val)
+		runeLength := len(r)
+		if runeLength < 4 {
+			result[i] = val
+		} else {
+			result[i] = fmt.Sprintf("%c%d%c", r[0], runeLength-2, r[runeLength-1])
+		}
```
camscale, line 18: just a suggestion, no need to make this change. I think this `if` block is more readable as:
```
if runeLength > 3 {
    val = fmt.Sprintf("%c%d%c", r[0], runeLength-2, r[runeLength-1])
}
result[i] = val
```

This captures that the only special thing being done is for strings with more than 3 runes and that it changes the value. As the assignment to the result slice is common, it is moved out of the conditional and done once.

04_numeronym/sirigithub/main_test.go
```diff
@@ -0,0 +1,45 @@
...
+		{description: "Emoji strings", input: []string{"😄🐷🙈🐷🏃😄", "😄🏃😄"},
+			expected: []string{"😄4😄", "😄🏃😄"}},
+		{description: "Unicode Strings", input: []string{"日本語日本語", "äöß€’üüüöäßß"},
+			expected: []string{"日4語", "ä10ß"}},
+		{description: "Strings with special characters",
+			input:    []string{"a##67&a$", "**(]]"},
+			expected: []string{"a6$", "*3]"}},
+	}
+	for _, test := range tests {
+		actual := numeronyms(test.input...)
```
camscale: This is part of the test and should be run inside the test closure. The same for the expected value really.

Outside the test (`t.Run`), I'd just do a `test := test` to capture the loop variable and move the actual & expected lines inside the test closure.

camscale (changes requested):
LGTM. A good set of test cases too. Just one thing to fix for executing the test case.

04_numeronym/sirigithub/main.go
sirigithub: @camh-anz that is a neat trick thank you, i made the change.

sirigithub: Done.

04_numeronym/sirigithub/main_test.go
sirigithub: Agree. Done.

camscale (approved):
LGTM. Great looking Go code. Thanks.

When you merge, can you either do a squash merge, or you could do a local rebase and squash and then force push. This ensures that master does not contain any fixup commits - just to good stuff. It keeps the history clean. If you do a rebase and force-push, ping me on slack when you've done it and i'll re-approve right away (since force pushing drops any current approvals).

PR #339, opened 2019-06-11
Adds Fibbonacci and Tests

mohitnag:
Fixes #340 

Review of colleague's PR #

#### Changes proposed in this PR:
- Adds the Fibonacci Function and its tests

PR #545, opened 2019-07-07
Lab3 - Add Letter frequency and sort letters functions

### LINKED ISSUE #544: Lab3 - Letter Frequency
Write a function that returns a mapping of each letter to its frequency
Write a function that returns a sorted slice of strings with elements "{key}:{val}"
- Concurrent implementation Letter frequency
- Provide comprehensive tests and benchmark

davidbroughsmyth:
Fixes #544 

Review of colleague's PR #339 

- Concurrent implementation Letter frequency `letters(s string) map[rune]int`
- Standard Letter frequency implementation `letterFreq(s string) map[rune]int`
- Function that returns a sorted slice of strings with elements "{key}:{val}"
- Provide comprehensive tests and benchmark for standard and Concurrent Letter frequency functions

### DIFF
--- a/03_letters/davidbroughsmyth/main.go
+++ b/03_letters/davidbroughsmyth/main.go
@@ -19,27 +19,27 @@ func letterFreq(s string) map[rune]int {
 	return f
 }
 
-// Concurrent implementation of letters for speed
-// Returns the letter frequency for string
+// Returns the letter frequency for a string
 func letters(s string) map[rune]int {
-	stringSlice := strings.Fields(s)
-	combineFreq := map[rune]int{}
-	size := len(stringSlice)
+	str := strings.Fields(s)
+	size := len(str)
 	results := make(chan map[rune]int, size)
 
-	for _, ss := range stringSlice {
+	for _, ss := range str {
 		go func(sx string) {
 			results <- letterFreq(sx)
 		}(ss)
 	}
 
-	//Combine the letter frequency maps passed back via the results channel
+	sum := map[rune]int{}
+	// Combine the letter frequency maps passed back via the results channel
 	for i := 0; i < size; i++ {
-		for r, freq := range <-results {
-			combineFreq[r] += freq
+		result := <-results
+		for r, freq := range result {
+			sum[r] += freq
 		}
 	}
-	return combineFreq
+	return sum
 }
 
 func sortLetters(m map[rune]int) []string {
--- a/03_letters/davidbroughsmyth/main.go
+++ b/03_letters/davidbroughsmyth/main.go
@@ -1,0 +1,56 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func letterFreq(s string) map[rune]int {
+	f := map[rune]int{}
+	for _, r := range s {
+		f[r]++
+	}
+	return f
+}
+
+// Concurrent implementation of letters for speed
+// Returns the letter frequency for string
+func letters(s string) map[rune]int {
+	stringSlice := strings.Fields(s)
+	combineFreq := map[rune]int{}
+	size := len(stringSlice)
+	results := make(chan map[rune]int, size)
+
+	for _, ss := range stringSlice {
+		go func(sx string) {
+			results <- letterFreq(sx)
+		}(ss)
+	}
+
+	//Combine the letter frequency maps passed back via the results channel
+	for i := 0; i < size; i++ {
+		for r, freq := range <-results {
+			combineFreq[r] += freq
+		}
+	}
+	return combineFreq
+}
+
+func sortLetters(m map[rune]int) []string {
+	keys := make([]string, 0, len(m))
+	for key, value := range m {
+		keys = append(keys, string(key)+":"+strconv.Itoa(value))
+	}
+	sort.Strings(keys)
+	return keys
+}
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
--- a/03_letters/davidbroughsmyth/main_test.go
+++ b/03_letters/davidbroughsmyth/main_test.go
@@ -85,30 +85,54 @@ var tests = map[string]struct {
 
 func TestLetterFreq(t *testing.T) {
 	for name, test := range tests {
-		testList := test
+		test := test
 		t.Run(name, func(t *testing.T) {
-			got := letterFreq(testList.input)
-			assert.Equal(t, testList.wantLetters, got)
+			got := letterFreq(test.input)
+			assert.Equal(t, test.wantLetters, got)
 		})
 	}
 }
 
 func TestLetter(t *testing.T) {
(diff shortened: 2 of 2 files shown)


03_letters/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,56 @@
...
+
+func letterFreq(s string) map[rune]int {
+	f := map[rune]int{}
+	for _, r := range s {
+		f[r]++
+	}
+	return f
+}
+
+// Concurrent implementation of letters for speed
```
camscale: Go doc strings have a particular format that should be followed: https://blog.golang.org/godoc-documenting-go-code.

The comment "Concurrent implementation of letters for speed" does not belong here either as that would end up in the documentation which is not relevant to callers of the function.

03_letters/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,56 @@
...
+	size := len(stringSlice)
+	results := make(chan map[rune]int, size)
+
+	for _, ss := range stringSlice {
+		go func(sx string) {
+			results <- letterFreq(sx)
+		}(ss)
+	}
+
+	//Combine the letter frequency maps passed back via the results channel
```
camscale: nitpick: You're using mixed styles for comments - some with `// ` (slash-slash-space) and some with `//` (slash-slash). I suggest you stick to the former as it is the most common Go style.

03_letters/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,56 @@
...
+	for i := 0; i < size; i++ {
+		for r, freq := range <-results {
```
camscale: This is a little confusing because it kind of conflates ranging over a map and ranging over a channel. I think it would be clearer to add a line:
```
result := <-results
for r, freq := range result {
```
Go has deliberately omitted certain features (ternary operator, post inc/dec in expressions, etc) as they often lead to confusing code, so good Go code generally aims to keep it simple. Splitting the channel read and the map iteration keeps it simple. And since you have benchmarks (cool), you can see it makes no difference to this generated code.

03_letters/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,126 @@
...
+		input:       "大云寺赞大云寺赞",
+		wantLetters: map[rune]int{'大': 2, '云': 2, '寺': 2, '赞': 2},
+		wantSort:    []string{"云:2", "大:2", "寺:2", "赞:2"},
+	},
+	"symbols_emoji": {
+		input:       "😸♞🁻∑👾👾😸😛😛😛",
+		wantLetters: map[rune]int{'♞': 1, '🁻': 1, '∑': 1, '👾': 2, '😸': 2, '😛': 3},
+		wantSort:    []string{"∑:1", "♞:1", "🁻:1", "👾:2", "😛:3", "😸:2"},
+	},
+}
```
camscale, line 84: You have no tests that have spaces in the input, yet the code uses `String.Fields` to split on spaces. So none of the tests are using much in the way of concurrency and that is the more complicated part of your code. You really want to ensure that is being tested more rigorously.

03_letters/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,126 @@
...
+
+func TestLetterFreq(t *testing.T) {
+	for name, test := range tests {
+		testList := test
```
camscale: `testList` is a funny name for the test case. If you were struggling to find an appropriate name, you can just re-use `test` - this is pretty common in this case:
```
test := test
```
It initially looks a bit weird, but that's useful because it highlights a weird aspect of Go closures over loop variables.

03_letters/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,56 @@
...
+		f[r]++
+	}
+	return f
+}
+
+// Concurrent implementation of letters for speed
+// Returns the letter frequency for string
+func letters(s string) map[rune]int {
+	stringSlice := strings.Fields(s)
+	combineFreq := map[rune]int{}
```
camscale: I'd suggest moving this closer to where it is used (just above the loop that reads the results). This looks too much like C-style - declaring vars at the top of the function, and is not idiomatic go - keeping the "range" of a variable short so you can stick to shorter names is the usual practice.

camscale (dismissed):
A couple of Go-oriented comments, plus some other general software engineering oriented ones. I let the latter ones slide, but not the former. Since you have some of the former, you may as well address them all.

However since you have gone with a more complicated implementation "for speed", were this a real project I'd be asking you to include the results of your benchmarking in the git commit message. More complicated code adds a maintenance burden and it needs to be justified. Optimising without showing the need for it should be avoided.

Even better would be two commits - the first with the basic implementation and the second with the faster one, with the commit message having the benchmark details. That way you have a baseline for the comparison. The tests in the first commit would be comprehensive enough that they apply unchanged to the concurrent implementation. But I won't hold you to that - it's just an idea and how I'd probably structure my commits in this case.

camscale:
Could you also update your commit message to adhere to the "Recipe for great commit message" as outlined in the course slides: https://go-course.org/gocourse-2019-05-melbourne.slide#31

I let it slide on lab 2, but since you need to address comments on this commit, it's time to get that right. Thanks.

03_letters/davidbroughsmyth/main.go
davidbroughsmyth: I've removed this comment

03_letters/davidbroughsmyth/main.go
davidbroughsmyth: Fixed this, it was unintentional and slipped through.

03_letters/davidbroughsmyth/main.go
davidbroughsmyth: I like this better too. The change has been made.

03_letters/davidbroughsmyth/main_test.go
davidbroughsmyth: I've added specific tests with spaces.

03_letters/davidbroughsmyth/main_test.go
davidbroughsmyth: I've made this change. It does look weird, but I also like the way closures work in Go.

03_letters/davidbroughsmyth/main.go
davidbroughsmyth: I've done this change.

davidbroughsmyth:
I've updated the commit message and add benchmark data showing the speed improvement.

davidbroughsmyth:
The speed I expected when implementing this Concurrent version of Letter frequency was not achieved, the benchmarks show that it is about 4 times slower than the standard synchronous version. Which is fine by me. I learned so much more about the go language and goroutines than any of the labs so far. I would like to keep the implementation as-is, assuming it meets the code owners criteria.

juliaogris (approved):
Looks great!

Mini-nit: I've updated your PR description Markdown (empty line after bulletin adds undesired paragraph)

PR #547, opened 2019-07-08
Lab4 - Add numeronyms function

### LINKED ISSUE #546: Lab4 - Implement numeronyms function
Provide test cases for numeronyms

davidbroughsmyth:
Fixes #546 

Review of colleague's PR #386 

 - Create `function numeronyms(vals ...string) []string` 
- When passed in "accessibility", "Kubernetes", "abc" returns a slice [a11y K8s abc]
- numeronyms supports unicode characters

### DIFF
--- a/04_numeronym/davidbroughsmyth/main.go
+++ b/04_numeronym/davidbroughsmyth/main.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"unicode/utf8"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	result := make([]string, len(vals))
+	for i, val := range vals {
+		c := utf8.RuneCountInString(val)
+		if c > 3 {
+			f, _ := utf8.DecodeRuneInString(val)
+			l, _ := utf8.DecodeLastRuneInString(val)
+			val = fmt.Sprintf("%c%d%c", f, c-2, l)
+		}
+		result[i] = val
+	}
+	return result
+}
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
--- a/04_numeronym/davidbroughsmyth/main_test.go
+++ b/04_numeronym/davidbroughsmyth/main_test.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := strconv.Quote("[a11y K8s abc]\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equal(t, expected, actual, "Unexpected output from main()")
+}
+
+var tests = map[string]struct {
+	input, want []string
+}{
+	"empty":        {input: []string{}, want: []string{}},
+	"SpecialChars": {input: []string{"@@!##%&&"}, want: []string{"@6&"}},
+	"3unicode":     {input: []string{"😛😸😛"}, want: []string{"😛😸😛"}},
+	"4unicode":     {input: []string{"😛😛😛大"}, want: []string{"😛2大"}},
+	"MixCodes":     {input: []string{"😛@#Xy 大Z2赞"}, want: []string{"😛8赞"}},
+	"Alphanumeric": {
+		input: []string{"8abcABC5Z", "cAt", "b1Rd"},
+		want:  []string{"87Z", "cAt", "b2d"},
+	},
+	"MaryPoppins": {
+		input: []string{"It's", "supercalifragilisticexpialidocious"},
+		want:  []string{"I2s", "s32s"},
+	},
+}
+
+func TestNumeronyms(t *testing.T) {
+	for name, test := range tests {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			got := numeronyms(test.input...)
+			assert.Equal(t, test.want, got)
+		})
+	}
+}


04_numeronym/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,39 @@
...
+	main()
+	expected := strconv.Quote("[a11y K8s abc]\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equal(t, expected, actual, "Unexpected output from main()")
+}
+
+var tests = map[string]struct {
+	input, want []string
+}{
+	"empty":    {input: []string{}, want: []string{}},
```
sirigithub: @davidbroughsmyth LGTM. My suggestion is to add a few more tests, special characters, alphanumeric strings.

sirigithub (commented):
Good implementation. Suggestion to add more tests.

04_numeronym/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	ook := make([]string, len(vals))
```
sirigithub: nitpick , it is not very obvious to me what the output variable ook means, is it just a random name in which case would you consider something more obvious like out/result/x  etc

davidbroughsmyth: Thanks for the review. I wondered if anyone would pick that up, it's my goto var name for simple functions like this. I should have changed it before committing.

camscale: Please revise your commits according to your reviewers comments. Then re-add the "Ready for review" label.

camscale:
I haven't reviewed this yet because of the outstanding comment to be addressed, but I have noted that your commit message does not follow the "Recipe for great commit message": https://go-course.org/gocourse-2019-05-melbourne.slide#31. Please ensure you follow that with your commit messages as a lot of git tooling assumes you will.

04_numeronym/davidbroughsmyth/main_test.go
davidbroughsmyth: @sirigithub I've added a few more tests as suggested.

04_numeronym/davidbroughsmyth/main.go
davidbroughsmyth: Fixed my strange variable name and added a few more tests.

04_numeronym/davidbroughsmyth/main_test.go
davidbroughsmyth: @sirigithub I've added more test to handle the suggest test conditions

davidbroughsmyth:
@camh-anz  I've addressed the reviews @sirigithub made. And changed the commit messages.

davidbroughsmyth:
Squashed my commits

04_numeronym/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+}{
+	"empty":        {input: []string{}, want: []string{}},
+	"SpecialChars": {input: []string{"@@!##%&&"}, want: []string{"@6&"}},
+	"Alphanumeric": {input: []string{"8abcABC5Z"}, want: []string{"87Z"}},
+	"StarTrek":     {input: []string{"Beam", "me", "up", "Scotty"}, want: []string{"B2m", "me", "up", "S4y"}},
+	"MaryPoppins": {
+		input: []string{"It's", "supercalifragilisticexpialidocious"},
+		want:  []string{"I2s", "s32s"},
+	},
+}
```
camscale, line 36: You need to add some test cases that start and/or end with multi-byte runes, as well as including strings containing multi-byte runes. Make sure you have three runes where some are multi-byte. By multi-byte I mean what people refer to as unicode characters - anything that is utf-8 encoded as multiple bytes such as emojis.

Have a read of https://blog.golang.org/strings for the gory details of strings.
I think you'll find some of your test case start failing in those cases.

04_numeronym/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func numeronyms(vals ...string) []string {
+	result := make([]string, len(vals))
+	for i, val := range vals {
+		if len(val) > 3 {
+			result[i] = val[0:1] + strconv.Itoa(len(val)-2) + val[len(val)-1:]
+		} else {
```
camscale: Rather than an `else` here, you could just assign to `val` in the `if` case and assign to `result[i]` unconditionally after it. That more clearly expresses the intent that the value is modified if the length is greater than 3, and is unchanged otherwise. The assignment to `result[i]` is unconditional. Just a small point but a lot of Go code tries to keep as much code further left as possible, leaving the indented code as the more special cases.

04_numeronym/davidbroughsmyth/main_test.go
camscale, line 36: You should have a test with string of three runes, as that is a boundary condition. This ensures the function is not trying to shorten strings of 3 runes and that the condition for that case is implemented properly.

camscale (dismissed):
A couple of extensions to the test cases will help, one of which will reveal a bug in your code. Have a read of the linked blog post for more details, and if that's not clear, let me know and I can explain the issue.

04_numeronym/davidbroughsmyth/main_test.go
davidbroughsmyth: Oh, I see that now.

davidbroughsmyth: Thanks, the link gave me the ideas to fix this function.
I have refactored this solution to handle Unicode. And provide test cases for the multi-byte runes.

davidbroughsmyth: I've added tests to handle the three rune boundary condition.

04_numeronym/davidbroughsmyth/main.go
davidbroughsmyth: Done.

davidbroughsmyth:
Refactored to handle unicode with supporting test cases.

04_numeronym/davidbroughsmyth/main.go
```diff
@@ -0,0 +1,28 @@
...
+	for i, val := range vals {
+		c := utf8.RuneCountInString(val)
+		if c > 3 {
+			f, _ := utf8.DecodeRuneInString(val)
+			l, _ := utf8.DecodeLastRuneInString(val)
+			val = fmt.Sprintf("%c%d%c", f, c-2, l)
+		}
+		result[i] = val
+	}
+	return result
```
juliaogris, line 23: IMO it's nicer to factor this assignment into two functions: `numeronyms` and `numeronym`:
```go
func numeronyms(vals ...string) []string {
	result := make([]string, len(vals))
	for i, val := range vals {
                result[i] =  numeronym(val)
        }
        return result
}

func numeronyms(s string) string {
    // ...
}
```

juliaogris (approved):
LGTM, just one little stylistic comment FYI
I've also updated the Markdown formatting of this PR's description (no blank line after bulletins).

04_numeronym/davidbroughsmyth/main.go
davidbroughsmyth: I did consider this but given how simple the function is, all the workings of this single function I can keep in my head. To me, this is cleaner and has greater readability.

PR #388, opened 2019-06-15
lab2 - implemented bubble sorting and insertion sorting - alfredxiao

alfredxiao:
Fixes #414
Review colleague's PR #293 

Changes proposed in this PR:
- bubble sort
- insertion sort

PR #549, opened 2019-07-08
Lab5 - Add IPAddr type and implement fmt.Stringer

### LINKED ISSUE #548: Lab5 - Add Stringer

davidbroughsmyth:
Fixes #548 

Review of colleague's PR #388 

- Create new IPAddr type `IPAddr [4]byte`
- create fmt.Stringer to print IP address as a dotted quad

### DIFF
--- a/05_stringer/davidbroughsmyth/main_test.go
+++ b/05_stringer/davidbroughsmyth/main_test.go
@@ -1,0 +1,40 @@
+package main
+
+import (
+	"bytes"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := strconv.Quote("127.0.0.1\n")
+	actual := strconv.Quote(buf.String())
+	assert.Equal(t, expected, actual, "Unexpected output from main()")
+}
+
+var tests = map[string]struct {
+	input IPAddr
+	want  string
+}{
+	"Empty":     {input: IPAddr{}, want: "0.0.0.0"},
+	"Loopback":  {input: IPAddr{127, 0, 0, 1}, want: "127.0.0.1"},
+	"Partial":   {input: IPAddr{0, 255}, want: "0.255.0.0"},
+	"Broadcast": {input: IPAddr{255, 255, 255, 255}, want: "255.255.255.255"},
+	"Zero mask": {input: IPAddr{0, 0, 0, 0}, want: "0.0.0.0"},
+	"Local lan": {input: IPAddr{192, 168, 0, 1}, want: "192.168.0.1"},
+}
+
+func TestIPAddr(t *testing.T) {
+	for name, test := range tests {
+		test := test
+		t.Run(name, func(t *testing.T) {
+			got := test.input.String()
+			assert.Equal(t, test.want, got)
+		})
+	}
+}
--- a/05_stringer/davidbroughsmyth/main.go
+++ b/05_stringer/davidbroughsmyth/main.go
@@ -1,0 +1,19 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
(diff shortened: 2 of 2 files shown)


05_stringer/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"Broadcast": {input: IPAddr{255, 255, 255, 255}, want: "255.255.255.255"},
+	"Zero mask": {input: IPAddr{0, 0, 0, 0}, want: "0.0.0.0"},
+	"Local lan": {input: IPAddr{192, 168, 0, 1}, want: "192.168.0.1"},
+}
+
+func TestIPAddr(t *testing.T) {
+	for name, test := range tests {
+		testList := test
+		t.Run(name, func(t *testing.T) {
+			got := test.input.String()
```
golangcibot, line 36: Using the variable on range scope `test` in function literal (from `scopelint`)

05_stringer/davidbroughsmyth/main_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	expected := strconv.Quote("127.0.0.1\n")
+	actual := strconv.Quote(buf.String())
```
pfannerj, line 16: This is the only thing I could find to comment on... I don't think strconv is required in this case.  It can just be 
`expected := "127.0.0.1\n"
actual := buf.String()`

pfannerj:
This all LGTM!  One observation is noted in main_test.go but the code and tests are thorough.

05_stringer/davidbroughsmyth/main_test.go
davidbroughsmyth: Thanks for the feedback. You are right, I've tried out your suggestion and it works fine.
I will look out for this for future labs.

camscale: @pfannerj the `strconv.Quote` can still be useful even in these simple cases as it makes unprintable characters printable in the output. Technically `actual` and `expected` to not need to be quoted, and only quoted when printed - you can do this when you do direct comparisons and `Printf` if not using `testify/assert`, but since `assert` will print out the values, having them quoted may help. Consider the case where the wrong format specifier was used - e.g. `%c` instead of `%d` - this may result in unprintable or control characters in the `actual` string, and when printed, are hard to see. Quoting make it readable. I'm making the assumption that `testify` does not do any quoting when outputting the values though. Maybe it does.

camscale (dismissed):
The code is fine, but I would like to see the commit messages formatted properly as requested in your other PRs: https://go-course.org/gocourse-2019-05-melbourne.slide#31

You can also practice a squash-rebase-force-push here too. We don't want to merge fixup commits into master so you can either squash merge when merging, or do a rebase-squash and force push. This results in a clean history with all mistakes erased (unlike real life :) )

05_stringer/davidbroughsmyth/main_test.go
davidbroughsmyth: In some of the other labs had that exact issue when using `assert`, the `strconv.Quote` made the values as printable characters. I also like how clean the code looks when using `assert`.

05_stringer/davidbroughsmyth/main_test.go
davidbroughsmyth: Rework the pin variable

05_stringer/davidbroughsmyth/main_test.go
davidbroughsmyth: The consensus is then to use` strconv.Quote`. I'll make no changes here.

davidbroughsmyth:
I have done a fixup on the commit messages and rebase-squash the commits. 

Also from the review from my Lab3 I've taken on the suggestion to rework the test case loop variable pinned with `test := test`

juliaogris (approved):
LGTM, nice work.

PR #371, opened 2019-06-13
Lab 2 - Bubble Sort

k-vikingsson:
Fixes #370 

Review of colleague's PR #

PR #554, opened 2019-07-12
Lab2: implement bubble sort

### LINKED ISSUE #560: Lab 2 - Bubble sort
Write a function that returns a sorted copy of `int slice s` using bubble sort:
`func bubble(s []int) []int`

`[1 2 3 5]`

ailyntang:
Fixes #560 
Review of colleague's PR #371 

* Implement bubble sort
* Implement tests for bubble sort

### DIFF
--- a/02_bubble/ailyntang/sort.go
+++ b/02_bubble/ailyntang/sort.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+	sCopy := make([]int, len(s))
+	copy(sCopy, s)
+
+	for i := len(s); i > 0; i-- {
+		swapped := false
+
+		for j := 1; j < i; j++ {
+			if sCopy[j-1] > sCopy[j] {
+				sCopy[j-1], sCopy[j] = sCopy[j], sCopy[j-1]
+				swapped = true
+			}
+		}
+
+		if !swapped {
+			break
+		}
+	}
+	return sCopy
+}
--- a/02_bubble/ailyntang/sort_test.go
+++ b/02_bubble/ailyntang/sort_test.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSortEdgeCases(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{}, bubble([]int{}), "empty slice as input")
+	assert.Equal([]int{4}, bubble([]int{4}), "one integer as input")
+	assert.Equal([]int{-8, 3}, bubble([]int{3, -8}), "two numbers as input")
+}
+
+func TestSortInputOptions(t *testing.T) {
+	assert := assert.New(t)
+	longInput := []int{999, 888, 777, 666, 555, 444, 333, 222, 111, 99, 88, 77, 66, 55, 44, 33, 22, 11, 1}
+	longOutput := []int{1, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111, 222, 333, 444, 555, 666, 777, 888, 999}
+
+	assert.Equal([]int{1, 11, 111, 222, 333}, bubble([]int{1, 11, 111, 222, 333}), "input is already sorted")
+	assert.Equal([]int{11, 22, 33, 44}, bubble([]int{44, 33, 22, 11}), "input is reverse sorted")
+	assert.Equal([]int{-88, -3, 0, 3, 88}, bubble([]int{88, -3, -88, 0, 3}), "positive and negative numbers")
+	assert.Equal([]int{-5, -5, -5, -2, -2, 3}, bubble([]int{-2, 3, -2, -5, -5, -5}), "duplicate numbers in input")
+	assert.Equal(longOutput, bubble(longInput), "sort 19 numbers")
+}
+
+func TestMainOutput(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	assert.Equal(t, "[1 2 3 5]\n", buf.String())
+}


02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,44 @@
...
+			copy(beforeSort, afterSort)
+			sortAdjacentPairs(afterSort)
+		}
+	}
+
+	return afterSort
+}
+
+func sortAdjacentPairs(s []int) []int {
+	for i := 0; i < len(s) - 1; i++ {
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	for i := 0; i < len(s)-1; i++ {
```

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSortingEdgeCases(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{}, bubble([]int{}), "empty array as input")
+	assert.Equal([]int{4}, bubble([]int{4}), "one integer as input")
+	assert.Equal([]int{0}, bubble([]int{0}), "0 as input")
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	assert.Equal([]int{0}, bubble([]int{0}), "0 as input")
```

golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	assert.Equal([]int{0}, bubble([]int{0}), "0 as input")
```

golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	assert.Equal([]int{0}, bubble([]int{0}), "0 as input")
```

go.sum
```diff
@@ -10,6 +10,7 @@ github.com/onsi/ginkgo v1.7.0/go.mod h1:lLunBs/Ym6LB5Z9jYTR76FiuTmxDTDusOGeTQH+W
 github.com/onsi/gomega v1.4.3/go.mod h1:ex+gbHU/CVuBBDIJjb2X0qEXbFg53c61hWP/1CpauHY=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/objx v0.1.0 h1:4G4v2dO3VZwixGIRoQ5Lfboy6nUhCyYzaqnIAPPhYs4=
```
krnicolson: Was this intentionally committed? I don't think it's necessary

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+func TestSorting(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{1, 11, 111, 222, 333}, bubble([]int{1, 11, 111, 222, 333}), "input is already sorted")
+	assert.Equal([]int{11, 22, 33, 44}, bubble([]int{44, 33, 22, 11}), "input is reverse sorted")
+	assert.Equal([]int{1, 11, 22, 55}, bubble([]int{11, 55, 22, 1}), "positive numbers")
+	assert.Equal([]int{-33, -22, -1}, bubble([]int{-22, -33, -1}), "negative numbers")
+	assert.Equal([]int{-88, -3, 0, 3, 88}, bubble([]int{88, -3, -88, 0, 3}), "positive and negative numbers")
+	assert.Equal([]int{-5, -5, -5, -2, -2, 3}, bubble([]int{-2, 3, -2, -5, -5, -5}), "duplicate numbers in input")
```
krnicolson, line 26: I would suggest including a test case with a large input array. I did this with your solution and found that it doesn't actually sort correctly. If you test with an input array of size ten at least, you should see some failures

02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,44 @@
...
+		if num != beforeSort[i] {
+			copy(beforeSort, afterSort)
+			sortAdjacentPairs(afterSort)
+		}
+	}
+
+	return afterSort
+}
+
+func sortAdjacentPairs(s []int) []int {
```
krnicolson: This function looks good to me. I think it does what it's meant to but it would be a good idea to write some tests for the function in isolation to be sure it does what's expected

02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,44 @@
...
+
+	beforeSort := make([]int, len(s))
+	copy(beforeSort, s)
+	afterSort := make([]int, len(s))
+	copy(afterSort, s)
+
+	sortAdjacentPairs(afterSort)
+
+	for i, num := range afterSort {
+		if num != beforeSort[i] {
```
krnicolson: I don't think this for loop is doing what you expect. It checks at indexes in the sorted slice and then calls the sorting function again if numbers don't match after sorting. But the fact that there has been a swap doesn't actually mean another sort is required. I would suggest addressing my other comments about test coverage etc and then you'll be in a better place to debug this logic and figure what it's doing wrong.

krnicolson (changes requested):

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+func TestSortAdjacentPairsInputOptions(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{4, 3, 2, 1, 5}, sortAdjacentPairs([]int{5, 4, 3, 2, 1}), "all pairs need swapping")
+	assert.Equal([]int{1, 2, 3, 4, 5}, sortAdjacentPairs([]int{2, 1, 3, 4, 5}), "only first pair needs swapping")
+	assert.Equal([]int{1, 2, 3, 4, 5}, sortAdjacentPairs([]int{1, 2, 3, 5, 4}), "only last pair needs swapping")
+	assert.Equal([]int{1, 2, 3, 4, 5}, sortAdjacentPairs([]int{1, 2, 4, 3, 5}), "only middle pair needs swapping")
+	assert.Equal([]int{11, -33, 0, 22, -3, 55}, sortAdjacentPairs([]int{11, 55, -33, 0, 22, -3}), "positive and negative numbers")
```
golangcibot: line is 127 characters (from `lll`)

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,54 @@
...
+	assert.Equal([]int{4, 5, 3, 2, 3, 5}, sortAdjacentPairs([]int{5, 4, 5, 3, 2, 3}), "duplicate numbers in input")
+	assert.Equal([]int{11, 33, 111, 22, 1, 222, 44, 555, 77, 333, 777, 66, 99, 666, 88, 888}, sortAdjacentPairs([]int{11, 222, 33, 111, 22, 1, 777, 44, 555, 77, 333, 888, 66, 99, 666, 88}), "swapping with large number of elements in array")
```
golangcibot: line is 237 characters (from `lll`)

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestSortInputOptions(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{1, 11, 111, 222, 333}, bubble([]int{1, 11, 111, 222, 333}), "input is already sorted")
+	assert.Equal([]int{11, 22, 33, 44}, bubble([]int{44, 33, 22, 11}), "input is reverse sorted")
+	assert.Equal([]int{1, 11, 22, 55}, bubble([]int{11, 55, 22, 1}), "positive numbers")
+	assert.Equal([]int{-33, -22, -1}, bubble([]int{-22, -33, -1}), "negative numbers")
+	assert.Equal([]int{-88, -3, 0, 3, 88}, bubble([]int{88, -3, -88, 0, 3}), "positive and negative numbers")
+	assert.Equal([]int{-5, -5, -5, -2, -2, 3}, bubble([]int{-2, 3, -2, -5, -5, -5}), "duplicate numbers in input")
+	assert.Equal([]int{1, 11, 22, 33, 44, 55, 66, 77, 88, 99, 111, 222, 333, 444, 555, 666, 777, 888, 999}, bubble([]int{999, 888, 777, 666, 555, 444, 333, 222, 111, 99, 88, 77, 66, 55, 44, 33, 22, 11, 1}), "sort 19 numbers")
```
golangcibot: line is 222 characters (from `lll`)

go.sum
ailyntang: Thanks for letting me know. I wasn't sure if I should commit it. I've reverted it now.

02_bubble/ailyntang/sort_test.go
ailyntang: Great pick up, thanks for that. I've added an array with 19 elements to the test case. I'll keep this in mind for future labs and tests.

02_bubble/ailyntang/sort.go
ailyntang: Done, thanks. I'll remember to add tests for all standalone functions in the future.

02_bubble/ailyntang/sort.go
ailyntang: Thanks for the call out. I realised I needed to reset the index of the for loop so that every element of the before and after slices are properly compared to each other.

jessodri (dismissed):
This looks good to me, I would just suggest that since you have a lot of commits squash them down so you only have 1 or 2.

krnicolson (dismissed):

ailyntang:
> This looks good to me, I would just suggest that since you have a lot of commits squash them down so you only have 1 or 2.

Thanks very much for reviewing @jessodri . I've enabled squash merging in my settings, so hopefully this happens automatically. Great call out though, I'll remember to squash for future projects.

ailyntang:
> > This looks good to me, I would just suggest that since you have a lot of commits squash them down so you only have 1 or 2.
> 
> Thanks very much for reviewing @jessodri . I've enabled squash merging in my settings, so hopefully this happens automatically. Great call out though, I'll remember to squash for future projects.

I realised I misunderstood what that setting does. I'll squash my commits now. Thanks for pointing this out!

02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,52 @@
...
+
+func bubble(s []int) []int {
+	if len(s) < 2 {
+		return s
+	}
+
+	beforeSort := make([]int, len(s))
+	copy(beforeSort, s)
+	afterSort := make([]int, len(s))
+	copy(afterSort, s)
```
juliaogris: It seems to me that you want to separate copying of the input (so you don't mutate the original input) and the sorting algorithm, which I think is a great idea.
You could do something like:

```go
func bubbleSort(s []int) []int {
	scopy := make([]int, len(s))
	copy(scopy, s)
	bubbleSortInPlace(scopy)
	return scopy
}

func bubbleSortInPlace(s []int) {
	for i := len(s); i > 0; i-- {
              // ...
	}
}

```

juliaogris (changes requested):
I'm a little confused by `bubble` and `sortAdjacentPairs` everything else looks good to me.

02_bubble/ailyntang/sort.go
ailyntang: Thanks for the feedback @juliaogris . You've got it right, I was trying to avoid mutating the input. I thought about how I could reduce the copies of the input from two slices to one, but based on the way I've implemented the sort, I don't think it will work unless I'm happy to mutate the original input.

I've added some comments to my code to make it more readable and to explain the difference between `bubble` and `sortAdjacentPairs`. 

It's different to the bubble sorting algorithms listed on the wiki. I look forward to hearing your thoughts. If it's still confusing, I'm happy to change it to a more traditional algorithm. 

Thanks again for reviewing.

juliaogris (changes requested):
I have commented on your comment ;)
I understand now where the copies are coming from, thanks for explaining.
I still feel you make things unnecessarily complex and runtime in-efficient and you are missing one key part in bubble sort - having the inner loop only iterate to `length`, `length-1`, `length-2`, ...,  

Feel free to give me further push back ;)

ailyntang:
@juliaogris Thanks very much for the feedback. I had forgotten about optimising for efficiency with bubble sort. That makes a lot of sense. Hopefully the new code is less complex and more efficient.

juliaogris (dismissed):
Sorry, but I'm still confused by this PR.
Bubble sort as I know it usually looks very similar to this:

```go
for i := len(s); i > 0; i-- {
	for j := 1; j < i; j++ {
		if s[j-1] > s[j] {
			s[j-1], s[j] = s[j], s[j-1]
		}
	}
}
```
and if you want to add the early bail it would be:
```go
for i := len(s); i > 0; i-- {
	swapped := false
	for j := 1; j < i; j++ {
		if s[j-1] > s[j] {
			s[j-1], s[j] = s[j], s[j-1]
			swapped = true
		}
	}
	if !swapped {
		return s
	}
}
```
I think your algorithm does something fairly similar but it is less obvious what is actually going on (at least to me)

You don't need the early return at the beginning of the function either. Copying a 1 element slice incurs a negligible performance increase, but taking out all special cases from your code and unnecessary assignments make it more readable.

See if you can incorporate some of the suggested changes please.

Also please fold up your fixup commits into the initial one.
This PR should really only consist of one, maximum two commits.

Thanks for hanging in there :)

02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,31 @@
...
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+	for i := len(s); i > 0; i-- {
+		swapped := false
+
+		for j := 1; j < i; j++ {
+			if s[j-1] > s[j] {
+				s[j-1], s[j] = s[j], s[j-1];
```
golangcibot: File is not `goimports`-ed (from `goimports`)

ailyntang:
@juliaogris I've updated the code to be a traditional bubble sorting algorithm, as per your feedback. Before I was trying to do it from first principles, and also to implement something more efficient. Instead of ignoring only one fewer element each time, I was ignoring potentially more than that, based on the index of the last pair of elements that were sorted.

However it's clear my code was confusing. A good learning for me not to try to reinvent the wheel.

Thank you for being patient with this PR and articulating your thoughts. I really appreciate it.

02_bubble/ailyntang/sort.go
```diff
@@ -0,0 +1,31 @@
...
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
```
n0npax, line 15: Due to: https://github.com/anz-bank/go-course/blob/master/02_bubble/README.md
You should be working on slice copy (`Write a function that returns a sorted copy of int slice s using bubble sort`)

Please consider adding:
`ts := append(s[:0:0], s...)`
and working on `ts` instead of `s`

n0npax (changes requested):
please copy input slice.

02_bubble/ailyntang/sort.go
ailyntang: Thanks for the pickup, I forgot about the copy.

ailyntang:
> please copy input slice.

@n0npax Thanks for the feedback and the reminder, I have made that change.

n0npax:
LGTM.

No action needed:
* As small nits: you can remove blank lines from bubble functions. They are not needed in my mind.

n0npax (dismissed):

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSortEdgeCases(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{}, bubble([]int{}), "empty array as input")
+	assert.Equal([]int{4}, bubble([]int{4}), "one integer as input")
+	assert.Equal([]int{0}, bubble([]int{0}), "0 as input")
```
camscale: I would add one more test to the edge cases:
A slice with 2 elements

This is one more boundary condition in your code. In the first one (empty slice), the outer loop should not be entered. In the second, the inner loop should not be entered. For a 2-element slice, the inner loop should be entered. I'd use an input that is not sorted to ensure you do the sorting.

I would also suggest removing some of the other tests - all negative and all positive do not work the code under test any differently. A mixture of positive and negative covers both of those other tests, and also validates that you are using the correct types and not mishandling negative numbers (although in a statically-typed language like Go, the compiler will typically pick up those sorts of problems).

02_bubble/ailyntang/sort_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSortEdgeCases(t *testing.T) {
+	assert := assert.New(t)
+
+	assert.Equal([]int{}, bubble([]int{}), "empty array as input")
```
camscale: technically this should be "empty slice" not "empty array".

camscale (dismissed):
My comments on the test cases relate to general software engineering, and are not Go-specific, so I'm happy to approve this PR for merging. If you do want to re-work the PR in response to those comments, let me know and I'll re-review quickly.

02_bubble/ailyntang/sort_test.go
ailyntang: Thanks for the explanation about testing the different `for` loops, and about the redundancy of the other tests. 

Both make a lot of sense. I've updated the code for this.

02_bubble/ailyntang/sort_test.go
ailyntang: Thanks for the pickup. I've updated this.

ailyntang:
> My comments on the test cases relate to general software engineering, and are not Go-specific, so I'm happy to approve this PR for merging. If you do want to re-work the PR in response to those comments, let me know and I'll re-review quickly.

@camh-anz Thanks for the helpful comments and explanations. I've updated the code.

juliaogris (approved):
lgtm - thanks for hanging in there!

ailyntang:
> lgtm - thanks for hanging in there!

@juliaogris Thanks for being so patient! :) merging now

PR #569, opened 2019-08-02
Add travis check for bin files

### LINKED ISSUE #568: travis check for commited binary files

n0npax:
Fixes #568 

Review of colleague's PR: N/A

From time to time someone forgets to rm `bin` files from PR. We can use Travis to check this.


.travis.yml
```diff
@@ -7,6 +7,7 @@ before_script:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -x -- -b $(go env GOPATH)/bin v1.17.1
 script:
   - golangci-lint run
+  - find * -executable -type f  | xargs  -IITEM sh -c 'echo "FAIL - Binary file ITEM was commited)"; exit 1'
```
juliaogris: It would look good to me ;)
showed to @camh-anz who had a whole bunch of comments (they call him Mr. shell for a reason).
IIRC: should `find.` and a bunch of other things...

n0npax: I'd disagree with `find .`. I'd like to find files in all directories except hidden ones. In the example, I want to skip`.git`, and an asterisk evaluates to all not hidden files (depends on `noglob`)


If you prefer for loop instead of `xargs` I'm happy to provide one.
@camh-anz could you please tell me what should be fixed here?

camscale: Using `-executable` will exclude all files with the executable bit set, whether they are "binary" or not. This will prevent us adding shell scripts for instance. If you want to check for "binary" files (and what exactly is a binary these days anyway, with utf-8 and other unicode encodings for text), you will need to find another way. Be aware that people could be creating executables on a Mac or Windows, and that travis is running the checks on Linux, so just seeing if it would run would not work either.

This failure to properly detect binaries is my main objection to this. But there are some others.

You say you want to skip all hidden directories, and they say you want to skip just `.git`. You need to decide which you want and do that. I can't see that it makes sense to skip all hidden dirs if you just want to skip `.git`. Just skip `.git` if that's what you want. Also using a wildcard has hidden failure modes that is also sometimes exploitable. What if I add a file called `-delete`? You have no idea what is going to end up on your command line and the command line is not protected from that (usually a `--` can be used to signify no more arguments to protect against wildcards hijacking the command line).

Then there's the ordering - why do this quick check after running lint? Do the quick checks first - fail fast.

Then there are a couple of little things that are just a bit messy - two spaces before the pipe, one after. Two spaces before `-IITEM`. A closing paren in the error message with no opening paren. It's trivial but over time, these sloppy things accumulate and people just stop caring because it's all a mess.

Also, s/commited/committed/

`xargs` is fine. No need for a loop.

n0npax: @camh-anz Thanks for explanations.

I can check if a file is  binary using `file`
```
 n0npax@narrenturm  ~/Downloads   file -i restic_0.9.5_*
restic_0.9.5_darwin_amd64:      application/x-mach-binary; charset=binary
restic_0.9.5_openbsd_amd64:     application/x-executable; charset=binary
restic_0.9.5_windows_amd64.exe: application/x-dosexec; charset=binary
```
I hope you're ok with this solution.

Due to hidden directories. I don't expect any hidden directories except `.git`. Seems I've got the wrong assumption.

Sorry for double spaces, asterisk and `)`. Shame on me


I'll close this PR and open a new one with branch without the typo in name.

n0npax:
moving to #576

PR #573, opened 2019-08-09
Lab 7 - Errors

### LINKED ISSUE #572: Lab 7 - Errors

kasunfdo:
Fixes #572

Review of colleague's PR #574

- Implement errors
- Implement leveldb Storer
- Add locking for sync.Map

### DIFF
--- a/07_errors/kasunfdo/error.go
+++ b/07_errors/kasunfdo/error.go
@@ -1,0 +1,37 @@
+package main
+
+import "fmt"
+
+type ErrCode uint32
+
+type Error struct {
+	Message string
+	Code    ErrCode
+}
+
+const (
+	ErrInvalid  ErrCode = 400
+	ErrNotFound ErrCode = 404
+)
+
+func (e ErrCode) String() string {
+	switch e {
+	case ErrInvalid:
+		return "invalid input: %v"
+	case ErrNotFound:
+		return "not found: %v"
+	default:
+		return "error occurred"
+	}
+}
+
+func (e *Error) Error() string {
+	return e.Message
+}
+
+func NewError(code ErrCode, args ...interface{}) *Error {
+	return &Error{
+		Message: fmt.Sprintf(code.String(), args...),
+		Code:    code,
+	}
+}
--- a/07_errors/kasunfdo/error_test.go
+++ b/07_errors/kasunfdo/error_test.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestErrCodeString(t *testing.T) {
+	assert.Equal(t, "invalid input: %v", ErrInvalid.String())
+	assert.Equal(t, "not found: %v", ErrNotFound.String())
+
+	var ErrFoo ErrCode = 900
+	assert.Equal(t, "error occurred", ErrFoo.String())
+}
+
+func TestError(t *testing.T) {
+	err := NewError(ErrNotFound, "error message")
+	assert.Equal(t, "not found: error message", err.Error())
+}
--- a/07_errors/kasunfdo/main.go
+++ b/07_errors/kasunfdo/main.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	store := NewMapStore()
+	puppy := Puppy{Breed: "Husky", Colour: "White", Value: 4999.98}
+	id, _ := store.CreatePuppy(puppy)
+	fmt.Fprintf(out, "Puppy(%d) added to store\n", id)
+}
--- a/07_errors/kasunfdo/main_test.go
+++ b/07_errors/kasunfdo/main_test.go
@@ -1,0 +1,23 @@
+package main
+
+import (
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMainOutput(t *testing.T) {
+	os.RemoveAll("./data")
+	defer os.RemoveAll("./data")
+
+	var buf bytes.Buffer
+	out = &buf
+
+	main()
+
+	expected := "Puppy(1) added to store\n"
+	actual := buf.String()
+	assert.Equal(t, expected, actual)
+}
--- a/07_errors/kasunfdo/mapstore.go
+++ b/07_errors/kasunfdo/mapstore.go
@@ -1,0 +1,64 @@
+package main
+
+import "fmt"
+
+// MapStore is map based implementation of Storer interface
+type MapStore struct {
+	store  map[uint64]Puppy
+	nextID uint64
+}
+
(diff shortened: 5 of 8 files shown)


07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "fmt"
+
+const (
+	ErrInvalidVal  = 400
```
golangcibot: File is not `goimports`-ed (from `goimports`)

07_errors/kasunfdo/leveldbstore.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+func NewLeveldbStore(path, dataFileName string) (*LeveldbStore, error) {
+	store := &LeveldbStore{
+		count:         0,
+		path:          path,
+		dataFileName:  dataFileName,
+		indexFileName: "index.db", indexName: "last_index"}
```
juliaogris: each field on a new line, curly on a new line.

07_errors/kasunfdo/leveldbstore.go
```diff
@@ -0,0 +1,154 @@
...
+		return store, err
+	}
+	defer closeDB(db)
+
+	exists, err := db.Has([]byte(store.indexName), nil)
+	if !exists {
+		log.Printf("Initialized store with index: %v\n", store.count)
+		return store, nil
+	}
+	logrus.Debug(err)
```
juliaogris: Return on error instead?

07_errors/kasunfdo/leveldbstore.go
```diff
@@ -0,0 +1,154 @@
...
+
+	value, err := db.Get([]byte(store.indexName), nil)
+	logrus.Debug(err)
```
juliaogris: return on error?

07_errors/kasunfdo/leveldbstore.go
```diff
@@ -0,0 +1,154 @@
...
+
+	store.count, err = strconv.ParseUint(string(value), 10, 64)
+	log.Printf("Initialized store with index: %v", store.count)
```
juliaogris: return on error?

juliaogris: or panic.

07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
+package main
+
+import "fmt"
+
+const (
+	ErrInvalidVal = 400
```
juliaogris: `ErrInvalid` is enough I'd say

07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
...
+	ErrNotFound   = 404
+	ErrInternal   = 500
+	ErrOpenDB     = 501
+)
+
+var ErrStr = map[int]string{
+	ErrInvalidVal: "value cannot be negative",
```
juliaogris: Too specific. Should be "invalid input" + custom message when you create the error.

07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
...
+	ErrNotFound:   "puppy with id=%v not found",
```
juliaogris: Same, `not founc` or  is enough.

07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
...
+	ErrInternal:   "internal system error",
+	ErrOpenDB:     "error occurred while opening db %v",
```
juliaogris: fold it into `ErrInternal` or create `ErrSetup` IMO

07_errors/kasunfdo/error.go
```diff
@@ -0,0 +1,25 @@
...
+}
+
+func NewError(code int, args ...interface{}) *Error {
+	return &Error{Code: code, Message: fmt.Sprintf(ErrStr[code], args...)}
```
juliaogris: IMO this should be something like the following (doesn't compile as Cause is only partially folded in)/.
```go
type ErrCode uint8

const (
	ErrInvalidInput ErrCode = 400
        // ...
)

func (e ErrCode) String() string {
	switch e {
	case ErrInvalidInput:
		return "invalid input"
         // ....
         }
}

NewError(code int, msg string, args ...interface{}) *Error {
     return &Error{
		Message: fmt.Sprintf(format, args...),
		Code:    code,
    }
}

type Error struct {
	Message string
	Code    ErrCode
	Cause   error // can be useful if wrapping underlying error
}

func (e *Error) Error() string {
	msg := fmt.Sprintf("%s: %s", e.Code.String(), e.Message)
	if e.Cause == nil {
		return msg
	}
	return fmt.Sprintf("%s\n\t%s", msg, e.Cause.Error())
}

```

juliaogris (changes requested):
a few generic comments:
- Rework errors a little please
- Handle errors, at least return them. Don't just log them (of if so only on the very top level).
- Doc strings on public types and functions
- Please add comments from lab 6 here too.

07_errors/kasunfdo/leveldbstore.go
kasunfdo: Agree. I did the same initially. But could not find a way to test that scenario so testing could not cover it. Any tips to test that scenario?

kasunfdo: Hi @juliaogris ,

Any hint on testing error paths? I couldn't hit some of the error paths and that causes the build failure due to lack of code coverage.

go.mod
```diff
@@ -5,6 +5,7 @@ require (
 	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
```
nicko-lee: Was wondering why you committed this? You must have reason

nicko-lee (commented):
Very nice work doing the bonus requirement for LevelDB!

07_errors/kasunfdo/leveldbstore.go
kasunfdo: Done

07_errors/kasunfdo/error.go
kasunfdo: Done

07_errors/kasunfdo/error.go
kasunfdo: Changed

07_errors/kasunfdo/error.go
kasunfdo: Agreed.

07_errors/kasunfdo/error.go
kasunfdo: Changed

go.mod
kasunfdo: I had used logrus. However, I removed it as no longer in use. :)

juliaogris (approved):
Great work!

PR #469, opened 2019-06-21
Lab6

CameronSchafer:
Fixes #470

Review of colleague's PR #265

#### Changes proposed in this PR:

- Add puppy struct with CRUD functionality
- Add test cases using testify suite

PR #574, opened 2019-08-10
Create CRUD puppy solution with map and syncMap

### LINKED ISSUE #575: Lab 5 #574
Fix #574

hsy3418:
Fixes #575

Review of colleague's PR #469

- Create CRUD puppy solution with map and syncMap


06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewMapStore() *MapStore {
+	return &MapStore{
+		puppies: map[int32]Puppy{},
+	}
+}
+
+// CreatePuppy adds a nuw puppy to the puppies store
+func (m *MapStore) CreatePuppy(puppy Puppy) error {
+	if _, ok := m.puppies[puppy.ID]; !ok {
```
kasunfdo: `exists` would be better instead of `ok`

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+		m.puppies[puppy.ID] = puppy
+		return nil
+	}
+	return fmt.Errorf("puppy with %d ID already exists", puppy.ID)
+}
+
+// ReadPuppy retrieves the puppy for a given id from puppies store
+func (m *MapStore) ReadPuppy(id int32) (Puppy, error) {
+	if _, ok := m.puppies[id]; !ok {
```
kasunfdo, line 28: `exists` would be better instead of `ok`

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+		return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
+	}
+	return m.puppies[id], nil
+}
+
+//UpdatePuppy updates the puppy for the given id
+func (m *MapStore) UpdatePuppy(id int32, puppy Puppy) error {
```
kasunfdo: `exists` would be better instead of `ok`

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+	if _, ok := m.puppies[id]; !ok {
```
kasunfdo: What if id != puppy.ID?

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+	m.puppies[id] = puppy
+	return nil
+}
+
+//DeletePuppy delete the puppy for the given id from puppies store
+func (m *MapStore) DeletePuppy(id int32) (bool, error) {
+	if _, ok := m.puppies[id]; ok {
+		delete(m.puppies, id)
+		return true, nil
```
kasunfdo: Why return a bool and an error both?

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+		expected string
+	}{
+		{"Create new puppy", toBeCreatedPuppy(), "<nil>"},
+		{"Create existing puppy", toBeCreatedPuppy(), "puppy with 101 ID already exists"},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		suite.T().Run(tc.title, func(t *testing.T) {
+			err := suite.store.CreatePuppy(tc.input)
+			assert.Equal(tc.expected, fmt.Sprint(err))
```
kasunfdo: It would be better if you make `expected` type as `error` and compare them.

```go
func (suite *storerTestSuite) TestCreatePuppy() {
	assert := assert.New(suite.T())
	testCases := []struct {
		title    string
		input    Puppy
		expected error
	}{
		{"Create new puppy", toBeCreatedPuppy(), nil},
		{"Create existing puppy", toBeCreatedPuppy(), fmt.Errorf("puppy with 101 ID already exists")},
	}
	for _, tc := range testCases {
		tc := tc
		suite.T().Run(tc.title, func(t *testing.T) {
			err := suite.store.CreatePuppy(tc.input)
			assert.Equal(tc.expected, err)
		})
	}

}

```

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+}
+
+func (suite *storerTestSuite) TestUpdatePuppy() {
+	assert := assert.New(suite.T())
+	testCases := []struct {
+		title         string
+		inputID       int32
+		inputPuppy    Puppy
+		expectedError string
```
kasunfdo: type should be error

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+}
+
+func (suite *storerTestSuite) TestReadPuppy() {
+	assert := assert.New(suite.T())
+	testCases := []struct {
+		title         string
+		input         int32
+		expected      Puppy
+		expectedError string
```
kasunfdo: type should be error

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+}
+
+func (suite *storerTestSuite) TestDeletePuppy() {
+	assert := assert.New(suite.T())
+	testCases := []struct {
+		title         string
+		input         int32
+		expected      bool
+		expectedError string
```
kasunfdo: type should be error

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,55 @@
...
+func (m *SyncStore) ReadPuppy(id int32) (Puppy, error) {
+	if p, ok := m.Load(id); ok {
+		puppy, _ := p.(Puppy)
+		return puppy, nil
+	}
+	return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
+}
+
+//UpdatePuppy updates the puppy for the given id
+func (m *SyncStore) UpdatePuppy(id int32, puppy Puppy) error {
```
kasunfdo: use mutex lock here as well.

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/suite"
+)
+
+var (
```
kasunfdo: I'm curious why did you defined methods to create puppies as below. You could have stored them in variables in `suite` and defined at `SetupTest()`.


```go
func (suite *storerTestSuite) SetupTest() {
	suite.store = suite.makeStore()
	suite.toBeCreatedPuppy = Puppy{ID: 101, Breed: "Poodle", Colour: "White", Value: 1000.5}
	suite.existsPuppy = Puppy{ID: 102, Breed: "Poodle", Colour: "White", Value: 1280.5}
	err := suite.store.CreatePuppy(suite.existsPuppy)
	if err != nil {
		suite.FailNow("Failed to setup test")
	}
}

type storerTestSuite struct {
	suite.Suite
	store     Storer
	makeStore func() Storer
	toBeCreatedPuppy     Puppy
	existsPuppy     Puppy
}

func (suite *storerTestSuite) TestCreatePuppy() {
	assert := assert.New(suite.T())
	testCases := []struct {
		title    string
		input    Puppy
		expected error
	}{
		{"Create new puppy", suite.toBeCreatedPuppy, nil},
		{"Create existing puppy", suite.toBeCreatedPuppy, fmt.Errorf("puppy with 101 ID already exists")},
	}
	for _, tc := range testCases {
		tc := tc
		suite.T().Run(tc.title, func(t *testing.T) {
			err := suite.store.CreatePuppy(tc.input)
			assert.Equal(tc.expected, err)
		})
	}

}
```

kasunfdo (commented):
Looks good so far. Good unit test coverage. I have included few suggestions for improvements. The main concern I'm having is comparing error string in test.

06_puppy/hsy3418/main.go
```diff
@@ -0,0 +1,26 @@
...
+	syncStore := NewSyncStore()
+
+	_ = mapStore.CreatePuppy(p1)
+	_ = syncStore.CreatePuppy(p2)
+
+	puppy, _ := mapStore.ReadPuppy(101)
+	puppy2, _ := syncStore.ReadPuppy(102)
+	fmt.Fprintf(out, "Puppy ID %d is %v", puppy.ID, puppy.Value)
+	fmt.Fprintf(out, "Puppy ID %d is %v", puppy2.ID, puppy2.Value)
+
```
juliaogris: Unnecessary blank line.
(below as well)

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
...
+
+func NewMapStore() *MapStore {
+	return &MapStore{
+		puppies: map[int32]Puppy{},
+	}
+}
+
+// CreatePuppy adds a nuw puppy to the puppies store
+func (m *MapStore) CreatePuppy(puppy Puppy) error {
+	if _, exists := m.puppies[puppy.ID]; !exists {
```
juliaogris: use `ok` instead of `exisits` ([comma go idiom](https://golang.org/doc/effective_go.html#maps)) (below as well).

06_puppy/hsy3418/types.go
```diff
@@ -0,0 +1,17 @@
...
+type Puppy struct {
+	ID     int32   `json:"id"`
+	Value  float32 `json:"value"`
+	Breed  string  `json:"breed"`
+	Colour string  `json:"colour"`
+}
+
+//Storer define standard CRUD operations for puppys
+type Storer interface {
+	CreatePuppy(Puppy) error
```
juliaogris: IMO Create should provision a system unique ID and return it - the user/client of this API probably doesn't have complete knowledge of all other entries. Happy for you to argue against it (leave it).

juliaogris (changes requested):
Looks pretty solid, just a few minor nitpicks inline.

06_puppy/hsy3418/mapStore.go
juliaogris: @kasunfdo this a golang this: [comma ok idiom](https://golang.org/doc/effective_go.html#maps)
however, in my opinion you typically keep the error path inside the if so use `if  _, ok := m.puppies[puppy.ID]; ok { return err }`

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
...
+
+func NewSyncStore() *SyncStore {
+	return &SyncStore{}
+}
+
+// CreatePuppy adds a nuw puppy to the puppies store
+func (m *SyncStore) CreatePuppy(puppy Puppy) error {
+	m.Lock()
+	defer m.Unlock()
+	if _, exists := m.Load(puppy.ID); exists {
```
juliaogris: `ok` instead of exists :)

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
...
+		return fmt.Errorf("puppy with %d ID already exists", puppy.ID)
+	}
+	m.Store(puppy.ID, puppy)
+	return nil
+}
+
+// ReadPuppy retrieves the puppy for a given id from puppies store
+func (m *SyncStore) ReadPuppy(id int32) (Puppy, error) {
+	if p, exists := m.Load(id); exists {
```
juliaogris: ok instead of exists and invert logic so that error is inside if: if !ok return err

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
...
+		return puppy, nil
+	}
+	return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
+}
+
+//UpdatePuppy updates the puppy for the given id
+func (m *SyncStore) UpdatePuppy(puppy Puppy) error {
+	m.Lock()
+	defer m.Unlock()
+	if _, exists := m.Load(puppy.ID); !exists {
```
juliaogris: ok

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	m.Store(puppy.ID, puppy)
+	return nil
+}
+
+//DeletePuppy delete the puppy for the given id from puppies store
+func (m *SyncStore) DeletePuppy(id int32) error {
+	m.Lock()
+	defer m.Unlock()
+	if _, exists := m.Load(id); exists {
```
juliaogris: ok

06_puppy/hsy3418/types.go
juliaogris: resolving the comment is not exactly arguing against it ;) .
please put forward your case why you think the user/client of your system would have knowledge of (available) ID or change implementation to provision on ID (e.g. by incrementing a counter).

juliaogris (changes requested):
a few more small things.

06_puppy/hsy3418/syncStore.go
hsy3418: I need the p within the if block, so a puppy could be created if ok, otherwise I need to have an else block for create the puppy if ok, but if doing so, I couldn't pass lint..So it seems like I need to keep this logic.

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
...
+type SyncStore struct {
+	sync.Mutex
+	sync.Map
+}
+
+func NewSyncStore() *SyncStore {
+	return &SyncStore{}
+}
+
+// CreatePuppy adds a nuw puppy to the puppies store
```
rokane: adds a **new** puppy to the puppy store

06_puppy/hsy3418/syncStore.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import (
+	"fmt"
+	"sync"
+)
+
+type SyncStore struct {
```
rokane, line 9: You have done really well documenting your interface methods, but you forgot these as well. Since the SyncStore and NewSyncStore are also included as part of the exported package, these should also have documentation around them.

06_puppy/hsy3418/types.go
rokane: I agree with @juliaogris on this one. The user should not be expected to understand the complete set of entries. All they want to do is create a puppy, and then have some method of retrieving that puppy. Having the ID returned means the API handles that logic and the user doesn't need to worry about what else exists. They have a means of accessing their object through the id returned from a successful create. 

@hsy3418 if you don't agree or have some other view / ideas, please speak up and let us know. I would be happy to discuss and hear your thoughts. Alternatively, please update this one accordingly.

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,49 @@
+package main
+
+import "fmt"
+
+// MapStore is a implementation of storer for the storage of puppies
+type MapStore struct {
+	puppies map[int32]Puppy
+}
+
+func NewMapStore() *MapStore {
```
rokane, line 12: Again here, this is also a part of the exported package. Please document

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,120 @@
...
+		{"Create existing puppy", suite.toBeCreatedPuppy, fmt.Errorf("puppy with 101 ID already exists")},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		suite.T().Run(tc.title, func(t *testing.T) {
+			err := suite.store.CreatePuppy(tc.input)
+			assert.Equal(tc.expected, err)
+		})
+	}
+
```
rokane: Little nit. Another new line not required

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,120 @@
...
+		{"Update non-existing puppy", Puppy{}, fmt.Errorf("puppy with 0 ID does not exist")},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		suite.T().Run(tc.title, func(t *testing.T) {
+			err := suite.store.UpdatePuppy(tc.inputPuppy)
+			assert.Equal(tc.expectedError, err)
+		})
+	}
+
```
rokane: New line here can be removed.

06_puppy/hsy3418/syncStore.go
rokane: If you flip the logic, you won't need the p inside the if statement, and can go about it that way. 

```go
func (m *SyncStore) ReadPuppy(id int32) (Puppy, error) {
  p, ok := m.Load(id)
  if !ok {
    return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
  }
  puppy, _ := p.(Puppy)
  return puppy, nil
}
```

I think what Julia is suggesting here is to have your structure following the convention of: **if err -> fail early**

Normally if there are multiple error cases, I think this is a good approach. When there is a single error case, I am not as fussed either way. I will leave this one up to you.

rokane (changes requested):
Code and tests are looking really good, you've done a great job so far! I have a few comments throughout which I would like addressed :) 

A lot of these have already been requested from previous reviewers, but you have resolved the conversations without applying changes. I would advise against doing this. Please in future either make a case for why you think it should remain, which can spark discussion on the review. Or make the changes and then resolve. But don't get in the habit of resolving the conversation without making any changes.

Main things from me are:
* Full set of documentation
* Removing extra new lines from each test case
* Adjusting the interface API so that an id is returned on `CreatePuppy`

06_puppy/hsy3418/types.go
hsy3418: @juliaogris @rokane  I totally agree with you the client who is going to create the puppy wouldn't know the ID, they will just enter the puppy info and ID should be handled in the back-end logic, didn't think about that deeply before.

hsy3418: By returning the ID as a sign for creating the puppy successfully make sense to the client definitely, I think I will change the method accordingly, but my question is if there is no error return, would it work similarly to the client? Cause my understanding is if the client just wants to know if the puppy been created or not, they don't really need the ID, unless the front-end logic needs to show the exact ID. Correct me if I am wrong.

hsy3418: I am going to take that back :), just rethink about it, since the ID is created in backend, there is no error could be returned since there wouldn't be duplicate puppys.

06_puppy/hsy3418/syncStore.go
hsy3418: Got it!

06_puppy/hsy3418/main.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	p1 := Puppy{Breed: "Poodle", Colour: "White", Value: 1280.5}
+	p2 := Puppy{Breed: "Poodle", Colour: "Grey", Value: 1340.5}
+	mapStore := NewMapStore()
+	syncStore := NewSyncStore()
+	_ = mapStore.CreatePuppy(p1)
```
camscale, line 16: you should save the id from `CreatePuppy` and use that to do the `ReadPuppy` and not just make the assumption that the puppy ID will be zero. As `main()` is a user of the store, it should not make assumptions about the internal workings of the store. The store could start IDs at 1, not 0. Since there is no need to assume because `CreatePuppy` returns the ID, you should just use that ID.

06_puppy/hsy3418/main.go
```diff
@@ -0,0 +1,22 @@
...
+	_ = syncStore.CreatePuppy(p2)
+	puppy, _ := mapStore.ReadPuppy(0)
+	puppy2, _ := syncStore.ReadPuppy(0)
+	fmt.Fprintf(out, "Puppy ID %d is %v", puppy.ID, puppy.Value)
```
camscale, line 20: Put a `\n` at the end of this output and the next. This should output a complete line to be user-friendly.

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,50 @@
...
+
+// ReadPuppy retrieves the puppy for a given id from puppies store
+func (m *MapStore) ReadPuppy(id int32) (Puppy, error) {
+	if _, ok := m.puppies[id]; !ok {
+		return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
+	}
+	return m.puppies[id], nil
+}
+
+//UpdatePuppy updates the puppy for the given id
```
camscale, line 34: nit: Put a space after `//` like you have for the doc strings above. Do the same for the next comment and also for the comments in `syncStore.go`

06_puppy/hsy3418/mapStore.go
```diff
@@ -0,0 +1,50 @@
...
+	m.puppies[puppy.ID] = puppy
+	return puppy.ID
+}
+
+// ReadPuppy retrieves the puppy for a given id from puppies store
+func (m *MapStore) ReadPuppy(id int32) (Puppy, error) {
+	if _, ok := m.puppies[id]; !ok {
+		return Puppy{}, fmt.Errorf("puppy with %d ID does not exist", id)
+	}
+	return m.puppies[id], nil
```
camscale, line 31: Instead of looking up the map again, when you did it already before, you should save the value from from the first lookup. It means you can't put it all inside the `if`, but that doesn't matter:
```
puppy, ok := m.puppies[id]
if !ok {
    ...
}
return puppy, nil
```

06_puppy/hsy3418/store_test.go
```diff
@@ -0,0 +1,116 @@
...
+	for _, tc := range testCases {
+		tc := tc
+		suite.T().Run(tc.title, func(t *testing.T) {
+			err := suite.store.DeletePuppy(tc.input)
+			assert.Equal(tc.expectedError, err)
+		})
+	}
+}
+
+func TestStorers(t *testing.T) {
```
camscale, line 109: I think it is more readable if this function is placed at the top of the file before `SetupTest`. This is where the tests in this file start, so it reads better to start here, and then move onto `SetupTest`, and then finally the tests themselves.

camscale (changes requested):
This is very good. Only minor comments on the code.

However, this PR also contains some changes that should not be here. Have a look at the "Files changed" tab on this PR in guthub and the first five changed files should not be here:

1. Your change to `.gitignore` is for your personal setup, so should not go into the repository. Also, even if it made sense to add, you just put at at the end of the file under the comment `# Output of the go coverage tool, specifically when used with LiteIDE`. That would be in the wrong spot even if it were valid to put in. Please pay attention to the details - details matter when writing software.

2. You've changed the mode of some files in `00_hello_world/juliaogris`. Please don't add this change to your commit.

3. You've deleted your lab 3 submission in `03_letters/hsy3418`. Please don't do that in this commit (or do it at all).

Apart from these git issues, the code itself is looking very good. Thanks.

PR #576, opened 2019-08-10
Check if binary file was committed

### LINKED ISSUE #568: travis check for commited binary files

n0npax:
Fixes #568

Continuation of: #569 - fixed branch name
Review of colleague's PR: N/A

Changes proposed in this PR:
From time to time someone forgets to rm bin files from PR. We can use Travis to check this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ go:
 before_script:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -x -- -b $(go env GOPATH)/bin v1.17.1
 script:
+  - for f in $(find -executable -type f); do file -i "${f}" | awk '{print $NF}' | grep -q binary && { echo "FAIL - bin file found (${f})"; exit 1; }; done; echo "OK - no bin files"
   - golangci-lint run
   - go test -v -race -coverprofile=coverage.out -covermode=atomic ./...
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,6 +6,7 @@ go:
 before_script:
   - curl -sfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -x -- -b $(go env GOPATH)/bin v1.17.1
 script:
+  - for f in $(find -executable -type f); do file -i "${f}" | awk '{print $NF}' | grep -q binary && { echo "FAIL - bin file found (${f})"; exit 1; }; done; echo "OK - no bin files"
```
camscale, line 9: putting `$(find -executable -type f)` on the command line will break if any files have spaces in their name. You can use `find ... -print0 | xargs -0` to do this safely.

I also suggest getting rid of awk and just grepping for `charset=binary$` or using awk to do it all (it can grep and print too).

But this is getting too much to put on one line of a yaml file. it feels like it belongs in a script now.

Finally, I'm still not convinced that this is a problem to solve this way. What if I need to check in an image (gif, png, jpeg) - since we have a webserver, maybe I want to serve a favicon? Will this block it? Can't we rely on reviewers to make sure the binaries do not make it in?

n0npax: @camh-anz Thank you for explanation. Fair enough, It can relay on manual review.
Closing PR.

PR #590, opened 2019-08-19
Implement custom error handling to Storer interface

### LINKED ISSUE #589: Lab 7 - Errors
# Lab 7 - Errors

- Create errors for:
        - Value < 0
- Bonus points: Add a third `Storer` implementation using [leveldb](https://github.com/syndtr/goleveldb)

patrickmarabeas:
Implements custom error handling to the Storer interface. All methods must now return an error.

Errors that are covered:
- Puppy value < 0
- Puppy ID not found

Fixes / improvements also included:
- The value property in Puppy has had its type changed to an int to represent a numerical cent figure so that it may be properly throw an error if the value supplied is less than 0
- Proper locking in SyncMap

Fixes #589

### DIFF
--- a/07_errors/patrickmarabeas/errors.go
+++ b/07_errors/patrickmarabeas/errors.go
@@ -1,0 +1,32 @@
+package main
+
+import "fmt"
+
+type ErrorCode int
+
+type Error struct {
+	Code    ErrorCode
+	Message string
+}
+
+const (
+	NegativeValue ErrorCode = iota + 1001
+	IDNotFound
+	Unknown = 1999
+)
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("[%d] %s", e.Code, e.Message)
+}
+
+// NewError creates a new error with the given enum
+func NewError(code ErrorCode) error {
+	switch code {
+	case NegativeValue:
+		return &Error{code, "Puppy value must be greater than 0"}
+	case IDNotFound:
+		return &Error{code, "Nonexistent Puppy ID"}
+	default:
+		return &Error{Unknown, "Unknown error"}
+	}
+}
--- a/07_errors/patrickmarabeas/errors_test.go
+++ b/07_errors/patrickmarabeas/errors_test.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestUnknownError(t *testing.T) {
+	a := assert.New(t)
+
+	error := NewError(123)
+	a.Equal(error, NewError(Unknown))
+}
+
+func TestError(t *testing.T) {
+	a := assert.New(t)
+
+	error := Error{1, "message"}
+	formattedError := error.Error()
+	a.Equal(formattedError, "[1] message")
+}
--- a/07_errors/patrickmarabeas/main.go
+++ b/07_errors/patrickmarabeas/main.go
@@ -1,0 +1,5 @@
+package main
+
+func main() {
+
+}
--- a/07_errors/patrickmarabeas/main.go
+++ b/07_errors/patrickmarabeas/main.go
@@ -1,5 +1,13 @@
 package main
 
-func main() {
+import (
+	"fmt"
+	"io"
+	"os"
+)
 
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, "hello")
 }
--- a/07_errors/patrickmarabeas/mapStore.go
+++ b/07_errors/patrickmarabeas/mapStore.go
@@ -9,43 +9,50 @@ func NewMapStore() Storer {
 }
 
 // Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
-func (store *MapStore) Create(puppy Puppy) int {
+func (store *MapStore) Create(puppy Puppy) (int, error) {
+	if puppy.Value < 0 {
+		return -1, NewError(NegativeValue)
+	}
+
 	puppy.ID = store.uuid
 	store.store[puppy.ID] = puppy
 	store.uuid++
 
-	return puppy.ID
+	return puppy.ID, nil
 }
 
 // Read returns the puppy matching the provided uuid.
 // An empty Puppy struct is returned if the identifier does not exist.
-func (store *MapStore) Read(id int) Puppy {
+func (store *MapStore) Read(id int) (Puppy, error) {
 	if _, ok := store.store[id]; ok {
-		return store.store[id]
+		return store.store[id], nil
 	}
 
-	return Puppy{}
+	return Puppy{}, NewError(IDNotFound)
 }
 
 // Update modifies the puppy matching the provided uuid in the store with the provided Puppy struct.
 // It returns a bool whether a matching identifier was modified or not.
-func (store *MapStore) Update(id int, puppy Puppy) bool {
+func (store *MapStore) Update(id int, puppy Puppy) (bool, error) {
 	if _, ok := store.store[id]; !ok {
-		return false
+		return false, NewError(IDNotFound)
+	}
+	if puppy.Value < 0 {
+		return false, NewError(NegativeValue)
 	}
 
 	puppy.ID = id
 	store.store[id] = puppy
-	return true
+	return true, nil
 }
 
 // Destroy removes the puppy matching the provided uuid from the store.
 // It returns a bool whether a matching identifier was deleted or not.
-func (store *MapStore) Destroy(id int) bool {
+func (store *MapStore) Destroy(id int) (bool, error) {
 	if _, ok := store.store[id]; !ok {
-		return false
+		return false, NewError(IDNotFound)
 	}
 
 	delete(store.store, id)
-	return true
+	return true, nil
 }
--- a/07_errors/patrickmarabeas/mapStore.go
+++ b/07_errors/patrickmarabeas/mapStore.go
@@ -1,0 +1,51 @@
+package main
+
+// NewMapStore returns a pointer to a new instance of the MapStore struct which implements the Storer interface.
+func NewMapStore() Storer {
+	return &MapStore{
+		uuid:  0,
+		store: map[int]Puppy{},
+	}
+}
+
+// Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
+func (store *MapStore) Create(puppy Puppy) int {
+	puppy.ID = store.uuid
+	store.store[puppy.ID] = puppy
+	store.uuid++
+
+	return puppy.ID
+}
+
+// Read returns the puppy matching the provided uuid.
+// An empty Puppy struct is returned if the identifier does not exist.
+func (store *MapStore) Read(id int) Puppy {
+	if _, ok := store.store[id]; ok {
+		return store.store[id]
+	}
+
+	return Puppy{}
+}
+
+// Update modifies the puppy matching the provided uuid in the store with the provided Puppy struct.
+// It returns a bool whether a matching identifier was modified or not.
+func (store *MapStore) Update(id int, puppy Puppy) bool {
+	if _, ok := store.store[id]; !ok {
+		return false
+	}
+
+	puppy.ID = id
+	store.store[id] = puppy
+	return true
+}
+
+// Destroy removes the puppy matching the provided uuid from the store.
+// It returns a bool whether a matching identifier was deleted or not.
+func (store *MapStore) Destroy(id int) bool {
+	if _, ok := store.store[id]; !ok {
+		return false
+	}
+
+	delete(store.store, id)
+	return true
+}
--- a/07_errors/patrickmarabeas/store_test.go
+++ b/07_errors/patrickmarabeas/store_test.go
@@ -14,70 +14,99 @@ type StoreSuite struct {
 
 func (suite *StoreSuite) TestCreate() {
 	a := assert.New(suite.T())
-	id := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: "450"})
 
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: 450})
 	a.Equal(id, 0)
+	a.Equal(error, nil)
 }
 
 func (suite *StoreSuite) TestCreateSecond() {
 	a := assert.New(suite.T())
-	id := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: "300"})
 
+	id, error := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: 300})
 	a.Equal(id, 1)
+	a.Equal(error, nil)
+}
+
+func (suite *StoreSuite) TestCreateNegativeNumber() {
+	a := assert.New(suite.T())
+
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: -100})
+	a.Equal(id, -1)
+	a.Equal(error, NewError(NegativeValue))
 }
 
 func (suite *StoreSuite) TestRead() {
 	a := assert.New(suite.T())
-	data := suite.store.Read(0)
 
-	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: "450"})
+	data, error := suite.store.Read(0)
+	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: 450})
+	a.Equal(error, nil)
 }
 
 func (suite *StoreSuite) TestReadNonExistent() {
 	a := assert.New(suite.T())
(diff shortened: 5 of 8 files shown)


07_errors/patrickmarabeas/main.go
```diff
@@ -0,0 +1,13 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprint(out, "hello")
+}
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
}
```

07_errors/patrickmarabeas/errors.go
```diff
@@ -0,0 +1,32 @@
...
+
+type ErrorCode int
+
+type Error struct {
+	Code    ErrorCode
+	Message string
+}
+
+const (
+	NegativeValue ErrorCode = iota + 1001
```
camscale, line 13: It is not clear why you have decided to start at 1001 for the error codes, and also set `Unknown` to 1999. What if in future you need >1000 error codes? Since this is an enum, the actual values for the enum are not relevant, unless they map to some external system. In that case you should document that mapping. But if there is one, don't suggest there is one but you're not telling by using a different set of meaningless numbers to the default.

I'd suggest starting with "Unknown = iota" and going up from there. That does not give you that artificial limit of the number of error codes when you do eventually get to more than 1000 error codes (which will come up in lab 22).

Perhaps you have chosen to use 1000+ so you get a consistent formatting width when converting the error to a string, but I'd suggest you do formatting at the place where formatting is specified (i.e. `[%04d]`)

07_errors/patrickmarabeas/errors.go
```diff
@@ -0,0 +1,32 @@
...
+	IDNotFound
+	Unknown = 1999
+)
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("[%d] %s", e.Code, e.Message)
+}
+
+// NewError creates a new error with the given enum
```
camscale, line 22: nitpick: Godoc comments should end with a full-stop, as they appear as sentences in the documentation (https://blog.golang.org/godoc-documenting-go-code).

another nitpick: I'd say "given error code" instead of "given enum" - seems a little clearer to me.

07_errors/patrickmarabeas/errors.go
```diff
@@ -0,0 +1,32 @@
...
+func NewError(code ErrorCode) error {
+	switch code {
+	case NegativeValue:
+		return &Error{code, "Puppy value must be greater than 0"}
+	case IDNotFound:
+		return &Error{code, "Nonexistent Puppy ID"}
+	default:
+		return &Error{Unknown, "Unknown error"}
```
camscale, line 30: This throws away the error code passed in, which may be useful when trying to debug where a strange error is coming from. Perhaps set the string as `fmt.Sprintf("Unknown error (%d)", code)` if `code != Unknown`? Just an idea.

07_errors/patrickmarabeas/errors_test.go
```diff
@@ -0,0 +1,22 @@
...
+	a.Equal(error, NewError(Unknown))
+}
+
+func TestError(t *testing.T) {
+	a := assert.New(t)
+
+	error := Error{1, "message"}
+	formattedError := error.Error()
+	a.Equal(formattedError, "[1] message")
+}
```
camscale, line 22: I don't see that you have 100% test coverage with this commit. There are a couple of branches in `NewError()` that are not tested. They probably get tested in a later commit by using it, but since you have structured your PR as a series of small cohesive commits (good), you should ensure each commit passes all the tests and has complete coverage. There is some `git rebase --exec` magic you can use to run all your tests and coverage checks on every one of your commits automagically, but that usually requires a makefile or script to run all the checks you want to.

I will share the magic in a subsequent comment where I have noticed your tests failing...

07_errors/patrickmarabeas/types.go
```diff
@@ -1,21 +1,19 @@
...
-	"sync"
-)
+import "sync"
 
 type Puppy struct {
 	ID    int
 	Breed string
 	Color string
-	Value string
+	Value int // cents
```
camscale, line 9: Two ways you could convey this information without a comment:
1) Change the field name to `ValueCents`
2) Add a new type: `type Cents int`

Just a thought.

07_errors/patrickmarabeas/types.go
```diff
@@ -1,21 +1,19 @@
...
 }
 
 type Storer interface {
-	Create(puppy Puppy) int
-	Read(ID int) Puppy
-	Update(ID int, puppy Puppy) bool
-	Destroy(ID int) bool
+	Create(puppy Puppy) (int, error)
```
camscale, line 13: This commit on its own breaks the code. If you run `go test ./...` on this commit, the tests will fail  - the build will fail too. This is bad as it breaks `git bisect` - you want every commit to build and pass tests.

The most minimal version of this change would be to change the implementation to always return `nil` for the error when implementing it, and a subsequent commit can start returning non-nil errors where appropriate.

07_errors/patrickmarabeas/mapStore.go
```diff
@@ -9,43 +9,50 @@ func NewMapStore() Storer {
...
 // An empty Puppy struct is returned if the identifier does not exist.
-func (store *MapStore) Read(id int) Puppy {
+func (store *MapStore) Read(id int) (Puppy, error) {
 	if _, ok := store.store[id]; ok {
-		return store.store[id]
+		return store.store[id], nil
 	}
 
-	return Puppy{}
+	return Puppy{}, NewError(IDNotFound)
```
camscale, line 31: Two comments about this function (and I know my comments really apply to lab 6, not this one but here we are now):
1) Usually you want to align the happy path to the left edge, so the indented case should be returning the error,
2) Since you have the non-error path inside the if, rather than indexing the map again you should save the value rather than using the blank identifier:
```
if p, ok := store.store[id]; ok {
    return p, nil
}
```

unless you use a `var` to declare `p` outside the scope of the `if`. Combined, it would look like:
```
func (store *MapStore) Read(id int) (Puppy, error) {
    var p Puppy
    if p, ok := store.store[id]; !ok {
        return Puppy{}, NewError(IDNotFound)
    }
    return p, nil
}
```

07_errors/patrickmarabeas/types.go
```diff
@@ -1,21 +1,19 @@
...
+	Read(ID int) (Puppy, error)
+	Update(ID int, puppy Puppy) (bool, error)
```
camscale, line 15: Since the return value of `Update()` was an error of sorts before, does it makes sense to keep the bool here? Just return `nil` or an error perhaps.

07_errors/patrickmarabeas/types.go
```diff
@@ -1,21 +1,19 @@
...
+	Destroy(ID int) (bool, error)
```
camscale, line 16: Same here

07_errors/patrickmarabeas/mapStore.go
```diff
@@ -9,43 +9,50 @@ func NewMapStore() Storer {
 }
 
 // Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
-func (store *MapStore) Create(puppy Puppy) int {
+func (store *MapStore) Create(puppy Puppy) (int, error) {
+	if puppy.Value < 0 {
+		return -1, NewError(NegativeValue)
```
camscale, line 14: This should be `return 0, NewError(NegativeValue)`

When an error is returned, the value of the other return parameters should be the zero value for their type. Otherwise people may start checking the wrong value for errors, or try to use those values thinking there might be some significance to the non-zero-value.

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
 
 func (suite *StoreSuite) TestCreate() {
 	a := assert.New(suite.T())
-	id := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: "450"})
 
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: 450})
```
camscale, line 18: This should be called `err` not `error` as `error` is the built-in error type.

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
 	a.Equal(id, 0)
+	a.Equal(error, nil)
```
camscale, line 20: Use `a.NoError(err)` rather than comparing to `nil`

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
+	id, error := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: 300})
 	a.Equal(id, 1)
+	a.Equal(error, nil)
+}
+
+func (suite *StoreSuite) TestCreateNegativeNumber() {
+	a := assert.New(suite.T())
+
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: -100})
+	a.Equal(id, -1)
```
camscale, line 35: `id` should be irrelevant for the error case - no point testing it.

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
+	a.Equal(error, NewError(NegativeValue))
```
camscale, line 36: `a.Error(...)`

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
 }
 
 func (suite *StoreSuite) TestRead() {
 	a := assert.New(suite.T())
-	data := suite.store.Read(0)
 
-	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: "450"})
+	data, error := suite.store.Read(0)
+	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: 450})
+	a.Equal(error, nil)
```
camscale, line 44: `a.NoError(err)`

Going to stop saying this now. Please fix throughout.

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
+	a.Equal(id, -1)
+	a.Equal(error, NewError(NegativeValue))
 }
 
 func (suite *StoreSuite) TestRead() {
 	a := assert.New(suite.T())
-	data := suite.store.Read(0)
 
-	a.Equal(data, Puppy{ID: 0, Breed: "Wolf", Color: "Grey", Value: "450"})
+	data, error := suite.store.Read(0)
```
camscale, line 42: s/error/err/

Going to stop saying this now. Please fix throughout

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
 
+	id, error := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: 300})
 	a.Equal(id, 1)
+	a.Equal(error, nil)
+}
+
+func (suite *StoreSuite) TestCreateNegativeNumber() {
+	a := assert.New(suite.T())
+
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: -100})
```
camscale, line 34: s/error/err/

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
+	id, error := suite.store.Create(Puppy{Breed: "Wolf", Color: "Grey", Value: 450})
 	a.Equal(id, 0)
+	a.Equal(error, nil)
 }
 
 func (suite *StoreSuite) TestCreateSecond() {
 	a := assert.New(suite.T())
-	id := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: "300"})
 
+	id, error := suite.store.Create(Puppy{Breed: "Boxer", Color: "Brown", Value: 300})
```
camscale, line 26: s/error/err/

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
 	a.Equal(success, false)
+	a.Equal(error, NewError(IDNotFound))
 }
 
 func (suite *StoreSuite) TestIdIncrementOnDelete() {
 	a := assert.New(suite.T())
-	id := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
-	suite.store.Destroy(id)
-	newID := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: "700"})
+	id, _ := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: 700})
```
camscale, line 105: Rather than ignoring the error here, perhaps use `require`?
```
import "github.com/stretchr/testify/require"
...
require.NoError(suite.T(), err)
```

`require` will stop the test there, and I find it useful to use it to differentiate between test setup that is failing and the results of the tests themselves (for which I use assert).

07_errors/patrickmarabeas/store_test.go
```diff
@@ -14,70 +14,99 @@ type StoreSuite struct {
...
+	success, _ := suite.store.Destroy(id)
+	a.Equal(success, true)
 
+	newID, _ := suite.store.Create(Puppy{Breed: "Greyhound", Color: "Light Brown", Value: 700})
```
camscale, line 109: Similar. You do not expect this `Create()` call to fail, so `require` that it does not.

07_errors/patrickmarabeas/syncStore.go
```diff
@@ -8,43 +8,50 @@ func NewSyncStore() Storer {
 }
 
 // Create increments the uuid and adds the provided Puppy struct to the store with this identifier.
-func (store *SyncStore) Create(puppy Puppy) int {
+func (store *SyncStore) Create(puppy Puppy) (int, error) {
+	if puppy.Value < 0 {
+		return -1, NewError(NegativeValue)
```
camscale, line 13: similar to mapstore - use a zero value for unspecified return values.

07_errors/patrickmarabeas/syncStore.go
```diff
@@ -13,19 +13,23 @@ func (store *SyncStore) Create(puppy Puppy) (int, error) {
 		return -1, NewError(NegativeValue)
 	}
 
+	store.Lock()
 	puppy.ID = store.uuid
 	store.Store(puppy.ID, puppy)
 	store.uuid++
+	store.Unlock()
```
camscale, line 20: The usual Go pattern is to put this right after the `Lock()` with `defer`:
```
store.Lock()
defer store.Unlock()
```

This way if you later add an early return in the middle for error handling, you do not forget to unlock.

07_errors/patrickmarabeas/syncStore.go
```diff
@@ -13,19 +13,23 @@ func (store *SyncStore) Create(puppy Puppy) (int, error) {
...
 }
 
 // Read returns the puppy matching the provided uuid.
 // An empty Puppy struct is returned if the identifier does not exist.
 func (store *SyncStore) Read(id int) (Puppy, error) {
+	store.RLock()
 	if value, ok := store.Load(id); ok {
 		return value.(Puppy), nil
 	}
+	store.RUnlock()
```
camscale, line 32: This is a bug, because you are not unlocking the store inside the `if` statement. Instead use the `defer` pattern above and not worry about it any more.

07_errors/patrickmarabeas/syncStore.go
```diff
@@ -52,6 +57,9 @@ func (store *SyncStore) Destroy(id int) (bool, error) {
 		return false, NewError(IDNotFound)
 	}
 
+	store.Lock()
 	store.Delete(id)
+	store.Unlock()
```
camscale, line 62: ditto. use the `defer` pattern, even for a simple one line block like this one.

camscale (changes requested):
Generally pretty good. And I really like the split up of commits, but there's one issue there with making all commits buildable and working.

`main()` is a bit odd - maybe it makes sense to use the storers you implement somehow, but i note there is no requirement to in the README, so i guess that's ok!

Have a read of your commit messages - i noticed a spelling mistake in one but github does not allow me to comment at the commit level, and now I've forgotten exactly which one it was.

juliaogris:

PR #586, opened 2019-08-15
Lab-4

NicholeL:
Implement functions
Implement tests
Fixes #586 

Review of colleague's PR 

Changes proposed in this PR:
Implement lab 4 of go-course
Implement testing.

PR #600, opened 2019-08-22
Add lab3 - letter count

### LINKED ISSUE #599: Lab3 - Implement letter frequency
`func letters(s string) map[rune]int`
- Write a function that returns a sorted slice of strings with elements "{key}:{val}". 
`func sortLetters(m map[rune]int) []string`

harirakr:
Fixes #599 

Review of colleague's PR #586 

- Implement function to return the letter frequency
- Implement function to return a sorted list from the letter frequency
- Add tests


camscale (approved):
Excellent work. I have no comments except to approve this for merging.

Just one small suggestion to think about for future labs - try to make every test case test something that the other test cases do not. For instance in `TestSortLetters()` you have two tests that have the same four element values but with different keys. I don't see that the second one of these tests anything different to the first, so personally I would drop that second test. Every line of code is a line of code to be maintained, so the less the better in general.

One way to keep this idea in mind is to use table-driven tests (https://github.com/golang/go/wiki/TableDrivenTests) and give a name to every test. Trying to come up with a good name can often tell you when there is nothing unique about your test.

Anyway, just something to consider. Great work!

harirakr:
Thanks @camh-anz . Will follow your suggestion for the future labs.

PR #602, opened 2019-08-22
Add lab4 - numeronyms

### LINKED ISSUE #601: Lab4 - implement numeronym
- Write a function that returns a slice of numeronyms for its input strings:
`func numeronyms(vals ...string) []string`

harirakr:
Fixes #601 

Review of colleague's PR #586 

- Implement numeronym
- Add tests


camscale (approved):
LGTM. Another excellent PR. This may be the shortest, most concise version of this lab that I've seen, and as a result, it is very clear and simple (shorter is not necessarily always better, but if you can describe an algorithm clearly and concisely, that is always better).

I would make the same comment I made on lab3 - perhaps a few too many tests, in particular the testing of the `numeronyms()` function - you want to be testing what the function does, not necessarily the results of what that function calls - for example, `numeronyms()` only maps a slice of strings into another slice of strings - you just need to test that it does that with a reasonable set of different slices and that it actually calls `numeronym()` on them. So an empty slice, and a slice with at least one element of length > 3 should be sufficient as a test case for `numernyms()`.

But, that's rather minor. Well done on the nice code.

harirakr:
Thanks @camh-anz . I get what you mean. Will follow this.

harirakr:
Thanks @camh-anz . I get what you mean. Will follow this.

PR #604, opened 2019-08-23
Add lab5 - stringer interface

### LINKED ISSUE #603: Lab 5 - stringer interface

harirakr:
Fixes #603 

Review of colleague's PR #586 

- Implement stringer interface
- Add tests


05_stringer/harirakr/main_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	main()
+	assert.Equal(t, "127.0.0.1\n", buf.String())
+
```
rokane: please remove extra new line here

05_stringer/harirakr/main_test.go
```diff
@@ -0,0 +1,28 @@
...
+}
+
+func TestStringer(t *testing.T) {
+	assert.Equal(t, "0.0.0.0", fmt.Sprintf("%v", IPAddr{}))
```
rokane: You should look at structuring these as a table of tests, since they all follow the same structure (take an input, get a response and assert for equality.

rokane (dismissed):
Looking good. I have left a couple of real little comments, if you could please take a look at those then this should be fine 👍

05_stringer/harirakr/main_test.go
harirakr: Don.

05_stringer/harirakr/main_test.go
harirakr: Done.

harirakr:
Thanks @rokane . I made the suggested changes. Please take a look.

juliaogris (approved):
Perfect :)

PR #610, opened 2019-08-26
Lab1 fib first submission

BaiJieCN:
Fixes #609 

Review of colleague's PR #196 

#### Changes proposed in this PR:

- Implement Lab1 Fibonacci numbers
- Test Lab1

PR #617, opened 2019-08-27
Lab9-Puppy Store with json tags and cmd line tool

### LINKED ISSUE #616: Lab9 - JSON puppy
# Lab 9 - JSON puppy

- Create directory `09_json/USERNAME` containing a copy of upstream master `08_project/USERNAME`
- Add JSON tags to puppy data type
- Test marshalling and unmarshalling using [require.JSONEq](https://godoc.org/github.com/stretchr/testify/require#JSONEq)
- Add command line flag `-d FILE` with long form `--data FILE` using [kingpin.v2](https://godoc.org/gopkg.in/alecthomas/kingpin.v2)
- FILE should contain an array of puppies in JSON format. Parse this file and store its contents.

mohitnag:
Fixes #616 

Review of colleague's PR #610 

# Changes proposed in this PR:
- Adds Json tags to puppy struct
- Adds command line tool using kingpin parser
- Adds json marshalling and unmarshalling
- Adds tests


09_json/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,58 @@
...
+			return err
+		}
+		if err := s.CreatePuppy(puppy); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func readJSON(filename string) string {
```
camscale: This should probably be called `readFile` or something (`readFileOrPanic` might be a bit long, but that's what it does), as it does not do anything with JSON. It just reads bytes from a file. Alternatively, move the `json.Unmarshal` call into this function and call it `readPuppies` and hide all the json stuff inside it.

If you stick to just returning file data, it should return `[]byte` since you are converting `[]byte` to `string` and then converting back to `[]byte` when you use the result.

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func TestMainError(t *testing.T) {
+	assert := assert.New(t)
+	args = []string{"-d", "./../../pkg/puppy/store/testdata/invalid-puppies.json"}
+	assert.Panics(main)
+}
+
+func TestReadJson(t *testing.T) {
+	t.Run("Valid Json File should pass", func(t *testing.T) {
```
camscale: There's no reason to use `t.Run` here - just split this into two separate test functions: `TestReadJson` and `TestReadJsonBadPath`. Using `t.Run` just adds boilerplate for no benefit.

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	assert := assert.New(t)
+	puppy := puppy.Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "1"}
+	mapStore := store.MapStore{}
+	syncStore := store.SyncStore{}
+	_ = mapStore.CreatePuppy(puppy)
+	_ = syncStore.CreatePuppy(puppy)
+	err := initialisePuppyStore(&mapStore, &syncStore, "./../../pkg/puppy/store/testdata/invalid-puppies.json")
+	assert.Error(err)
+	newmapStore := store.MapStore{}
+	err = initialisePuppyStore(&newmapStore, &syncStore, "./../../pkg/puppy/store/testdata/invalid-puppies.json")
```
camscale: You have repeated the string `"./../../pkg/puppy/store/testdata/invalid-puppies.json"` a few times in this test. Perhaps put it in a const (along with the good file path too) at global scope. That will make it easier to change if the paths need to change.

09_json/mohitnag/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,148 @@
...
+		ExpectedError string
+	}{
+		{"Read already existing Puppy", 1, s.puppy, "<nil>"},
+		{"Reading a non-existing Puppy should fail", 32, puppy.Puppy{},
+			"puppy with Id 32 does not exists :(code: puppy not found)"},
+	}
+	for _, tc := range testCases {
+		tc := tc
+		s.T().Run(tc.Scenario, func(t *testing.T) {
+			fmt.Println(s.puppy)
```
camscale: some stray debugging left behind here. please remove.

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	t.Run("Non existing file should fail", func(t *testing.T) {
+		assert := assert.New(t)
+		assert.Panics(func() { readJSON("bad path") })
+	})
+}
+
+func TestUnmarshallingError(t *testing.T) {
+	assert := assert.New(t)
+	var buf bytes.Buffer
+	out = &buf
```
camscale: This test is not using `out` at all. This makes it an equivalent test to `TestMainError`. Either this test should be testing something else or it should be removed.

09_json/mohitnag/pkg/puppy/types.go
```diff
@@ -0,0 +1,12 @@
...
+// Puppy defines the data structure corresponding to a Puppy
+type Puppy struct {
+	ID     uint32 `json:"puppyid"`
+	Breed  string `json:"breed"`
+	Colour string `json:"colour"`
+	Value  string `json:"value"`
+}
+
+// Puppies is an array of Puppy
+type Puppies []Puppy
```
camscale: This type is used only in `main.go` and only in one place. This does not warrant creating a type for this. Just use `[]Puppy{}` in `main.go`, which is clearer there what is going on.

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+	args = []string{"--data", "./../../pkg/puppy/store/testdata/invalid-puppy.json"}
+	assert.Panics(main)
+}
+
+func TestInitialisePuppyStoreErrors(t *testing.T) {
```
camscale: This test is quite unclear with exactly what constitutes an error and how it is triggering those errors, because it is doing too much. I would favour breaking this up into multiple tests. For example I don't see why this test creates another MapStore and does `initialisePuppyStore` again. The first test fails i think because the store already contains a puppy with ID 1, and then you load a file with two more puppies with ID 1. Rather than being that complex, start with a known state and load a file with 1 puppy with a duplicate ID. The have another test that tests the same failure with a different store failing.

09_json/mohitnag/pkg/puppy/store/testdata/invalid-puppies.json
```diff
@@ -0,0 +1,11 @@
+[
+    {
+        "puppyid": 1,
+        "breed": "dog",
+        "colour": "white",
+        "value": "2"
+    },
+    {
+        "puppyid": 1
```
camscale: I think this file should be called `duplicate-puppies.json` rather than invalid, because the second puppy has a value  of zero (the zero value) so it is not invalid - invalid is only < 0.

camscale (changes requested):
Looking pretty good, Mohit. The code is all rather correct, but there are a few places I think that could be simplified.

09_json/mohitnag/cmd/puppy-server/main.go
mohitnag: Done. Change readFile which returns bytes

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Done, split the functions in two

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Done. Added the constants

09_json/mohitnag/pkg/puppy/store/store_test.go
mohitnag: Done. thanks for pointing out

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Removed the `out`. However this test is testing the unmarshalling error

09_json/mohitnag/pkg/puppy/types.go
mohitnag: Done

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Tried to make the test more clear. It tests the duplicate error while initialising puppystore

09_json/mohitnag/pkg/puppy/store/testdata/invalid-puppies.json
mohitnag: Renamed

mohitnag:
Thanks for your review Cam.  I tried to fix all of your review comments . Please have a look again.

nicko-lee (dismissed):

09_json/mohitnag/README.md
```diff
@@ -0,0 +1,28 @@
...
+# Installation
+- Installations of Go v1.12+ according to [official instructions](https://golang.org/doc/install), including [GOPATH env variables](https://golang.org/doc/code.html#GOPATH).
+
+# Build, install, test the PuppyStore with:
+
+```bash
+go build ./...
+go install ./...
+go test ./...
+```
```
juliaogris, line 15: blank line before heading - worth while running code formatter `prettier` over markdown.

09_json/mohitnag/cmd/puppy-server/main.go
juliaogris: Still `func readJSON(filename string) string ` - please fix.

juliaogris (dismissed):
Doesn't look like you have addressed Cam's comments?
Please do.

09_json/mohitnag/README.md
mohitnag: Done, ran `prettier` formatter

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+	defaultPuppy     = "./../../pkg/puppy/store/testdata/default-puppy.json"
+	invalidPuppyJSON = "./../../pkg/puppy/store/testdata/invalid-puppy.json"
+	duplicatePuppies = "./../../pkg/puppy/store/testdata/duplicate-puppies.json"
+)
+
+func TestMain(t *testing.T) {
+	assert := assert.New(t)
+	var buf bytes.Buffer
+	out = &buf
+	args = []string{"--data", "./../../pkg/puppy/store/testdata/many-puppies.json"}
```
camscale: nit: this path belongs with the other as a `const` further up. Keeping them together would make it easier if in future the source code were to be restructured.

09_json/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,58 @@
...
+
+var (
+	out  io.Writer = os.Stdout
+	app            = kingpin.New("puppyStore", "Puppy Store")
+	args           = os.Args[1:]
+)
+
+func main() {
+	mapStore := store.MapStore{}
+	syncStore := store.SyncStore{}
```
camscale: nit: I would move the creation of the stores after the parsing of the command line args. It makes very little difference in this case, but if there was a different sort of store, such as a SQL store, 1) you may need the command line flags to initialise the store (such as a database connection string) and 2) if there is a failure to parse the command line args, there is no point creating the store.

As I said, it's very minor and I'm not asking you to make this change - I'm just giving you something to think about and explain how I would do it.

09_json/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+func TestReadFile(t *testing.T) {
+	assert := assert.New(t)
+	expect := puppy.Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "2"}
+	actual := puppy.Puppy{}
+	buff := readFile(defaultPuppy)
+	err := json.Unmarshal(buff, &actual)
+	assert.NoError(err)
+	assert.Equal(expect, actual)
+
```
camscale: nit: remove unnecessary blank line

09_json/mohitnag/pkg/puppy/store/testdata/duplicate-puppies.json
```diff
@@ -0,0 +1,11 @@
...
+    {
+        "puppyid": 1,
+        "breed": "dog",
+        "colour": "white",
+        "value": "2"
+    },
+    {
+        "puppyid": 1
+    }
+]
```
camscale: nit: there is no end-of-line at the end of this file. All lines should be terminated by a newline, including the last line. See if you can configure your editor to always do that.

camscale (dismissed):
LGTM. This is really good, Mohit. There are a couple of minor comments but I'm not going to block you merging. If you do want to fix them up, ping me on slack and i'll rereview ASAP.

09_json/mohitnag/pkg/puppy/store/testdata/duplicate-puppies.json
mohitnag: Done

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: DOne

09_json/mohitnag/cmd/puppy-server/main.go
mohitnag: it makes sense. I moved  the creation of the stores after parsing CLI args

09_json/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Done

camscale (approved):
LGTM. Great work, Mohit.

PR #624, opened 2019-09-01
Add lab 2, bubblesort

### LINKED ISSUE #623: Lab 2 - Bubblesort
Lab 2: 

Create an executable go program in directory 02_bubble/USERNAME
Write a function that returns a sorted copy of int slice s using bubble sort:
func bubble(s []int) []int
Call fmt.Println(bubble([]int{3, 2, 1, 5})) in main to print:
[1 2 3 5]

alextmz:
Fixes #623

Review of colleague's PR #631

- Implement lab 2, bubblesort


alanhillgemann:
@alextmz you could add test for no params and negative ints

alextmz:
> @alextmz you could add test for no params and negative ints

Fixed - thanks!

02_bubble/alextmz/main.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+
```
juliaogris, line 18: Unnecessary blank line.

02_bubble/alextmz/main.go
```diff
@@ -0,0 +1,36 @@
...
+	for len(s) > 0 {
+		for k, v := range s {
```
juliaogris: This is not bubble sort as I know it. Without early bail it should look something like this:
```go
	for i := len(s); i > 0; i-- {
		for j := 1; j < i; j++ {
			if s[j-1] > s[j] {
				s[j-1], s[j] = s[j], s[j-1]
			}
		}
        }
```
What does len(s) > 0 mean as the loop condition with s not being modified? Shouldn't it just be a forever? `for { ...`

Also, you are meant to sort a copy and not modify the input (see [lab 2 assignment](https://github.com/anz-bank/go-course/blob/master/02_bubble/README.md)).

juliaogris (changes requested):
Please send link to review of colleague's PR. You have write permissions.

02_bubble/alextmz/main.go
alextmz: Yep, I've tried to do a bubblesort from head and kinda focused more on the tests, so it was awfully wrong. Tests were very wrong too btw! :P All fixed now. Thanks for the review!

02_bubble/alextmz/main.go
```diff
@@ -0,0 +1,32 @@
...
+	fmt.Fprintln(out, bubble([]int{3, 2, 1, 5}))
+}
+
+func bubble(s []int) []int {
+	r := make([]int, len(s))
+	copy(r, s)
+
+	for {
+		sorted := true
+		for i := 1; i < len(r); i++ {
```
juliaogris: Still not bubble sort as I understand it, sorry ;)
My understanding of bubble sort is: 
you compare adjacent elements and if left bigger than right you swap. Doing so will have the largest element in the right most position of the slice (at position `len(r)-1`) when the first outer iteration is completed. You therefore only need to sort to `len(r)-2` in the next outer iteration and so forth.... The loop indices in bubble sort according to my understanding should run:
``` go
for i := len(s); i > 0; i-- {
		for j := 1; j < i; j++ {
```
If I read your code correctly you resort to the end on every outer iteration even though you already know that the rightmost `n` elements are sorted, with `n` being the count of the outer loop.

juliaogris (changes requested):
One more point around what bubble sort actually does, the rest looks great.

juliaogris (changes requested):

02_bubble/alextmz/main_test.go
```diff
@@ -0,0 +1,66 @@
...
+		result := buf.String()
+
+		if result != want {
+			t.Errorf("expected %#v, got %#v", want, result)
+		}
+	})
+}
+
+func Test_bubble(t *testing.T) {
+
```
juliaogris, line 25: unnecessary blank line.

juliaogris (approved):
Just a couple of minor nits - for next time. feel free to merge:
- unnecessary blank lines (newer versions of golangci-lint complains about this now).
- usually in unit tests its expected|want vs actual|got and not the other way round (in your if statements).

The rest LGTM

PR #514, opened 2019-06-29
Lab 1: Fibonacci

krnicolson:
Fixes #539 

Review of colleague's PR #379 

#### Changes proposed in this PR: 
* fib(n) function to print fibonacci series up to n
* call to fib() from main with supplied input
* test main and fib

PR #625, opened 2019-09-03
Lab1 - Generate Fibonacci Series

### LINKED ISSUE #630: Lab1 - Fib

jimbosoft:
Fixes #630 

Review of colleague's PR #514 

Submit lab1 Fibonacci from the go course
## -


01_fib/jimbotech/.gitignore
```diff
@@ -0,0 +1,2 @@
+/.vscode/*
```
willshen8: Only source files should be committed.

01_fib/jimbotech/README.md
```diff
@@ -0,0 +1,15 @@
...
+As part of the ANZ "go" course the first lab is to produce a piece of software that will produce the Fibonacci series for fib(7). It was also an exercise in the tour.golang.org.
+
+Generating the series was implemented as "closure" which was a requirement in the tour-golang . There are other ways to solve this problem, but I kept it as an example of closure for later reference.
+
+The function maintains state and hence has to be called sequentially.
+
+
+
+
+> Written with [StackEdit](https://stackedit.io/).
```
willshen8: Great to see you putting together a ReadME file, I don't think anyone else has done it. Perhaps remove the part that is auto-generated by the tools. e.g. 
`> Written with [StackEdit](https://stackedit.io/).`

01_fib/jimbotech/ex-fib_test.go
```diff
@@ -0,0 +1,46 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestFib(t *testing.T) {
```
willshen8: It would be worthwhile to test the boundary cases such as `fib(0)` and `fib(1)` and even better larger fib number.

01_fib/jimbotech/ex-fibonacci.go
```diff
@@ -0,0 +1,47 @@
...
+var out io.Writer = os.Stdout
+
+// fibonacci returns a function that returns
+// an number in the fibonacci sequence.
+// This is an implmentation of "closure".
+// The inside function returns the current number
+// in the fibonnacci sequence, starting at 1
+// and sets up the next which will be returned
+// in the subsequent call
+func fibonacci() func() int {
```
willshen8: The tab takes up 8 character spaces here. I would think normally a tab/indentation would be 2 or 4 spaces.

01_fib/jimbotech/ex-fibonacci.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+	"io"
+	"math"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+// fibonacci returns a function that returns
```
willshen8: I can't comment on closure and it is definitely an interesting way of doing it. I personally feel if you have to write this much comment to explain what the function does, then it needs to be simplified. From a learning perspective, it might be easier to learn how it is done as using go structure and syntax.

01_fib/jimbotech/ex-fibonacci.go
```diff
@@ -0,0 +1,47 @@
...
+		secondLast, last = last, secondLast+last
+		return result
+	}
+}
+
+func fib(n int) {
+	f := fibonacci()
+
+	if n < 0 {
+		for i := 0; i > n; i-- {
```
willshen8: There is more idiomatic way of writing [for loops ](https://golang.org/doc/effective_go.html) and you have used already on line 14 of the test file.

willshen8 (changes requested):
Apart from the comments below, I think the filename needs to be snakecase to align with convention:
https://stackoverflow.com/questions/35694137/file-name-convention-for-compound-words

willshen8:
Also please open an 'issue' and put the issue number inside `Fixes #` of the PR description. The go-course require commit messages to be squashed, so you might have to do an interactive rebase.

01_fib/jimbotech/.gitignore
jimbosoft: I usually check in the .gitignore and others also consider it normal practise as per https://stackoverflow.com/questions/5765645/should-you-commit-gitignore-into-the-git-repos, however in this context it might not add a lot, so I removed it.

01_fib/jimbotech/README.md
jimbosoft: StackEdit is an open source community and I feel I should acknowledge their efforts by crediting them in my work, so if it is ok, I think leaving in the credit is the right thing to do.

01_fib/jimbotech/ex-fib_test.go
jimbosoft: Done

01_fib/jimbotech/ex-fibonacci.go
jimbosoft: I agree, I much prefer 4 characters, but after having spent a lot of time beating my editor into submission, I found "go fmt" simply puts the 8 space tab right back. I also checked other peoples submissions and they all use 8 space tabs. As much as I am agreeing with you, I have been defeated and will not be able to implement your suggestion.

01_fib/jimbotech/ex-fibonacci.go
jimbosoft: "Closure" is a key concept in go as per the tour of go: https://tour.golang.org/moretypes/25 I added all the comments so in six months time I can still remember why I did it. Hope it is ok if I keep it just like this.

01_fib/jimbotech/ex-fibonacci.go
jimbosoft: In line 14 of my test I was "ranging" over a collection. This is not the case here. I am simply counting from 0 to n. I can't think of any other way to do it, unless you could show me how?

01_fib/jimbotech/README.md
willshen8: Fair enough.

01_fib/jimbotech/ex-fibonacci.go
willshen8: I think you are right, the editor shows up as normal indentation, and gofmt puts it back as 8.

willshen8:
I moved the issue number up in the PR description (into the first line), I think maybe I wasn't clear. Hope you see it now.

willshen8 (dismissed):

01_fib/jimbotech/ex_fibonacci.go
```diff
@@ -0,0 +1,47 @@
...
+		return result
+	}
+}
+
+func fib(n int) {
+	f := fibonacci()
+
+	if n < 0 {
+		for i := 0; i > n; i-- {
+			factor := int(math.Pow(-1, float64(i*-1)))
```
camscale: The computation of negative fibonacci numbers really belongs where the sequence is calculated. This function is responsible for printing `n` fib numbers. Instead you've split the computation over two separate functions for negative numbers. For instance, if you wanted to have another function return a slice of fib numbers instead of printing them, you would have to duplicate this negative number handling. But you've kind of backed yourself into a corner with your closure implementation of fib - it has no notion of which direction to go in so it cannot compute negative numbers.

Also the computation of -1 or 1 here is a bit obtuse. `i*-1` is more simply expressed as `-i` and using floating point calculations just to alternate between -1 and 1 seems unnecessary. If it were possible to write it as `-1^-i` in simple mathematical notation, then it may be worth doing. But with the casts and `math.Pow` instead of `^` or similar, it becomes that much harder to mentally parse and understand. An alternative is to use a variable `sign` initialised to `1`, and each time around the loop you just do `sign *= -1` - that way it alternates but is a much simpler expression.

01_fib/jimbotech/ex_fibonacci.go
```diff
@@ -0,0 +1,47 @@
...
+			fmt.Fprintln(out, f()*factor)
+		}
+
```
camscale: nitpick: stray blank line. unnecessarily spaces out the code, which matters on real code that does not fit on one screen. blank lines to separate "paragraphs" of code are good, but do think about what your "code paragraphs" are.

01_fib/jimbotech/ex_fib_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestFib(t *testing.T) {
+
+	fibResults := []int{1, 1, 2, 3, 5, 8, 13, 21, 34}
+
+	f := fibonacci()
+
+	for _, val := range fibResults {
+		if r := f(); r != val {
+			t.Errorf("Returned value %v does not match %v", r, val)
```
camscale: error messages in Go should start with lower case letters. This is because error messages are often wrapped as the propagate causing a SeQuEnCe of lower/upper case letters: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

01_fib/jimbotech/README.md
```diff
@@ -0,0 +1,15 @@
+# Fibonacci Number Generator
+## Overview
+As part of the ANZ "go" course the first lab is to produce a piece of software that will produce the Fibonacci series for fib(7). It was also an exercise in the tour.golang.org.
```
camscale: nitpick: Markdown is an extension of plain old ascii documents, in that ascii markup is added so it can be prettier when rendered as markdown. But the files should still be readable as ascii documents. This means wrapping text around 80 chars (the general convention), as people use all sorts of different width terminals, and some that require side scrolling to read wider text. So please make markdown text readable as plain ascii as originally intended.

Plenty has been written about line length / text width - here's one example: https://baymard.com/blog/line-length-readability

camscale (changes requested):
The file name `ex_fib_test.go` should be called `ex_fibonacci_test.go` - i.e. the same as the name as the file being tested but with `_test` appended before the extension. This is not something that is explicitly documented but is a convention I've never seen not followed.

My general approach to these reviews is to classify my comments into two types: Go-specific ones that need to be addressed to merge, and general software engineering practice ones which are advisory.

In this review, just the comments about error strings are Go-specific. So please at least fix those up (all instances - I did not leave a comment on every one, so look at all your error strings) before submitting for re-review. It is up to you as to whether you want to address the other comments, but it is good practice to at least respond to all comments with an acknowledgement or reason for ignoring.

Finally, a couple of comments on your commits and commit messages. Have a read of https://chris.beams.io/posts/git-commit/ (also linked from the slides) on what makes a good commit message. In particular, your first commit titled "INITIAL COMMIT" does not have a good title. It is shouty and non-descriptive. The initial split into five commits is fine and well done, but it does mean you need to take more action with your subsequent fixup commits - the should be squashed with the commits where the original code being fixed was written. We do not want to see fixup commits merged into master - the place to get your code and commits into order is the pull request. When merging it should look as though a series of commits was written "perfectly" the first time. If you started with one commit and intended to only keep one commit, you could do a "squash merge" when merging the PR to remove all evidence of the fix-ups. But if you want to retain your original split, you'll need to manually apply the fix-ups with `git rebase`. Remember, history is written by the victors. Be a victor!

01_fib/jimbotech/ex_fibonacci.go
jimbosoft: Done ... however I feel hardly done by. I feel I am handling the negative numbers in the right place as the series is being generated there. So now I have separated the generation from the printing. Am I going to be criticised for inefficiency now? I was hoping for bonus points not bonus corrections. Secondly, as this is a exerciser I felt proud of my play with the math library and expected applause. Yes I did not like floating point either. When I re-wrote it I needed ABS, guess what, floating point again. So can I use any of the math library?

01_fib/jimbotech/ex_fibonacci.go
jimbosoft: Done

01_fib/jimbotech/ex_fib_test.go
jimbosoft: Done

01_fib/jimbotech/README.md
jimbosoft: Done

camscale (changes requested):
There is something very wrong with this PR. It has 29 files changed in it with code from other people's lab. I suggest you checkout the latest master and rebase on that, then force push. That *should* clean it up, but I am not sure how you got here so cannot offer precise guidance of how to get out.

I cannot re-review until this is fixed.

camscale (approved):
LGTM.

BTW, if you wanted to incorporate negative fibonacci numbers into your closure way of generating numbers:
```
func fibonacci(direction int) func() int {
	if direction < 0 {
		direction = -1
	} else {
		direction = 1
	}
	secondLast, last := 0, 1

	return func() int {
		result := last
		secondLast, last = last, secondLast+(last*direction)
		return result
	}
}
```

Then you could just call `f := fibonacci(n)` and iterate over the result of `f()` which would remove the need for the intermediate slice. Or slightly more verbose but taking care of the iteration properly:
```
f := fibonacci(n)
limit := n
if n < 0 {
     limit = -n
}
for i := 0; i < limit; i++ {
    fmt.Fprintln(out, f())
}
```

I'm not requesting this change, just pointing out what I had in mind with my previous comments.

PR #631, opened 2019-09-04
Initial commit for lab2 the bubble sort

### LINKED ISSUE #632: Lab 2 - Bubble sort

jimbosoft:
Fixes #632 

Review of colleague's PR #610 

- Implement Lab2 Bubble sort
- Test Lab2


alextmz (commented):
Nice one doing a readme. It's a habit I should get into myself.

02_bubble/jimbotech/bubble_test.go
```diff
@@ -0,0 +1,61 @@
...
+
+func TestNotModifiedInput(t *testing.T) {
+	input := []int{64, 11, 1, 1, 2, 3, 5, 8, 13, 21, 34}
+	original := make([]int, len(input))
+	copy(original, input)
+
+	bubble(input)
+
+	if !reflect.DeepEqual(original, input) {
+		t.Errorf("The input slice was altered %v does not match %v", original, input)
```
alextmz: Don't capitalize error strings (from https://github.com/anz-bank/go-course/wiki/Lab-1:-Collected-Code-Review-Comments, see also https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Also valid for all other error strings in the tests.

02_bubble/jimbotech/bubble_test.go
```diff
@@ -0,0 +1,61 @@
+package main
+
+import (
+	"bytes"
+	"reflect"
+	"testing"
+)
+
+func TestNotModifiedInput(t *testing.T) {
+	input := []int{64, 11, 1, 1, 2, 3, 5, 8, 13, 21, 34}
```
alextmz, line 33: Nice one testing for non-modification of original string.

02_bubble/jimbotech/bubble_test.go
```diff
@@ -0,0 +1,61 @@
...
+	expected := []int{1, 1, 2, 3, 5, 8, 11, 13, 21, 34, 64}
+
+	sortedResult := bubble(input)
+
+	if !reflect.DeepEqual(sortedResult, expected) {
+		t.Errorf("Returned value %v does not match %v", sortedResult, expected)
+	}
+}
+func TestNoInput(t *testing.T) {
+	input := []int{}
```
alextmz, line 56: You could also test for nil slices, and for 1-element only slice, but it's really nitpicking.

02_bubble/jimbotech/bubble_test.go
```diff
@@ -0,0 +1,61 @@
...
+func TestNoSort(t *testing.T) {
+	input := []int{1, 1, 2, 3, 5, 8, 13, 21, 34}
+	expected := []int{1, 1, 2, 3, 5, 8, 13, 21, 34}
+
+	sortedResult := bubble(input)
+
+	if !reflect.DeepEqual(sortedResult, expected) {
+		t.Errorf("Returned value %v does not match %v", sortedResult, expected)
+	}
+}
```
alextmz, line 74: Consider [table driven tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) - optional.

02_bubble/jimbotech/bubble_test.go
```diff
@@ -0,0 +1,61 @@
...
+	out = &buf
+
+	main()
+
+	result := buf.String()
+
+	if result != expected {
+		t.Errorf("expected %v, got %v", expected, result)
+	}
+}
```
alextmz, line 88: I don't mind (and in fact agree) with the extra line spacing, but I have a hunch that reviewers are going to find it a bit much here.

alextmz (dismissed):
I think it's overall nice and simple, but error messages should be improved and optionally tests should be made table driven.

02_bubble/jimbotech/bubble.go
```diff
@@ -0,0 +1,29 @@
...
+
+var out io.Writer = os.Stdout
+
+func bubble(s []int) []int {
+	sorted := make([]int, len(s))
+	copy(sorted, s)
+
+	var ss []int
+	for k := range sorted {
+		ss = sorted[:len(sorted)-k-1]
```
camscale: Since `ss` is used only inside the loop, there's no point declaring the variable outside the loop. You can also use the `:=` syntax:
```
ss := sorted[:len(sorted)-k-1]
```

However, since you only use `ss` in one place too, you may as well just put it there without the variable:
```
for i, val := range sorted[:len(sorted)-k-1]
```

I was looking up and down the function to verify if `ss` was used anywhere else. Putting it in the `for...range` loop makes it clear it is not.

02_bubble/jimbotech/bubble.go
```diff
@@ -0,0 +1,29 @@
...
+		for i, val := range ss {
+			if val > sorted[i+1] {
```
camscale: Using a subslice to range over in the inner loop is interesting, but it does make this a little less clear since the relationship between `val` and `sorted[i]` is obscured. The inner `if` written as:
```
if sorted[i] > sorted[i+1] {
    sorted[i], sorted[i+1] = sorted[i+1], sorted[i]
}
```
is very clear what it is doing under what condition.

This is a lab exercise so it is ok and fun to explore other ways to do things, but one of the goals of Go is to keep it simple, so do try not to be too clever unless there is some other upside.

02_bubble/jimbotech/bubble_test.go
camscale: The nil slice and the empty slice are effectively the same thing in Go. Underlying memory allocation may differ but the behaviour is the same.

However, a test for a 1-element slice is definitely a good idea since the code being tested does an `i+1` in the loop, you want to make sure you have your boundary conditions correct.

camscale (changes requested):
LGTM mostly, but you're really taking the mickey with those issues you create. You could at least cut and paste the lab description like most others do, but you could go further and put the text you put in the readme into the issue comments too. Of course, creating good documentation around your code (via issues, commit messages,etc) is outside the scope of Go, so I'm not going to be too nit picky about it, but if you ever come and work with me in future, I might not be so relaxed :)

As Alex said, consider table-driven tests. Don't worry about re-working these commits, but it would be good to see for lab 4 as that is a rather data-driven set of tests that would benefit from table-driven tests.

Also as Alex has said, Go error strings should start with lower-case as I've pointed out on the other labs. Just fix them up and ping me and I'll re-review to approve for merge asap.

02_bubble/jimbotech/bubble_test.go
jimbosoft: Done

02_bubble/jimbotech/bubble_test.go
jimbosoft: Done

02_bubble/jimbotech/bubble_test.go
jimbosoft: Good idea, I got bored with my own tests, will do it for lab4.

02_bubble/jimbotech/bubble.go
jimbosoft: Done

02_bubble/jimbotech/bubble.go
jimbosoft: Done

camscale (approved):
LGTM.

Please use squash-merge when you merge to clean up the commit history (2 commits called "initial commit"?). It would be nice to see good commit/PR/issue hygiene too - the title of the PR is "initial commit for lab2 the bubble sort" which could be nicer (a PR is not a commit), the linked issue has no content, etc - just keep that in mind for future PRs please. You can merge this one now.

PR #634, opened 2019-09-06
Implement lab 3 - letters

jimbosoft:
Fixes #633

Review of colleague's PR #534

#### Changes proposed in this PR:
- Implement Lab3 letters
- Test Lab3

PR #636, opened 2019-09-08
Lab 10

### LINKED ISSUE #585: Lab 10 - Puppy REST

runnerdave:
Fixes #585

Review of colleague's PR #634

 - Implement Puppy server with chi router


10_rest/runnerdave/README.md
```diff
@@ -0,0 +1,92 @@
...
+Storage of Puppies with two different storage methods.
+
+## Pre-requisites
+
+You need to have [Go](https://golang.org/doc/install) installed
+
+## Build and install
+
+```
+go install ./...
```
juliaogris: nit/personal preference, feel free to ignore: I prefer to just indent my code blocks in Markdown unless I specify the language with it, so in you case to render the same result that would be:

```markdown
## Build and install

    go install ./...
```

10_rest/runnerdave/README.md
```diff
@@ -0,0 +1,92 @@
...
+    "value": 173.98
+  }'
+```
+### Delete puppies
+ Endpoint: {host:port}/api/puppy/{id}
+ Method: DELETE
+ Example:
+```
+curl -X DELETE http://localhost:3000/api/puppy/2
+```
```
juliaogris: great README!

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+func main() {
+	_, err := kingpin.CommandLine.Parse(args)
+	if err != nil {
+		fmt.Fprintf(out, "Command line could not be parsed, error: %v", err)
+		return
+	}
+
+	d, _ := ioutil.ReadFile(*data)
+	puppies, uerr := unmarshalPuppies(d)
```
juliaogris: nit: idiomatic go always uses err, unless you are for some reason tracking two errors at the same time (which you aren't here).

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,112 @@
...
+	if uerr != nil {
+		return
```
juliaogris: log / print error e.g. `log.Fatalf("cannot unmarshal puppies %v", err)` or `panic(err)`

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+		fmt.Fprintf(out, "Map store of puppies:%v", mapStore)
+		s = &mapStore
+	case "sync":
+		syncStore, err := storeInSync(puppies)
+		if err != nil {
+			fmt.Fprintf(out, "Could not setup sync storage:%v", err)
+		}
+		fmt.Fprintf(out, "Sync store of puppies:%v", syncStore)
+		s = syncStore
```
juliaogris: IMO this could be factored clearer.

I'd have something like:
```go
func load(s puppy.Storer, puppies []puppy.Puppy) err {
	for _, p := range puppies {
		err := s.CreatePuppy(p)
		if err != nil {
			return err
		}
	}
}
```
and 
```go
func NewStore(string db) (puppy.Storer,  error) {
	switch db {
	case "map":
		return store.NewMapStore(), nil
	case "sync":
		return store.NewSyncStore(), nil
	}
	return nil, fmt.Errorf("unknown store type")
}
```

and probably break it up inside main as:

```go
store, err := NewStore(string db)
// if err....
err = load(store, puppies)
// if err....
```

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,112 @@
...
+	default:
+		fmt.Fprintf(out, "Invalid storage")
+	}
+	return s
+}
+
+func unmarshalPuppies(d []byte) ([]puppy.Puppy, error) {
+	puppies := []puppy.Puppy{}
+	if err := json.Unmarshal(d, &puppies); err != nil {
+		fmt.Fprintf(out, "Could not unmarshall puppies, error: %v", err)
```
juliaogris: I'd get rid of all the debug messages and just print proper error handling in main.
Alternatively you can wrap your error:
`return nil, fmt.Errorf("cannot unmarshal puppies, error: %v", err)`

same below.

10_rest/runnerdave/puppydata/repeated-puppy.json
```diff
@@ -0,0 +1,14 @@
...
+    "color": "green",
+    "value": 2732.81
+  },
+  {
+    "id": 2,
+    "breed": "ULTRASURE",
+    "color": "green",
+    "value": 2732.81
+  }
+]
```
juliaogris: Add EOL at EOF
Same above.

juliaogris (changes requested):
Nit: fixup commits should be done as fixups `git ci --fixup <HASH>` squash merged and force pushed.
A few more nitpicks inside your code.
Looks pretty solid though. Great work!

10_rest/runnerdave/puppydata/repeated-puppy.json
runnerdave: added EOL to all data files

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: done, no debug message in functions, only handling of errors in main

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: implemented, it really makes it much simpler this way, thanks a lot

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: done

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: good point

10_rest/runnerdave/README.md
runnerdave: thanks

10_rest/runnerdave/README.md
runnerdave: much cleaner this way, thanks

nicko-lee (dismissed):
Looks good. Nice json file test coverage!

10_rest/runnerdave/README.md
```diff
@@ -0,0 +1,92 @@
+# Puppy server
+
+Storage of Puppies with two different storage methods.
+
+## Pre-requisites
+
+You need to have [Go](https://golang.org/doc/install) installed
+
+## Build and install
+
```
juliaogris, line 10: nit: one blank line is enough in markdown.
I usually run prettier over all my markdown files for uniform formatting. (similar to gofmt).

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,94 @@
...
+	store "github.com/anz-bank/go-course/10_rest/runnerdave/pkg/puppy/store"
+
+	"github.com/go-chi/chi"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var (
+	args           = os.Args[1:]
+	data           = kingpin.Flag("data", "data file").Short('d').Default("puppydata/data.json").ExistingFile()
+	port           = kingpin.Flag("port", "PORT").Short('p').Default("8080").String()
```
juliaogris: nit: int? (should have noticed this earlier, apologies)

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,94 @@
...
+	db             = kingpin.Flag("store", "STORE").Short('s').Default("sync").String()
```
juliaogris: nit: Enum? (should have noticed this earlier, apologies)

10_rest/runnerdave/cmd/puppy_server/main.go
juliaogris: nit: log.Fatal exits the program (with exit(1) that is), so you don't need the extra return line.

10_rest/runnerdave/cmd/puppy_server/main_test.go
```diff
@@ -0,0 +1,164 @@
...
+	{"breed":"Cacri","color":"Undefined","id":2,"value":1.3} `)
+	expected := strings.ReplaceAll(readable, "\\n\\t", "")
+	actual := strconv.Quote(s.String())
+	t.Logf("expected:%s", expected)
+	t.Logf("actual:%s", actual)
+
+	if expected != actual {
+		t.Errorf("Unexpected output in main()")
+	}
+
```
juliaogris: Unnecessary blank line.
below as well.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	chi "github.com/go-chi/chi"
+)
+
+// RestStorer store for puppies
+type RestStorer struct {
```
juliaogris: RestStorer does not seem like the right term.
1. `er` suffix usually indicates interface not implementation
2. Rest is to do with http request response not with storing. it might use a storer to persist/retrieve data, but its not its core purpose. Hope I make sense.

I usually call this type of struct `RestServer` or similar and make it implement the http.Handler interface so I can run it as `http.ListenAndServe(port, restServer)` - not 100% sure if that's great either....

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+const (
+	getPuppyRoute    = "/api/puppy/{id}"
+	postPuppyRoute   = "/api/puppy/"
+	putPuppyRoute    = "/api/puppy/{id}"
+	deletePuppyRoute = "/api/puppy/{id}"
+)
+
+// GetPuppyRoute Route for the GET method
+func (rs *RestStorer) GetPuppyRoute() string {
+	return getPuppyRoute
```
juliaogris: why consts and methods? const alone should suffice?

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	puppy "github.com/anz-bank/go-course/10_rest/runnerdave/pkg/puppy"
+	store "github.com/anz-bank/go-course/10_rest/runnerdave/pkg/puppy/store"
+)
+
+var (
+	rs puppy.RestStorer
+)
+
+func startAPIHandler() *chi.Mux {
```
juliaogris: this doesn't really start anything, it's more like it wires it together.
I usually call this: setupRoutes or wireRoutes.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+	return putPuppyRoute
+}
+
+// DeletePuppyRoute Route for the PUT method
+func (rs *RestStorer) DeletePuppyRoute() string {
+	return deletePuppyRoute
+}
+
+// GetPuppy HandlerFunction
+func (rs *RestStorer) GetPuppy(w http.ResponseWriter, r *http.Request) {
```
juliaogris: All these methods should be private imo.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,227 @@
...
+		Value:  2732.81,
+	}
+	if err := rs.Db.CreatePuppy(newPuppy); err != nil {
+		panic(err)
+	}
+
+	r.Get(rs.GetPuppyRoute(), rs.GetPuppy)
+	r.Post(rs.PostPuppyRoute(), rs.CreatePuppy)
+	r.Put(rs.PutPuppyRoute(), rs.UpdatePuppy)
+	r.Delete(rs.DeletePuppyRoute(), rs.DeletePuppy)
```
juliaogris: I think it'd rewrite this as something like:
I also think it doesn't belong with the test code but the rest.go file.
```go
	r.Route("/api/puppy", func(r chi.Router) {
		r.Post("/", rs.CreatePuppy)
	        r.Route("/{id}", func(r chi.Router) {
		     r.Use(CtxIDMiddleware) // see chi docs for sample, e.g. REST Preview: https://github.com/go-chi/chi#examples
		     r.Put("/", h.handlePut)
		     r.Delete("/", h.handleDelete)
		     r.Get("/history", h.handleGetHistory)
	    })
	})
```

juliaogris (dismissed):
a few comments. please enjoy your holidays Dave, I'll review until it's done, doesn't matter if takes a little longer!

10_rest/runnerdave/README.md
runnerdave: great tip, I installed prettier and applied it to the README, thanks a lot

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: good point, it was my clumsy way of exporting them, fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

10_rest/runnerdave/cmd/puppy_server/main_test.go
runnerdave: fixed

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: fixed

10_rest/runnerdave/cmd/puppy_server/main.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: very useful comment, thanks a lot, implemented

01_fib/jimbotech/ex_fibonacci.go
```diff
@@ -0,0 +1,54 @@
+package main
```
golangcibot: /usr/local/go/src/fmt/scan.go:1200:16: DecodeRuneInString not declared by package utf8 (from `typecheck`)

07_errors/mohitnag/error.go
```diff
@@ -0,0 +1,28 @@
+package main
```
golangcibot: /go/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:174:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

08_project/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,19 @@
+package main
```
golangcibot: /usr/local/go/src/fmt/scan.go:1200:16: DecodeRuneInString not declared by package utf8 (from `typecheck`)

08_project/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,37 @@
+package puppy
```
golangcibot: /usr/local/go/src/strconv/quote.go:405:12: ValidString not declared by package utf8 (from `typecheck`)

08_project/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
+package store
```
golangcibot: /usr/local/go/src/fmt/scan.go:1200:16: DecodeRuneInString not declared by package utf8 (from `typecheck`)

09_json/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,51 @@
+package main
```
golangcibot: /usr/local/go/src/strconv/quote.go:405:12: ValidString not declared by package utf8 (from `typecheck`)

09_json/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,43 @@
+package puppy
```
golangcibot: /go/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:174:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

09_json/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
+package store
```
golangcibot: /usr/local/go/src/regexp/regexp.go:967:22: DecodeRuneInString not declared by package utf8 (from `typecheck`)

09_json/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,58 @@
+package main
```
golangcibot: /usr/local/go/src/strconv/quote.go:405:12: ValidString not declared by package utf8 (from `typecheck`)

09_json/mohitnag/pkg/puppy/error.go
```diff
@@ -0,0 +1,28 @@
+package puppy
```
golangcibot: /go/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:174:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,96 @@
+package main
```
golangcibot: /usr/local/go/src/encoding/json/fold.go:83:20: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,49 @@
+package puppy
```
golangcibot: /usr/local/go/src/strings/strings.go:980:20: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,81 @@
+package store
```
golangcibot: /usr/local/go/src/bytes/reader.go:98:18: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/nickolee/pkg/puppy/errors.go
```diff
@@ -0,0 +1,20 @@
+package puppy
```
golangcibot: /usr/local/go/src/encoding/json/fold.go:83:20: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,146 @@
+package rest
```
golangcibot: /usr/local/go/src/bytes/reader.go:98:18: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/runnerdave/cmd/puppy_server/main.go
```diff
@@ -0,0 +1,82 @@
+package main
```
golangcibot, line 1: /usr/local/go/src/bytes/reader.go:98:18: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
+package puppy_test
```
golangcibot, line 1: /usr/local/go/src/path/filepath/match.go:215:15: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/runnerdave/pkg/puppy/errors.go
```diff
@@ -0,0 +1,36 @@
+package puppy
```
golangcibot, line 1: /usr/local/go/src/encoding/json/fold.go:83:20: DecodeRune not declared by package utf8 (from `typecheck`)

10_rest/runnerdave/pkg/puppy/store/map_store.go
```diff
@@ -0,0 +1,59 @@
+package store
```
golangcibot, line 1: /go/pkg/mod/github.com/stretchr/testify@v1.4.0/assert/assertions.go:174:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

runnerdave:
Hi @juliaogris I have implemented all the requested changes but golangci is throwing some odd errors on some package names, please help me reviewing the work so far and hopefully the golangci problem will be resolved separately

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+			r.Get("/", rs.getPuppy)
+		})
+	})
+
+	return r
+}
+
+func (rs *RestServer) getPuppy(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err == nil {
```
camscale: This test should be inverted (`if err != nil`) and you should be returning a 400 Bad Request http status code. If someone calls this REST endpoint with an invalid ID (not a number), then the function will fall off the bottom and by default will return a 200 OK response, with no body.

Inverting the test also keeps the happy path on the left edge of the function: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		p, rerr := rs.Db.ReadPuppy(int16(id))
+		if rerr != nil {
+			w.WriteHeader(http.StatusNotFound)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		bytes, _ := json.Marshal(p)
+		i, _ := w.Write(bytes)
+		w.Header().Set("Content-Length", string(i))
```
camscale: This cannot be done here, and is unnecessary: https://golang.org/pkg/net/http/#ResponseWriter

Once you have called `w.Write()`, it is too late to write more headers. The headers are sent before the body.

Also since this is a small message, the `Content-Length` header is added automatically.

Have a ready of the comments on the `ResponseWriter` interface that I linked to. It describes all this.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+
+	return r
+}
+
+func (rs *RestServer) getPuppy(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err == nil {
+		p, rerr := rs.Db.ReadPuppy(int16(id))
+		if rerr != nil {
+			w.WriteHeader(http.StatusNotFound)
```
camscale: You need a `return` after this, otherwise you fall through to sending a nil puppy

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		i, _ := w.Write(bytes)
+		w.Header().Set("Content-Length", string(i))
+	}
+}
+
+func (rs *RestServer) createPuppy(w http.ResponseWriter, r *http.Request) {
+	post := Puppy{}
+	derr := json.NewDecoder(r.Body).Decode(&post)
+	if derr != nil {
+		w.WriteHeader(http.StatusBadRequest)
```
camscale, line 52: This needs a `return` after it.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		w.Header().Set("Content-Type", "application/json")
+		bytes, _ := json.Marshal(p)
+		i, _ := w.Write(bytes)
+		w.Header().Set("Content-Length", string(i))
+	}
+}
+
+func (rs *RestServer) createPuppy(w http.ResponseWriter, r *http.Request) {
+	post := Puppy{}
+	derr := json.NewDecoder(r.Body).Decode(&post)
```
camscale: `derr` should just be called `err`. It doesn't matter if you re-use `err` in a function. Always call your error variables `err` unless you need to have multiple at once.

camscale: You should combine this with the `if` statement:
```
if err := json.NewDecoder(r.Body).Decode(&post); err != nil {
    w.WriteHeader(http.StatusBadRequest)
    return
}
```

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+	if derr != nil {
+		w.WriteHeader(http.StatusBadRequest)
+	}
+	cerr := rs.Db.CreatePuppy(post)
+	if cerr == nil {
```
camscale: This `if` condition is backwards (should be `!= nil`), should be called `err` and not `cerr` and should be combined on one line:
```
if err := rs.Db.CreatePuppy(post); err != nil {
    w.WriteHeader(http.StatusInternalServerError)
    return
}
```
Technically you don't need that `return` in there, but it is best to as that is the standard way you'll do error handling in a http handler.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	chi "github.com/go-chi/chi"
+)
+
+// RestServer store for puppies
+type RestServer struct {
+	Db Storer
```
camscale: The field `Db` should either be `DB` or `db` depending on whether you want to export it or not, as it is an initialism: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		w.Header().Set("Content-Type", "application/json")
+		bytes, _ := json.Marshal(p)
+		i, _ := w.Write(bytes)
+		w.Header().Set("Content-Length", string(i))
+	}
+}
+
+func (rs *RestServer) createPuppy(w http.ResponseWriter, r *http.Request) {
+	post := Puppy{}
```
camscale: nit: post is not a good name for this variable. You use `p` in `updatePuppy` - that's probably fine.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+	cerr := rs.Db.CreatePuppy(post)
+	if cerr == nil {
+		w.WriteHeader(http.StatusInternalServerError)
+	}
+}
+
+func (rs *RestServer) updatePuppy(w http.ResponseWriter, r *http.Request) {
+	p := Puppy{}
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err == nil {
```
camscale: Same problem with a bogus `id` parameter as `createPuppy` - you need to return a 400.

camscale: Same issue regarding happy path and left edge.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		derr := json.NewDecoder(r.Body).Decode(&p)
```
camscale: Same issue with non-`err` name for an error. Also fold into the if statement.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		if derr != nil {
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		uerr := rs.Db.UpdatePuppy(int16(id), &p)
```
camscale: ditto

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,84 @@
...
+		if uerr == nil {
+			w.WriteHeader(http.StatusInternalServerError)
+		}
+	}
+}
+
+func (rs *RestServer) deletePuppy(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err == nil {
```
camscale: Same comment about happy path. In this case, I note you're handling the 400 error. if done according to the happy path on the left edge, you'd check for an error and handle that. Then go left - i.e. don't use an `else` - use a `return` in the error path.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
...
+	store "github.com/anz-bank/go-course/10_rest/runnerdave/pkg/puppy/store"
+)
+
+var (
+	rs puppy.RestServer
+)
+
+func TestGetPuppy(t *testing.T) {
+	req, err := http.NewRequest("GET", "/api/puppy/1", nil)
+	if err != nil {
```
camscale: Since you have imported `stretchr/testify/require`, I suggest you use that to verify test setup:
```
require.NoError(err)
```
That will end the test there if there is an error, and it takes 1 line instead of three.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	rr := httptest.NewRecorder()
+	newPuppy := puppy.Puppy{
+		Breed:  "ULTRASURE",
+		Colour: "green",
+		Value:  2732.81,
+	}
+	db := store.NewSyncStore()
+	if err := db.CreatePuppy(newPuppy); err != nil {
+		panic(err)
```
camscale: Similar, instead of `panic`, use `require.NoError`. Using `require` for test setup and `assert` for test verification makes it easy to see what is actually being tested, vs what is just needed to ensure the test runs in the right state.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	expected := `{
+		"id": 1,
+		"breed": "ULTRASURE",
+		"color": "green",
+		"value": 2732.81
+	  }`
+	actual := rr.Body.String()
+	require.JSONEq(t, expected, actual, "GET Puppy was incorrect, got: %s, want: %s.",
```
camscale: For reasons mentioned above, use `assert` to check that the result is correct, not `require`.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	db := store.NewSyncStore()
+	if err := db.CreatePuppy(newPuppy); err != nil {
+		panic(err)
+	}
+	handler := rs.SetupRoutes(db)
+
+	handler.ServeHTTP(rr, req)
+
+	if status := rr.Code; status != http.StatusOK {
```
camscale: You should just use `assert.Equal(t, http.StatusOK, rr.Code)` here - it is a simple assertion of equality and it will output an appropriate error message itself. No need for the `status` variable either.

10_rest/runnerdave/pkg/puppy/rest_test.go
camscale: I would leave out the error message here. All the information you put in this message will be written out automatically - the name of the test that failed, and the actual vs expected values. It makes it much shorter then. I'd also move the previous assertion down to here too, so they are together, and get rid of the `actual` variable:
```
assert.Equal(t, http.StatusOK, rr.Code)
assert.JSONEq(t, expected, rr.Body.String())
```

By getting rid of those intermediate variables, these assertions are very clear what is being tested without needing to look at any other lines of code. Once someone becomes familiar with the `httptest.ResponseRecorder` type, this reads very naturally.

Please structure your other tests the same way.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func TestPostPuppy(t *testing.T) {
+	newPuppy := puppy.Puppy{
+		ID:     1,
+		Breed:  "EPLODE",
+		Colour: "brown",
+		Value:  3889.92,
+	}
+	body, _ := json.Marshal(newPuppy)
```
camscale: Don't ignore errors, especially in tests - it leads to false confidence that all is OK. Since this is setting up the test, you should use `require.NoError(err)`. Same in the other places you are ignoring errors. Usually you would never expect these errors to fire, but every now and then, something changes you did not expect, and it is very useful to have them then.

camscale (changes requested):
I've just focussed on the REST parts of this PR. The other parts are similar or the same as the previous labs, so even if I might do things differently, we'll assume it's all reviewed ok.

Most of my comments are in three categories:
1) Idiomatic error naming and handling - just use `err` as the error variable and reuse it throughout the functions; use `if err := xxx; err != nil` instead of two lines of `err := XXX` and `if err != nil`, and test for errors, not the absence of errors - i.e. `if err != nil` vs `if err == nil`.
2) Handle all REST errors and make sure you return an appropriate http status code
3) Use `require` and `assert` consistently in tests, and don't mix the more basic `t.Error` and similar functions with the `stretchr/testify` functions.

I'm happy to continue to review this past the final course deadline - you're so close, we'll get you across the line.

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: done, also great read thanks

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: I was not happy with this either but golint requires me to use the variable "i" so I am logging it now, what advice could you give regarding golint and the return values of Write(bytes)? any help would be greatly appreciated

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

runnerdave: thanks, good point, fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: made it DB to be able to export it

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: good catch, fixed

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: fixed across the file

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: fixed

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,90 @@
...
+	p := Puppy{}
+	if err := json.NewDecoder(r.Body).Decode(&p); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if err := rs.DB.CreatePuppy(p); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
```
camscale, line 60: According to REST, this should probably return a 201 Created status here and not the default 200 OK.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	if err = json.NewDecoder(r.Body).Decode(&p); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if err = rs.DB.UpdatePuppy(int16(id), &p); err == nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
```
camscale, line 78: I think a 204 No Content is the right response here for an update that returns nothing (at least by my reading of https://developer.mozilla.org/en-US/docs/Web/HTTP/Status). I think a 200 OK means the resource should be returned in the body.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,90 @@
...
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if err = rs.DB.DeletePuppy(int16(id)); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+}
```
camscale, line 91: Likewise, I think 204 No Content is the correct response here.

10_rest/runnerdave/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,235 @@
...
+	require.NoError(t, err)
+
+	rr := httptest.NewRecorder()
+	handler := rs.SetupRoutes(store.NewSyncStore())
+	handler.ServeHTTP(rr, req)
+
+	assert.Equal(t, http.StatusInternalServerError, rr.Code)
+}
+
+func TestDeleteInvalidIdPuppy(t *testing.T) {
```
camscale: nit: Id should be ID

10_rest/runnerdave/pkg/puppy/rest.go
camscale: It looks like golint is happy with:
```
_, _ = w.Write(bytes)
```
(note, changes `:=` to `=` and get's rid of the `log` import)

camscale (dismissed):
This is plenty good enough now. I had a couple of small comments mostly regarding REST status codes. It is interesting (to me) that in my last review I did not point out these status codes. With the restructured code with the happy path on the left edge, it really stood out that they were missing. After a while, you really get used to reading Go code in this style and you find yourself a lot faster at reading it too. It's one of those Go things (amongst a lot of others) that seems a bit nit-picky but after you get used to it, it really seems to make a difference.

Anyway, feel free to merge this now if you like, or if you want one more go-around, by all means do so, and I'll keep an eye on the review queue for your update.

Well done on making it to the end of the course labs. The code you've written for this last lab is very good looking Go code. There's a certain simplicity to it that can be hard to get to. But I think when you do, it's worth it.

10_rest/runnerdave/pkg/puppy/rest.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (rs *RestServer) deletePuppy(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		return
+	}
+	if err = rs.DB.DeletePuppy(int16(id)); err != nil {
+		w.WriteHeader(http.StatusNoContent)
```
camscale: I think this is backwards now isn't it? if `DeletePuppy` returns an error, you want to return an error. If no error, you want No Content.

It is viable to handle the case where deleting a non-existent puppy just returns "sure whatever, let's call that done" because the puppy isn't there at the end, but in that case you should probably differentiate between different errors - for example, if you had a SQL DB store implemented, one error returned here may be a "database unreachable" error. In that case, you should return 500 Internal Error.

All that said, the logic here is still a little off - if there is no error, you're still not returning No Content.

camscale (commented):
oh so close, but I think you just added a bug.

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: Hi Cam thanks a lot for your patience, I did a last minute fix without thinking what I wanted to do, my test was actually wrong, but thanks to the returning of proper status codes I understood what was wrong, it was a real privilege to receive your reviews, hope is ok this time

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: done

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: done

10_rest/runnerdave/pkg/puppy/rest.go
runnerdave: also done

10_rest/runnerdave/pkg/puppy/rest_test.go
runnerdave: understood

camscale (approved):
LGTM. Congratulations on completing the 10 labs!

PR #638, opened 2019-09-09
Add lab 3 - Letter Frequency

### LINKED ISSUE #637: Lab 3 - Letter frequency
# Lab 3 - Letter frequency

```
```

```
```

Call `fmt.Println(strings.Join(sortLetters(letters("aba")), "\n"))` in `main` to print:

```
a:2
b:1
```

alextmz:
Fixes #637

Review of colleague's PR #459

 - Implement Lab 3 and it's tests

### DIFF
--- a/03_letters/alextmz/main.go
+++ b/03_letters/alextmz/main.go
@@ -1,0 +1,35 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	fmt.Fprintln(out, strings.Join(sortLetters(letters("aba")), "\n"))
+}
+
+func letters(s string) map[rune]int {
+	m := make(map[rune]int)
+	for _, v := range s {
+		m[v]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	s := make([]string, len(m))
+	i := 0
+	for k, v := range m {
+		s[i] = string(k) + ":" + strconv.Itoa(v)
+		i++
+	}
+	sort.Slice(s, func(i, j int) bool { return s[i] < s[j] })
+	return s
+}
--- a/03_letters/alextmz/main_test.go
+++ b/03_letters/alextmz/main_test.go
@@ -1,0 +1,82 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_letters(t *testing.T) {
+
+	var tests = map[string]struct {
+		arg  string
+		want map[rune]int
+	}{
+		"empty string": {
+			arg:  "",
+			want: map[rune]int{}},
+		"1 element only string": {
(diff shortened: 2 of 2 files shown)


03_letters/alextmz/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+		test := tt
+		t.Run(name, func(t *testing.T) {
+			got := sortLetters(test.arg)
+			assert.Equal(t, got, test.want)
+		})
+	}
+}
+
+func Test_main(t *testing.T) {
+
```
n0npax: Hey Alex, some code reviewers are pointing that blank lines in places like this are not needed. Please remove them.

03_letters/alextmz/main.go
```diff
@@ -0,0 +1,35 @@
...
+		m[v]++
+	}
+	return m
+}
+
+func sortLetters(m map[rune]int) []string {
+	s := make([]string, len(m))
+	i := 0
+	for k, v := range m {
+		s[i] = string(k) + ":" + strconv.Itoa(v)
```
n0npax, line 30: No action needed:
this is ok, but `fmt.Sprintf("%c:%d",k,v)` seems to be cleaner.

alextmz: Yep, it is. Bad habit because Sprintf is in my mind a 'slow and heavy' function, but I wonder if that holds true for Go too.

03_letters/alextmz/main_test.go
alextmz: Indeed, I think I pointed it out myself in a review elsewhere... good to have more eyes looking at the same code.

03_letters/alextmz/main.go
n0npax: yes it's still true :cry: `Sprintf` seems to be 2.5  times slower

```
func fSprintf(i int, c rune) string {
	return fmt.Sprintf("%c:%d", c, i)
}
func fConcat(i int, c rune) string {
	return string(c) + ":" + strconv.Itoa(i)
}
```

```bash
 n0npax@narrenturm  /tmp/aa   go test -bench=. 
goos: linux
goarch: amd64
BenchmarkSprintf-8   	10000000	       134 ns/op
BenchmarkConcat-8    	30000000	        53.2 ns/op
PASS
ok  	_/tmp/aa	3.151s
```

alextmz: Man you're a Nerd with capital N (that's a compliment btw). You didn't need to go and measure it! But as you did, thanks! Now I know how to quickly bench those :)

n0npax: Thanks :) I was just curious what is the performance difference between those two.

03_letters/alextmz/main_test.go
juliaogris: Great, could you remove it then please ;) ?

juliaogris (dismissed):
Nice work, one nit left, feel free to ignore or update and ask for reapproval.

juliaogris (approved):
lgtm

PR #461, opened 2019-06-20
Feature/lab 4 numeronym

starfallz:
Fixes #460

Review of colleague's PR #

#### Changes proposed in this PR:

- Write a function that returns a slice of numeronyms for its input strings

PR #640, opened 2019-09-09
Add lab 4 - Numeronym

### LINKED ISSUE #639: Lab 4 - Numeronym
Lab 4 - Numeronym

Create an executable go program in directory 04_numeronym/USERNAME
Write a function that returns a slice of numeronyms for its input strings:

func numeronyms(vals ...string) []string

Call fmt.Println(numeronyms("accessibility", "Kubernetes", "abc")) in main to print:

[a11y K8s abc]

alextmz:
Fixes #639

Review of colleague's PR #461

- Add lab 4 - numeronym and tests


n0npax (dismissed):

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,29 @@
...
+	r := make([]string, len(vals))
+	for k, v := range vals {
+		r[k] = numeronym(v)
+	}
+	return r
+}
+
+func numeronym(s string) string {
+	if len(s) < 4 {
+		return (s)
```
camscale: unnecessary parens around `s`

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+	return s[:1] + strconv.Itoa(len(s)-2) + s[len(s)-1:]
```
camscale: Using a range (i.e. a colon in the slice) seems unnecessary when you want to extract a single element. Just use s[0] for the start and s[len(s)-1] for the end. That's also a bit clearer.

camscale: However, this wont work. See my comment on your test cases for what you've missed and have a read of https://blog.golang.org/strings for the ins and out of strings and what exactly is going on here.

04_numeronym/alextmz/main_test.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_numeronyms(t *testing.T) {
```
camscale: Don't use underscores (snake case) in Go. Use CamelCase. See: https://talks.golang.org/2014/names.slide#5

04_numeronym/alextmz/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+	var tests = map[string]struct {
+		arg  []string
+		want []string
+	}{
+		"empty string": {
```
camscale, line 41: You should write two test cases, one for `TestNumeronyms` and one for `TestNumeronym`. The `TestNumeronyms` test case should not have this much detail, because all the `numeronyms` function does is call `numeronym` on a slice of strings and returns the result as a slice. That is all you should be testing what works. In this case, I'd be testing the empty slice, and a slice with "some number" of elements. That would be enough to satisfy you that `numeronyms` is doing what it claims to do.

In `TestNumeronym`, you should be testing the function that converts one string into another. You would use a set of test cases to exercise the function well, but you wouldn't have all these slices of strings as you do in this test case. That would make it easier to read. Just pairs of inputs and expected values.

In the case that a test fails, this will make it easier to isolate the failure.

04_numeronym/alextmz/main_test.go
```diff
@@ -0,0 +1,57 @@
...
+		"multiple strings with 1 element": {
+			[]string{"a", "b", "c", "d"},
+			[]string{"a", "b", "c", "d"}},
+		"multiple long strings": {
+			[]string{"abracadabra", "alakazam", "hocuspocus", "mumbojumbo"},
+			[]string{"a9a", "a6m", "h8s", "m8o"}},
+		"mixed strings": {
+			[]string{"pirilimpimpim", "MIA", "", "Zzz", "Zoe", "Pizzaz"},
+			[]string{"p11m", "MIA", "", "Zzz", "Zoe", "P4z"}},
+	}
```
camscale, line 47: You are missing tests with non-ascii unicode runes in your strings. At the beginning, middle and end. If you add these, you'll find your numeronym function fails with strings that start or end with unicode runes represented as more than one byte in utf-8 encoding (the encoding that Go uses for strings).

camscale (changes requested):
Sorry I didn't get to this earlier, Alex. Please make sure you mark your PRs as "ready for review" if they still need a review. I do sometimes remember to look at [-label:"ready for review" label:"approved by colleague"] which is when I saw this one.

A couple of things to fix up here, but otherwise a good compact solution. You have made the classic mistake with this lab, which is good because you'll remember from now how exactly strings work WRT runes and bytes.

camscale:
Oh, also please fill out the issue you've linked to this PR. Have a read of https://github.com/anz-bank/go-course/wiki/Software-Engineering-with-Go for why we want this extra stuff.

04_numeronym/alextmz/main.go
alextmz: done

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,34 @@
...
+		r[k] = numeronym(v)
+	}
+	return r
+}
+
+func numeronym(s string) string {
+	if len(s) < 4 {
+		return s
+	}
+	firstrune, _ := utf8.DecodeRuneInString(s)
```
golangcibot, line 30: DecodeRuneInString not declared by package utf8 (from `typecheck`)

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,34 @@
...
+	lastrune, _ := utf8.DecodeLastRuneInString(s)
```
golangcibot, line 31: DecodeLastRuneInString not declared by package utf8 (from `typecheck`)

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,34 @@
...
+	numrunes := utf8.RuneCount([]byte(s))
```
golangcibot: RuneCount not declared by package utf8 (from `typecheck`)

alextmz:
Changes requested implemented. Golang-CI is acting weird.

juliaogris (dismissed):
lgtm

04_numeronym/alextmz/main.go
alextmz: done

04_numeronym/alextmz/main_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNumeronym(t *testing.T) {
+	var tests = map[string]struct {
+		arg  string
+		want string
+	}{
+		"empty string":               {"", ""},
+		"single string of 1 element": {"a", "a"},
```
camscale, line 16: Your test cases are a little lacking - you really should be testing boundary conditions, so you want a test case of length 3 and 4. You should also have some test cases at the boundary with non-ascii characters.

I suggest you add a test case for "💩", "💩💩", "💩💩💩" and "💩💩💩💩" perhaps (or any non-ascii rune - for some childish reason, pile-of-poo is my go-to non-ascii rune). If you add those, you may see some interesting results: https://goplay.space/#aaUgdA0erhN

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,34 @@
...
+func numeronyms(vals ...string) []string {
+	r := make([]string, len(vals))
+	for k, v := range vals {
+		r[k] = numeronym(v)
+	}
+	return r
+}
+
+func numeronym(s string) string {
+	if len(s) < 4 {
```
camscale: This exhibits a similar problem mentioned in my earlier review, and is not doing what you think it's doing. See my comment on the test cases for some examples of where this will fail.

As described in https://blog.golang.org/strings, a string is effectively a read-only slice of bytes, not runes, so the length of a string is not the number of runes, but the number of bytes.

camscale (changes requested):
There is a bug in here which suggests some more test cases are needed.

04_numeronym/alextmz/main_test.go
alextmz: Tests added

04_numeronym/alextmz/main.go
alextmz: Yep, I corrected this - the `len` usage was quite naive and the source of the issue.

04_numeronym/alextmz/main.go
```diff
@@ -0,0 +1,34 @@
...
+func numeronyms(vals ...string) []string {
+	r := make([]string, len(vals))
+	for k, v := range vals {
+		r[k] = numeronym(v)
+	}
+	return r
+}
+
+func numeronym(s string) string {
+	if utf8.RuneCountInString(s) < 4 {
```
golangcibot: RuneCountInString not declared by package utf8 (from `typecheck`)

camscale (dismissed):
This looks like it works now.

I see you are using both `RuneCountInString` and `RuneCount`. This will possibly cause confusion as it is an unnecessary distinction. When I see something like that, I wonder "why?" and when there's no reason, it just makes it harder to understand.

Also, there's no need to call it twice. If you move the `numrunes := ...` to the top of the function, you can have your test be `if numrunes < 4`.

But this is general software engineering practices and not related directly to Go, so I'll approve this.

Thanks for your perseverance.

alextmz:
Added the suggested changes... yes it looked too ugly the way it was, I should had paid more attention and not only done a quick-fix. Hey it's a software engineering course in Go, not programming in Go ;-)

camscale (approved):
LGTM

PR #642, opened 2019-09-10
Lab10

### LINKED ISSUE #641: lab10 - rest puppy

nicko-lee:
Fixes #641 

Review of colleague's PR #636 

#### Implement lab10 functionality + write tests
- Create puppyHandler which is essentially a REST wrapper around our Storer interface
- Which exposes the functionality we built over the past few labs over a REST API
- Expand flag for user to specify storer type as well as other config settings like port number and file to load data from
- Learn how to use net/http/httptest package for testing REST APIs


10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	invalidInputMsg = "Invalid input, ensure ID and JSON are valid"
+	notFoundMsg     = "Not Found: Apologies, for the puppy you seek does not exist"
+)
+
+// not entirely sure why u need this but get a vague sense. Check with Dan
+// a special guy who wraps around Storer types and gives them extra abilities to now handle API requests
+// thus enhancing their storing abilities significantly!
+type PuppyHandler interface {
+	handlePost(w http.ResponseWriter, r *http.Request)
```
golangcibot: U1000: func `PuppyHandler.handlePost` is unused (from `unused`)

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,139 @@
...
+	handleGet(w http.ResponseWriter, r *http.Request)
```
golangcibot: U1000: func `PuppyHandler.handleGet` is unused (from `unused`)

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,139 @@
...
+	handlePut(w http.ResponseWriter, r *http.Request)
```
golangcibot: U1000: func `PuppyHandler.handlePut` is unused (from `unused`)

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,139 @@
...
+	handleDelete(w http.ResponseWriter, r *http.Request)
```
golangcibot: U1000: func `PuppyHandler.handleDelete` is unused (from `unused`)

10_rest/nickolee/pkg/puppy/rest/rest.go
anzboi: Your linting issue comes from the fact that `SetupRoutes` takes a `PuppyHandlerAndStorer`, not a `PuppyHandler`.

This means `PuppyHandler` really isn't used at all, and cannot be used outside this package, so is effectively dead code.

To fix, your interface methods need to be exported, `SetupRoutes` needs to take the interface.

IMO PuppyHandler doesn't even need to be an interface, and can be the struct with the store. Some people disagree with this though, you can go either way.

In the end, you would have a package that looks like this

```go
type PuppyHandler interface {
    HandlePost(args) // notice it is exported
    // other methods
}

type PuppyHandlerAndStore struct {
  Store puppy.Storer
}

// implement PuppyHandler methods for puppyHandlerAndStore

func SetupRoutes(r chi.Router, ph PuppyHandler) {
  r.Get("path", ph.HandleGet)
  // other routes
}
```

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+-   Install `go 1.12` according to [official installation instruction](https://golang.org/doc/install)
+-   Clone this project outside your `$GOPATH` to enable [Go Modules](https://github.com/golang/go/wiki/Modules)
+-   Install `golangci-lint` according to [instructions](https://github.com/golangci/golangci-lint#local-installation)
+
+## How to build this project?
+
+To build the project, run the following command while you are in the root of the project folder:
+
+`go build ./...`
```
anzboi: Usually we like to be a little more specific as to what we are building.
```bash
go build ./cmd/puppy-server
```

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+Alternatively, simply run:
+`go run ./cmd/puppy-server/main.go`
+
+Which will simply revert to pre-specified default values.
+
+Note that there are three flags which work with main.go:
+
+1. **-d,--data:** if you wish to specify a .json file to load in seed data for the database, pass it as a value to this flag. Note that if no file is specified then a default file is used.
+2. **-s, --store:** either _map_ or _sync_ Note that if not set then the value defaults to _map_
```
anzboi: Missing period
```suggestion
2. **-s, --store:** either _map_ or _sync_. Note that if not set then the value defaults to _map_
```

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+`go run ./cmd/puppy-server/main.go --data puppy-data/puppies.json --port 7777 --store map`
+
+Alternatively, simply run:
+`go run ./cmd/puppy-server/main.go`
+
+Which will simply revert to pre-specified default values.
+
+Note that there are three flags which work with main.go:
+
+1. **-d,--data:** if you wish to specify a .json file to load in seed data for the database, pass it as a value to this flag. Note that if no file is specified then a default file is used.
```
anzboi: Should not use a default file, should just not load any initial data.
Your project could provide an example a user can use to try out the functionality.

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+2. **-s, --store:** either _map_ or _sync_ Note that if not set then the value defaults to _map_
+3. **-p, --port:** if set then an **API** server is started on the user specified port. Must be a valid port number between 0 and 65535
```
anzboi: What if it is not set?

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+- Learn how to test REST APIs with the net/http/httptest package
+- Learning how to use the Go standard library as well as external third party libraries such as [chi](https://godoc.org/github.com/go-chi/chi), [kingpin](https://godoc.org/gopkg.in/alecthomas/kingpin.v2) and [testify](https://godoc.org/github.com/stretchr/testify)
+
+## Prerequisites
+
+-   Install `go 1.12` according to [official installation instruction](https://golang.org/doc/install)
+-   Clone this project outside your `$GOPATH` to enable [Go Modules](https://github.com/golang/go/wiki/Modules)
+-   Install `golangci-lint` according to [instructions](https://github.com/golangci/golangci-lint#local-installation)
+
+## How to build this project?
```
anzboi: These headings are much bigger than normal
```
## Build

## Run

## Test
```

10_rest/nickolee/README.md
```diff
@@ -0,0 +1,103 @@
...
+To test this project, follow the steps below:
+
+1. Run test cases:`go test ./...`
+2. Lint it: `golangci-lint run`
+3. Ensure 100% test coverage `go test -coverprofile=coverage.out ./... && go tool cover -html=coverage.out`
+
+## API Spec
+
+### Create puppies
+ Endpoint: {host:port}/api/puppy/
```
anzboi: A few things here...

This doesn't print as well on github as you might hope.

You never include any of `scheme://host:port` in the api spec. This just tells your machine where to find (host:port) and how to connect (scheme) to the physical server serving the api.

Moreover, normally an API spec would be documented in the form of a `swagger.yaml` file. We don't require that you learn swagger. If you wanted to keep this here, you could do it in a form like the following

```
### POST /api/puppy

Adds a new puppy to the store, returns the stored puppy with a generated id.

Example:
    curl -X POST http://localhost:7777/api/puppy/ \
      -d '{
        "breed": "Iron Dog",
        "colour": "Red",
        "value": 9500
      }'

returns:
      -d '{
        "id": 1
        "breed": "Iron Dog",
        "colour": "Red",
        "value": 9500
      }'

### GET /api/puppy/{id}

...
```

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+var out io.Writer = os.Stdout
+
+var (
+	args = os.Args[1:]
+	// this loads your file once parseFlags() is run
+	// these are methods too prepare a variable. Basically telling kingpin how to read the flag from cli
+	fileName = kingpin.Flag("data", "This flag supplies a puppy JSON file name").
```
anzboi, line 29: While the input is a filename, kingpin is returning a file struct. This variable should probably be named so that is suggest it's not merely a `string`, eg:
`datafile`

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+// Config contains arguments parsed from commandline flags
+type Config struct {
+	file   *os.File
+	port   int
+	storer string
+}
+
+func main() {
+	// parse flags/check if user even provided any
+	cliConfig, err := parseFlags(args)
```
anzboi, line 44: I wouldn't bother with a separate parseFlags function.
```
func main() {
  kingpin.CommandLine.Parse(args)
  ...
  // use the parsed flag values directly
  puppies, err := readFileAndUnmarshalPuppies(*datafile)
}
```
I don't think this will affect coverage

WARNING: DO NOT use the flag variables outside of main, they are globals and WILL destroy your tests on race conditions.

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		fmt.Println(err)
+	}
+
+	// read JSON from file + parse/unmarshal the json and we now have []puppy (Go objects)
+	puppies, err := readFileAndUnmarshalPuppies(cliConfig.file)
```
anzboi: Move this block closer to where the puppies are loaded into the store, It's a little weird doing other stuff in the middle that are completely unrelated to these two tasks.

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// get the user specified port
+	addr := ":" + strconv.Itoa(cliConfig.port)
+
+	// create some puppies loaded in from json file and store in store (no pun intended lol)
+	for _, pup := range puppies {
+		pup := pup // to avoid scopelint error: Using a reference for the variable on range scope `pup`
+		createdPuppy, _ := phs.Storage.CreatePuppy(&pup)
+		fmt.Fprintf(out, "Puppy with ID %d has been created\n", createdPuppy)
+		retrievedPuppy, _ := phs.Storage.ReadPuppy(createdPuppy)
```
anzboi: Don't bother reading the puppy back out, this just produces extra noise that isn't necessary.
You should be able to trust that no error returned from `Create` means the puppy was indeed created.

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	phs := rest.NewPuppyHandlerAndStorer(newStorer)
+
+	// get the user specified port
+	addr := ":" + strconv.Itoa(cliConfig.port)
+
+	// create some puppies loaded in from json file and store in store (no pun intended lol)
+	for _, pup := range puppies {
+		pup := pup // to avoid scopelint error: Using a reference for the variable on range scope `pup`
+		createdPuppy, _ := phs.Storage.CreatePuppy(&pup)
+		fmt.Fprintf(out, "Puppy with ID %d has been created\n", createdPuppy)
```
anzboi, line 66: I like this, effectively logging successful loading of each puppy.

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	// setting up REST API wrapper
+	r := chi.NewRouter()
+	r.Use(middleware.RequestID)
+	r.Use(middleware.Logger)
+
+	rest.SetupRoutes(r, *phs)
+
+	fmt.Printf("Starting server on port %s. Try:\n", addr)
```
anzboi: This line is good, I wouldn't bother with the other lines though.

10_rest/nickolee/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,122 @@
...
+	fmt.Printf("  curl localhost%s/api/puppy -d ", addr)
```
anzboi: You don't normally see output like this when you start a server. Consider the following.
1. What if it's not running on localhost
2. What if you didn't provide a datafile (getting puppy 4 would fail)

Also, you are posting a puppy, so it needs `curl -X POST stuff`, though I wouldn't keep it.

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+	"strconv"
+
+	"github.com/anz-bank/go-course/10_rest/nickolee/pkg/puppy"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/render"
+)
+
+const (
+	invalidInputMsg = "Invalid input, ensure ID and JSON are valid"
```
anzboi: This error message is used both when ID and JSON are invalid. Think about how you might separate these two error cases out to give the consumer a more accurate error message.

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+	notFoundMsg     = "Not Found: Apologies, for the puppy you seek does not exist"
```
anzboi: `Apologies`
 so polite 👍

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+
+	// give client feedback. Opted for this rather than render.Status(r, http.StatusNoContent) as gives more feedback
+	render.JSON(w, r, "Puppy successfully deleted")
+}
+
+// SetupRoutes maps the exposed endpoints with their respective handlers
+func SetupRoutes(r chi.Router, phs PuppyHandlerAndStorer) {
+	r.Get("/api/puppy/{id}", phs.handleGet)
```
anzboi: I would suggest experimenting with subroutes
```
r.Route("/path", func(r chi.Router) {
  r.Get("/subpath", handler)
})
```

Also, usually API's served over http have a version in the route, `api/v1/path`

10_rest/nickolee/pkg/puppy/rest/rest_test.go
```diff
@@ -0,0 +1,205 @@
+package rest
```
anzboi, line 1: This file is too similar to @willshen8. `rest.go` was inevitably going to look similar due to you both using `chi`, but testing could go many different ways.

Suggestion: split the one `TestRestAPI` into four, one for each method (eg: `TestGetPuppy`). This will also simplify the test data since each test case doesn't need a method (you'll know what it is based on the function you are in).

You can also look at using `httptest.Server` instead of `Recorder`, see how that changes your tests.

anzboi (changes requested):
Unfortunately I have to insist on some changes to `rest_test.go`, they are too similar to @willshen8 and I think you will benefit from doing them again, I've suggested how you might go about this.

Other than that, left a lot of other comments on how you might change things a little just to clean it up a bit

10_rest/nickolee/pkg/puppy/rest/rest.go
nicko-lee: haha :)

10_rest/nickolee/README.md
nicko-lee: ah i see so if my project has got more than one binary - go build ./... will build all binaries. In this case it won't make a difference but that is clear. Got it. Being clear about exactly what i am building

nicko-lee: but if do go install that would be go install ./... usually because i want all the binaries

10_rest/nickolee/README.md
nicko-lee: Got it. Normally it should default to no file. If I don't give it a file it should default to having no puppies in it.

10_rest/nickolee/README.md
nicko-lee: It defaults to 7777

nicko-lee: Will add it to the README

10_rest/nickolee/README.md
nicko-lee: Got it

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: ok will delete this perhaps move to the readme

10_rest/nickolee/README.md
nicko-lee: Have changed this

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: have changed the logic around this to actually return a *string so that I can check if no file is provided without kingpin doing the checks and panicking me out

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: Removed

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: 👍

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: no problem have removed this. Just was following this (second) example from the course slides 😂
https://go-course.org/gocourse-2019-05-melbourne.slide#104

10_rest/nickolee/pkg/puppy/rest/rest_test.go
nicko-lee: Did what you suggested. Rewrote test cases. Broke them up into four. No longer using `Recorder` and using `httptest.Server` instead which did indeed change my tests significantly as you predicted

10_rest/nickolee/pkg/puppy/rest/rest.go
nicko-lee: Good suggestions. Didn't know you could put routers in a router. That's pretty neat. Done!

10_rest/nickolee/README.md
nicko-lee: defaults to port 7777. Have added that to the readme

10_rest/nickolee/cmd/puppy-server/main.go
anzboi: Fair enough

10_rest/nickolee/pkg/puppy/rest/rest_test.go
```diff
@@ -0,0 +1,291 @@
...
+	switch n {
+	case 0:
+		return store.NewMapStore()
+	default:
+		return store.NewSyncStore()
+	}
+}
+
+// a setup function that returns a preconfigured router that our test client can use
+func createTestRouter() *chi.Mux {
```
anzboi: Pass in a `*testing.T` object and use `require.NoError(t, err)` instead of panicking on errors.

This will give you the nice test output for errors in test setup, as well as the tests itself.

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,148 @@
...
+	Storage puppy.Storer // Using the thread safe sync.map implementation of Storer interface
+	// ask Dan what the significance would be if I used storage store.SyncStore vs just store.SyncStore
+}
+
+// This guy gives u a nicely initialised brand new *PuppyHandlerAndStorer
+func NewPuppyHandlerAndStorer(s puppy.Storer) *PuppyHandlerAndStorer {
+	return &PuppyHandlerAndStorer{Storage: s}
+}
+
+// Implementing 1/4 methods in PuppyHandler interface: puppy handler for GET /api/puppy/{id}
```
anzboi: `PuppyHandler` interface no longer exists, need to update this comment

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,148 @@
...
+func SetupRoutes(r chi.Router, phs PuppyHandlerAndStorer) {
+	r.Route("/api/puppy", func(r chi.Router) {
+		r.Get("/{id}", phs.handleGet)
+		r.Post("/", phs.handlePost)
+		r.Put("/{id}", phs.handlePut)
+		r.Delete("/{id}", phs.handleDelete)
+	})
+}
+
+// func (r *Rest) Handle(w http.ResponseWriter, r *http.Request) {
```
anzboi: Removed commented code

10_rest/nickolee/pkg/puppy/rest/rest.go
```diff
@@ -0,0 +1,148 @@
+package rest
```
anzboi, line 1: One more change I would like you to consider, having a `errorHandler` method that checks if a returned error is a `puppy.Error` and set response status and message based on that.

For example, currently a get calls `Store.Read`, which may return an error. You assume that error to mean 404 Not Found, but it could be 500 (think about other possible implementation of store, ie: sql database).

in other words, have an error handler function specifically for handling errors from the storer
```
func handlerStoreError(w http.ResponseWriter, err error) {
  switch e := err.(type) {
  case *puppy.Error:
    // use puppy error to define the response status and body
  default:
    // handle otherwise (probably a 500)
  }
}
```

You only need to use this function for errors returned from the store, for other error cases you can handle them as you are right now with `http.Error`

10_rest/nickolee/cmd/puppy-server/main.go
anzboi: It's fine if you choose not to do this, not a big deal (in fact, for more complicated services, usually the first thing you do is read any input into config).

10_rest/nickolee/cmd/puppy-server/main.go
anzboi: I would still like you to move this down. Basically, the current flow is
1. read data file
2. create store
3. create handler
4. create address
5. load initial data into store
6. setup router
7. serve


You can move the first one down to just before 5, and also move 3 to before 6 or 7, so it looks more like
1. create store
2. create handler
3. read data file
4. load initial data into store
5. setup router
6. create address
7. serve

anzboi (changes requested):
Good work nicko, looking much better.

Still have a few more comments to clean it up a little more

10_rest/nickolee/cmd/puppy-server/main.go
nicko-lee: ah i get what u mean now. Thanks for clarifying that! Makes sense to me now and have moved the block to a more logical order

10_rest/nickolee/pkg/puppy/rest/rest.go
nicko-lee: Fixed

10_rest/nickolee/pkg/puppy/rest/rest.go
nicko-lee: Removed

10_rest/nickolee/pkg/puppy/rest/rest.go
nicko-lee: fixed

10_rest/nickolee/pkg/puppy/rest/rest_test.go
```diff
@@ -0,0 +1,304 @@
...
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+)
+
+func TestHandleStorerError(t *testing.T) {
+	wr := httptest.NewRecorder()
+	err := errors.New("hello")
+	handleStorerError(wr, err)
+	resp := wr.Result() // can treat as if is the response from the handler
```
golangcibot, line 25: bodyclose: response body must be closed (from `bodyclose`)

anzboi (approved):

anzboi:
LGTM!
well done on completing 10 labs

PR #644, opened 2019-09-10
Add lab 5 - Stringer

### LINKED ISSUE #643: Lab 5 - Stringer
Lab 5 - Stringer

Create an executable go program in directory 05_stringer/USERNAME
Make the IPAddr type implement fmt.Stringer to print the address as a dotted quad
Find hints at tour of go exercise: stringers
Call fmt.Println(IPAddr{127, 0, 0, 1}) in main to print:

127.0.0.1

alextmz:
Fixes #643

Review of colleague's PR #215

- Add lab 5 - stringer and tests


kasunfdo (dismissed):
Clean!

05_stringer/alextmz/main.go
```diff
@@ -0,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+// IPAddr is below, violets are blue and this is here to shut up golint :-)
```
camscale: Amusing, but you can also shut up lint by writing a proper doc string :-)

If you really don't want to, you don't need to export the type - golint will stop complaining then too.

Either way, please approach these labs as you would real code, and write the comment as appropriate.

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func Test_IPAddrString(t *testing.T) {
+	var tests = map[string]struct {
+		arg  IPAddr
+		want string
+	}{
+		"empty IPv4": {
+			IPAddr{},
+			"0.0.0.0"},
```
camscale: Either put the closing brace on a line of its own, or use a more compact layout with everything on one line:
```
"empty IPv4": {IPAddr{}, "0,0.0.0},
...
```
The reason for putting the brace on its own line if you are making a vertical list, is that it makes every line consistently the same - they all have a comma at the end. If you extend the struct to add another field and add the extra element to each struct, you end up with a diff where you just have lines added. To modify your structure to add another element, you end up with a diff where the last line gets changed to change the `},` to `,` and then another line added.

Clearer diffs make reviewing easier, and makes it easier to spot mistakes, if for instance you accidentally delete a character on the line you would not otherwise touch.

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		"24bits IPv4": {
+			IPAddr{254, 254, 254},
+			"254.254.254.0"},
+		"full IPv4": {
+			IPAddr{254, 254, 254, 254},
+			"254.254.254.254"},
+	}
+
+	for name, tt := range tests {
+		test := tt
```
camscale: rather that using both `test` and `tt`, where `tt` is kind of invented because you need another name, you can just use the one name and shadow that inside the loop:
```
for name, test := range tests {
    test := test
    ...
}
```

When an experienced Go reviewer sees that, they understand exactly what is going on. Usually you would not want to shadow variables because it can cause confusion, but this is one of those cases where it is almost always done due to a wart in the language and makes sense.

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+			actual := test.arg.String()
+			assert.Equal(t, test.want, actual)
+		})
+	}
+}
+
+func Test_main(t *testing.T) {
+	want := "127.0.0.1\n"
+
+	t.Run("main test", func(t *testing.T) {
```
camscale: There is no point using `t.Run()` here. Just put the code of the test in `TestMain`.

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+		var buf bytes.Buffer
+		out = &buf
+		main()
+		got := buf.String()
+		assert.Equal(t, got, want)
```
camscale: assertions should be of the form (expected, actual), or in your case (want, got). See the docs here: https://godoc.org/github.com/stretchr/testify/assert#Assertions.Equal

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
...
+	for name, tt := range tests {
+		test := tt
+		t.Run(name, func(t *testing.T) {
+			actual := test.arg.String()
+			assert.Equal(t, test.want, actual)
+		})
+	}
+}
+
+func Test_main(t *testing.T) {
```
camscale: This should be called `TestMain` as noted above on your previous lab: https://talks.golang.org/2014/names.slide#5

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,51 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_IPAddrString(t *testing.T) {
```
camscale: This should be called `TestIPAddrString` as noted above on your previous lab: https://talks.golang.org/2014/names.slide#5

camscale (changes requested):
A few things to fix up to make this more idiomatic Go, but otherwise the code is fine and correct.

Also, please fill out the attached issue for reasons described at https://github.com/anz-bank/go-course/wiki/Software-Engineering-with-Go

When done, add the label "ready for review" so we see it in our lists.

05_stringer/alextmz/main.go
alextmz: Woops, this was not intended to stay there, but I guess I forgot to clean it out before pushing. Done.

05_stringer/alextmz/main_test.go
alextmz: Makes sense. Done.

05_stringer/alextmz/main_test.go
alextmz: Agreed.

05_stringer/alextmz/main_test.go
alextmz: Done

05_stringer/alextmz/main_test.go
alextmz: Yep, bad bug. Done.

05_stringer/alextmz/main_test.go
alextmz: Done

05_stringer/alextmz/main_test.go
alextmz: Done

05_stringer/alextmz/main.go
```diff
@@ -0,0 +1,19 @@
+package main
```
golangcibot, line 1: /go/pkg/mod/github.com/stretchr/testify@v1.3.0/assert/assertions.go:173:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

juliaogris (changes requested):
please add link to a review of a colleagues pr in your pr description.

juliaogris (dismissed):
lgtm

05_stringer/alextmz/main.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+type iPAddr [4]byte
```
camscale: This really should be called `ipAddr` - when you have an initialism/acronym at the start of a symbol that is private, you should make it all lowercase - `ip` in this case. If it were public, it would be `IP`. See https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

05_stringer/alextmz/main_test.go
```diff
@@ -0,0 +1,53 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_iPAddrString(t *testing.T) {
```
camscale: As mentioned in my last review, you should not use underscores in symbols in Go: https://talks.golang.org/2014/names.slide#9. In this case, I would name the function `TestIPAddrString` - I know this is not an exact match for the case of the type, but `TestipAddrString` is a bit weird, so I'd just go with the former.

camscale (changes requested):
A couple of little things to fix up still - as Go as a rather opinionated naming style I think it best to get these right before merging.

05_stringer/alextmz/main.go
alextmz: Agreed, I've been paying attention to these but started on the later labs.

05_stringer/alextmz/main_test.go
alextmz: Agreed, fixed.

camscale (approved):
LGTM. Merge away.

PR #645, opened 2019-09-12
Lab 4 - Numeronym

### LINKED ISSUE #648: Lab 4 - Numeronym
Lab exerciser for the ANZ go course that involves converting words to numeronyms, which are number-based words. As this is a vague definition, I chose to follow the rules from this web site: https://www.revolvy.com/page/Numeronym.

jimbosoft:
Fixes #648 

Review of colleague's PR #624 

lab 4 - numeronym and tests

### DIFF
--- a/04_numeronym/jimbotech/numeronym.go
+++ b/04_numeronym/jimbotech/numeronym.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"unicode"
+)
+
+var out io.Writer = os.Stdout
+
+// following rules from https://www.revolvy.com/page/Numeronym
+// Program need not convert words with length 3, and should not convert shorter words.
+// If an l (lowercase ell) would occur before a 1 (one), it should be made uppercase.
+// If an I (uppercase eye) would occur before a 1 (one), it should be made lowercase.
+// Input will be printable ASCII with spaces.
+func wordMixer() func(inputStr string) string {
+
+	knownConversions := map[string]string{
+		"year 2038":        "Y2K38",
+		"year 2000":        "Y2K",
+		"i love you":       "143",
+		"eyjafjallajökull": "E15",
+	}
+
+	return func(inputStr string) string {
+
+		inputStr = strings.TrimSpace(inputStr)
+		runes := []rune(inputStr)
+
+		if len(runes) <= 3 {
+			return inputStr
+		}
+
+		lowerRunes := []rune(inputStr)
+		for i, v := range lowerRunes {
+			lowerRunes[i] = unicode.ToLower(v)
+		}
+		if value, found := knownConversions[string(lowerRunes)]; found {
+			return value
+		}
+
+		firstChar := runes[0]
+		lastChar := runes[len(runes)-1]
+		len := len(runes) - 2
+
+		if len > 9 && len < 20 {
+			if firstChar == 'I' {
+				firstChar = 'i'
+			} else if firstChar == 'l' {
+				firstChar = 'L'
+			}
+		}
+		return fmt.Sprintf("%c%d%c", firstChar, len, lastChar)
+	}
+}
+
+func numeronyms(vals ...string) []string {
+
+	f := wordMixer()
+	results := []string{}
+
+	for _, val := range vals {
+		results = append(results, f(val))
+	}
+	return results
+}
+
+func main() {
+	fmt.Fprintln(out, numeronyms("accessibility", "Kubernetes", "abc"))
+}
--- a/04_numeronym/jimbotech/numeronym.go
+++ b/04_numeronym/jimbotech/numeronym.go
@@ -5,68 +5,47 @@ import (
 	"io"
 	"os"
 	"strings"
-	"unicode"
 )
 
 var out io.Writer = os.Stdout
 
-// following rules from https://www.revolvy.com/page/Numeronym
 // Program need not convert words with length 3, and should not convert shorter words.
 // If an l (lowercase ell) would occur before a 1 (one), it should be made uppercase.
 // If an I (uppercase eye) would occur before a 1 (one), it should be made lowercase.
-// Input will be printable ASCII with spaces.
-func wordMixer() func(inputStr string) string {
-
+func wordMixer(rawInput string) string {
 	knownConversions := map[string]string{
 		"year 2038":        "Y2K38",
 		"year 2000":        "Y2K",
 		"i love you":       "143",
 		"eyjafjallajökull": "E15",
 	}
-
-	return func(inputStr string) string {
-
-		inputStr = strings.TrimSpace(inputStr)
-		runes := []rune(inputStr)
-
-		if len(runes) <= 3 {
-			return inputStr
-		}
-
-		lowerRunes := []rune(inputStr)
-		for i, v := range lowerRunes {
-			lowerRunes[i] = unicode.ToLower(v)
-		}
-		if value, found := knownConversions[string(lowerRunes)]; found {
-			return value
-		}
-
-		firstChar := runes[0]
-		lastChar := runes[len(runes)-1]
-		len := len(runes) - 2
-
-		if len > 9 && len < 20 {
-			if firstChar == 'I' {
-				firstChar = 'i'
-			} else if firstChar == 'l' {
-				firstChar = 'L'
-			}
-		}
-		return fmt.Sprintf("%c%d%c", firstChar, len, lastChar)
+	rawInput = strings.TrimSpace(rawInput)
+	rw := []rune(rawInput)
+	if len(rw) <= 3 {
+		return rawInput
 	}
+	if known, found := knownConversions[strings.ToLower(rawInput)]; found {
+		return known
+	}
+	fc := rw[0]
+	lc := rw[len(rw)-1]
+	l := len(rw) - 2
+	if l >= 10 && l <= 19 {
+		if fc == 'I' {
+			fc = 'i'
+		} else if fc == 'l' {
+			fc = 'L'
+		}
+	}
+	return fmt.Sprintf("%c%d%c", fc, l, lc)
 }
-
 func numeronyms(vals ...string) []string {
-
-	f := wordMixer()
 	results := []string{}
-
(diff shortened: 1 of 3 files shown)


04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,76 @@
...
+		len := len(runes) - 2
+
+		if len > 9 && len < 20 {
+			if firstChar == 'I' {
+				firstChar = 'i'
+			} else if firstChar == 'l' {
+				firstChar = 'L'
+			}
+		}
+<<<<<<< HEAD
```
golangcibot: expected statement, found '<<' (from `typecheck`)

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var out io.Writer = os.Stdout
+
+// following rules from https://www.revolvy.com/page/Numeronym
+// Program need not convert words with length 3, and should not convert shorter words.
+// If an l (lowercase ell) would occur before a 1 (one), it should be made uppercase.
+// If an I (uppercase eye) would occur before a 1 (one), it should be made lowercase.
+// Input will be printable ASCII with spaces.
+func wordMixer() func(inputStr string) string {
```
camscale: I don't understand why `wordMixer()` returns a function instead of just computing the result from the input. If there was some state wrapped around the closure (such as mutating `knowConversions` [adding AI to this maybe?]) then it might make more sense. But as it stands, it just complicates the code for no advantage.

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+
+		lowerRunes := []rune(inputStr)
+		for i, v := range lowerRunes {
+			lowerRunes[i] = unicode.ToLower(v)
+		}
+		if value, found := knownConversions[string(lowerRunes)]; found {
+			return value
+		}
+
+		firstChar := runes[0]
```
camscale: `Char` is deliberately not used in Go: https://blog.golang.org/strings, so best not to use that term.

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+		lastChar := runes[len(runes)-1]
+		len := len(runes) - 2
```
camscale: You are shadowing the name `len` here. After this point in the function, you can no longer call `len()`. You should not do this.

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+func wordMixer() func(inputStr string) string {
+
+	knownConversions := map[string]string{
+		"year 2038":        "Y2K38",
+		"year 2000":        "Y2K",
+		"i love you":       "143",
+		"eyjafjallajökull": "E15",
+	}
+
+	return func(inputStr string) string {
```
camscale: The naming of all the variables in this function are rather un-Go-like. Variable names should not use type prefixes/suffixes like you have: `Str`, `Runes`, `Char`, etc: https://dave.cheney.net/2019/01/29/you-shouldnt-name-your-variables-after-their-types-for-the-same-reason-you-wouldnt-name-your-pets-dog-or-cat

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+
+		inputStr = strings.TrimSpace(inputStr)
+		runes := []rune(inputStr)
+
+		if len(runes) <= 3 {
+			return inputStr
+		}
+
+		lowerRunes := []rune(inputStr)
+		for i, v := range lowerRunes {
```
camscale: No need for this loop. Just use https://golang.org/pkg/strings/#ToLower:

```
if known, found := knownConversions[strings.ToLower(input)]; found {
        return known
}
```
(i've renamed `value` to `known` because `value` is not very descriptive. If you want to use it, just call it `v`: https://talks.golang.org/2014/names.slide

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+		if value, found := knownConversions[string(lowerRunes)]; found {
+			return value
+		}
+
+		firstChar := runes[0]
+		lastChar := runes[len(runes)-1]
+		len := len(runes) - 2
+
+		if len > 9 && len < 20 {
```
camscale: I really didn't understand this logic because it didn't gel that it referred to the comment above. This is a nitpick but if it were written as `if l >= 10 && l <= 19` then it would be clearer as it actually has a couple of the numbers that cause the "problem". Using "9" and "20" kind of hides that. I am also amused that I have renamed `len` to `l` as per my previous comment and it now exhibits the same problem this rule is intended to avoid.

04_numeronym/jimbotech/numeronym_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"reflect"
+	"testing"
+)
+
+type numeronymTest struct {
+	input    string
+	expected string
+}
+
+var numTests = []numeronymTest{
```
camscale: Move this into the function that uses it. No need to be global scope. It makes the test self-contained and easier to read without having to go back and forth.

You can also remove the `numeronymTest` struct - it really isn't needed:
```
func TestWordMixer(t *testing.T) {
    tests := []struct{
        input, expected string
    }{
        {"Kubernetes, "K8s"},
        ...
    }
    ...
}
```

See: https://talks.golang.org/2012/10things.slide#3 (and the rest of that slide deck while you're there)

04_numeronym/jimbotech/numeronym_test.go
```diff
@@ -0,0 +1,65 @@
...
+	{"year 2038", "Y2K38"},
+	{"somebody said hello", "s17o"},
+}
+
+type multiTest struct {
+	input    []string
+	expected []string
+}
+
+var multiNumTests = []multiTest{
```
camscale: Same for this test data.

04_numeronym/jimbotech/numeronym_test.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, tt := range multiNumTests {
+		actual := numeronyms(tt.input...)
+		if !reflect.DeepEqual(actual, tt.expected) {
+			t.Errorf("numeronyms(%v): expected %v, actual %v", tt.input, tt.expected, actual)
+		}
+	}
+}
+
+func TestMain(t *testing.T) {
+	expected := "[a11y K8s abc]\n"
```
camscale, line 47: Why no blank line here? I don't like your blank lines at the start of functions, but if you do, just be consistent.

04_numeronym/jimbotech/numeronym_test.go
```diff
@@ -0,0 +1,65 @@
...
+		actual := f(tt.input)
+		if actual != tt.expected {
+			t.Errorf("wordMixer(%v): expected %v, actual %v", tt.input, tt.expected, actual)
+		}
+	}
+}
+
+func TestNumeronyms(t *testing.T) {
+
+	for _, tt := range multiNumTests {
```
camscale, line 39: Your table for this test has one entry. Using table-driven testing in that case makes no sense. Just use direct variables in the test function instead.

04_numeronym/jimbotech/numeronym.go
```diff
@@ -0,0 +1,72 @@
...
+	"unicode"
+)
+
+var out io.Writer = os.Stdout
+
+// following rules from https://www.revolvy.com/page/Numeronym
+// Program need not convert words with length 3, and should not convert shorter words.
+// If an l (lowercase ell) would occur before a 1 (one), it should be made uppercase.
+// If an I (uppercase eye) would occur before a 1 (one), it should be made lowercase.
+// Input will be printable ASCII with spaces.
```
camscale: Your known conversions has one non-ASCII rune. You also do not check for printable runes. This sentence is just wrong and should be removed.

camscale (changes requested):
Nice that you've decided to define and document your specification of numeronyms. I would suggest that in the real world, your linked issue should contain the specific rules you are implementing rather than linking to an external website, as those links can go stale and you lose the info that defined this implementation.

But there are some fixups needed here. Please have a read of the linked slide decks and blog posts, even if you choose to disagree with them.

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym_test.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym_test.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym.go
jimbosoft: Done

04_numeronym/jimbotech/numeronym_test.go
camscale: Doesn't look done to me. But if you really want a 1-element table-driven test, go right ahead...

camscale (approved):
LGTM. Thank you for your silent assent of my way of doing things. It does make it much easier, and it will be easier on you when I'm in power.

PR #647, opened 2019-09-12
Add Lab 10 - Puppy REST

### LINKED ISSUE #646: Lab 10 - Puppy REST

kasunfdo:
Fixes #646

Review of colleague's PR #644

- Implementation of Lab 10 - Puppy REST

### DIFF
--- a/10_rest/kasunfdo/README.md
+++ b/10_rest/kasunfdo/README.md
@@ -1,0 +1,166 @@
+# Puppy Store - An in-memory store for puppies
+
+<!-- MarkdownTOC -->
+
+- [Overview](#overview)
+- [Prerequisites](#prerequisites)
+- [Build, execute, test, lint](#betl)
+    - [Build](#build)
+    - [Execute](#execute)
+        - [Flags](#flags)
+    - [Test](#test)
+    - [Format and Lint](#lint)
+    - [Review unit test coverage](#coverage)
+- [API Documentation](#api)
+    - [Create Puppy](#create)
+    - [Read Puppy](#read)
+    - [Update Puppy](#update)
+    - [Delete Puppy](#delete)
+
+<!-- /MarkdownTOC -->
+
+## Overview
+
+Puppy REST is a simple REST web application with [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete) 
+endpoints for [Puppies](pkg/puppy/types.go). Puppy server uses an in-memory store which is designed to store 
+[Puppy](pkg/puppy/types.go) objects and store is capable of creating, reading, updating, and deleting puppies in puppy store.
+
+## Prerequisites
+
+-   Install `go 1.12` according to [official installation instruction](https://golang.org/doc/install)
+-   Clone this project outside your `$GOPATH` to enable [Go Modules](https://github.com/golang/go/wiki/Modules)
+-   Install `goimports` according to [instructions](https://godoc.org/golang.org/x/tools/cmd/goimports)
+-   Install `golangci-lint` according to [instructions](https://github.com/golangci/golangci-lint#local-installation)
+
+## Build, execute, test, lint <a name="betl"></a>
+
+#### Build
+
+    go build -o puppystore cmd/puppy-server/main.go
+    
+#### Execute
+
+    ./puppystore -d <path to puppy data file in json format>
+    
+##### Flags (Optional) <a name="flags"></a>
+
+- `--port`, `-p`: Puppy server port
+- `--store`, `-s`: Puppy store type (`map`/`sync`)
+
+#### Test
+
+    go test ./...
+
+#### Format and lint <a name="lint"></a>
+
+    goimports -w .
+    golangci-lint run
+
+#### Review unit test coverage <a name="coverage"></a>
+
+    go test -coverprofile=coverage.out ./... && go tool cover -html=coverage.out
+
+## API
+
+#### Create puppy <a name="create"></a>
+
+Add a puppy to the store.
+
+URL : `/api/puppy/`
+
+Method : `POST`
+
+Data example :
+
+```json
+{
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+##### Response (Success): 
+
+Code : `201 Created`
+
+Content example :
+
+```json
+{
+    "id"     : 1,
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+#### Read puppy <a name="read"></a>
+
+Retrieve a puppy from the store.
+
+URL : `/api/puppy/{id}`
+
+Method : `GET`
+
+##### Response (Success): 
+
+Code : `200 OK`
+
+Content example :
+
+```json
+{
+    "id"     : 1,
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+#### Update puppy <a name="update"></a>
+
+Update an existing puppy in the store.
+
+URL : `/api/puppy/`
+
+Method : `PUT`
+
+Data example :
+
+```json
+{
+    "id"     : 1,
+    "breed"  : "Labrador",
+    "colour" : "Brown",
+    "value"  : 1200
+}
+```
+
+##### Response (Success): 
+
+Code : `200 OK`
+
+Content example :
+
+```json
+    "puppy updated"
+```
+
+#### Delete puppy <a name="delete"></a>
+
+Remove a puppy from the store.
+
+URL : `/api/puppy/{id}`
+
+Method : `DELETE`
+
+##### Response (Success): 
+
+Code : `200 OK`
+
+Content example :
+
+```json
+    "puppy deleted"
+```
--- a/10_rest/kasunfdo/cmd/puppy-server/main.go
+++ b/10_rest/kasunfdo/cmd/puppy-server/main.go
@@ -1,0 +1,96 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"os"
+
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy"
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy/store"
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+)
+
+// Config contains puppyserver configurations.
+type Config struct {
+	dataFile  io.Reader
+	port      int
+	storeType string
+}
+
+func main() {
+	config, err := parseArgs(os.Args[1:])
+	if err != nil {
+		log.Fatalf("Failed to parse args.\n%v", err)
+	}
+
+	log.Fatal(initPuppyServer(config))
+}
+
+func parseArgs(args []string) (Config, error) {
+	dataFile := kingpin.Flag("data", "Path to puppy data file").Short('d').Required().File()
+	port := kingpin.Flag("port", "Puppy server port").Short('p').Default("8080").Int()
+	storeType := kingpin.Flag("store", "Puppy store type").Short('s').Default("map").Enum("map", "sync")
+	_, err := kingpin.CommandLine.Parse(args)
+
+	return Config{*dataFile, *port, *storeType}, err
+}
+
+func initPuppyServer(config Config) error {
+	puppyStore, err := createStore(config.storeType)
+	if err != nil {
+		return err
+	}
+	err = createPuppies(config.dataFile, puppyStore)
+	if err != nil {
+		return err
+	}
+
+	apiHandler := puppy.NewAPIHandler(puppyStore)
+	router := chi.NewRouter()
+	router.Use(middleware.Logger)
+	router.Use(middleware.Recoverer)
+	apiHandler.WireRoutes(router)
+
+	serverAddr := fmt.Sprintf(":%d", config.port)
+	return http.ListenAndServe(serverAddr, router)
+}
+
+func createStore(storeType string) (puppy.Storer, error) {
+	switch storeType {
+	case "map":
+		return store.NewMapStore(), nil
+	case "sync":
+		return store.NewSyncStore(), nil
+	default:
+		return nil, puppy.ErrorEf(puppy.ErrInvalid, nil, "given store type %v is invalid", storeType)
+	}
+}
+
+func createPuppies(dataFile io.Reader, store puppy.Storer) error {
+	data, err := ioutil.ReadAll(dataFile)
+	if err != nil {
+		return puppy.ErrorEf(puppy.ErrInternal, err, "failed to read puppy data file")
+	}
+
+	var puppies []puppy.Puppy
+	if err = json.Unmarshal(data, &puppies); err != nil {
+		return puppy.ErrorEf(puppy.ErrBadFormat, err, "failed to decode")
+	}
+
+	for _, p := range puppies {
+		var id uint64
+		if id, err = store.CreatePuppy(p); err != nil {
+			return err
+		}
+		fmt.Printf("%v (id: %d) added to store\n", p, id)
+	}
+
+	return nil
+}
--- a/10_rest/kasunfdo/cmd/puppy-server/main_test.go
+++ b/10_rest/kasunfdo/cmd/puppy-server/main_test.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy/store"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMain(m *testing.M) {
+	os.Args = []string{""}
+	os.Exit(m.Run())
+}
+
+func TestParseArgsLong(t *testing.T) {
+	args := []string{"--data", "../../data/data.json", "--port", "1234", "--store", "sync"}
+	config, err := parseArgs(args)
+	require.NoError(t, err)
+	assert.NotNil(t, config.dataFile)
+	assert.Equal(t, 1234, config.port)
+	assert.Equal(t, "sync", config.storeType)
+}
+
+func TestParseArgsShort(t *testing.T) {
+	args := []string{"-d", "../../data/data.json", "-p", "1234", "-s", "sync"}
+	config, err := parseArgs(args)
+	require.NoError(t, err)
+	assert.NotNil(t, config.dataFile)
+	assert.Equal(t, 1234, config.port)
+	assert.Equal(t, "sync", config.storeType)
+}
+
+func TestParseArgsWrong(t *testing.T) {
+	args := []string{"-f", "../../data/data.json", "-r", "1234", "-t", "sync"}
+	_, err := parseArgs(args)
+	assert.Error(t, err)
+}
+
+func TestParseArgsDefault(t *testing.T) {
+	args := []string{"--data", "../../data/data.json"}
+	config, err := parseArgs(args)
+	require.NoError(t, err)
+	assert.NotNil(t, config.dataFile)
+	assert.Equal(t, 8080, config.port)
+	assert.Equal(t, "map", config.storeType)
+}
+
+func TestCreateStore(t *testing.T) {
+	storer, err := createStore("map")
+	require.NoError(t, err)
+	_, ok := storer.(*store.MapStore)
+	assert.True(t, ok)
+
+	storer, err = createStore("sync")
+	require.NoError(t, err)
+	_, ok = storer.(*store.SyncStore)
+	assert.True(t, ok)
+
+	storer, err = createStore("foo")
+	require.Error(t, err)
+	require.Nil(t, storer)
+}
+
+func TestCreatePuppies(t *testing.T) {
+	s := store.NewMapStore()
+	dataFile, err := os.Open("../../data/data.json")
+	require.NoError(t, err)
+	err = createPuppies(dataFile, s)
+	assert.NoError(t, err)
+}
+
+type mockInvalidReader struct{}
+
+func (mockInvalidReader) Read(_ []byte) (n int, err error) {
+	return 0, errors.New("mockInvalidReader:error")
+}
+
+func TestCreatePuppiesReadFail(t *testing.T) {
+	dataFile := mockInvalidReader{}
+	err := createPuppies(dataFile, nil)
+	assert.Error(t, err)
+}
--- a/10_rest/kasunfdo/pkg/puppy/error.go
+++ b/10_rest/kasunfdo/pkg/puppy/error.go
@@ -1,0 +1,49 @@
+package puppy
+
+import "fmt"
+
+type ErrCode uint32
+
+type Error struct {
+	Message string
+	Code    ErrCode
+	Cause   error
+}
+
+const (
+	ErrInvalid   ErrCode = 400
+	ErrNotFound  ErrCode = 404
+	ErrBadFormat ErrCode = 406
+	ErrInternal  ErrCode = 500
+)
+
+func (e ErrCode) String() string {
+	switch e {
+	case ErrInvalid:
+		return "invalid input"
+	case ErrNotFound:
+		return "not found"
+	case ErrBadFormat:
+		return "bad data format"
+	case ErrInternal:
+		return "internal error"
+	default:
+		return "unknown error"
+	}
+}
+
+func (e *Error) Error() string {
+	msg := fmt.Sprintf("%s: %s", e.Code, e.Message)
+	if e.Cause == nil {
+		return msg
+	}
+	return fmt.Sprintf("%s\n\t%s", msg, e.Cause)
+}
+
+func ErrorEf(code ErrCode, cause error, format string, args ...interface{}) *Error {
+	return &Error{
+		Message: fmt.Sprintf(format, args...),
+		Code:    code,
+		Cause:   cause,
+	}
+}
--- a/10_rest/kasunfdo/pkg/puppy/rest.go
+++ b/10_rest/kasunfdo/pkg/puppy/rest.go
@@ -1,0 +1,101 @@
+package puppy
+
+import (
+	"net/http"
+	"strconv"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/render"
+)
+
+// APIHandler implements REST API handlers of Puppy.
+type APIHandler struct {
+	store Storer
+}
+
+// NewAPIHandler creates a APIHandler with given storer.
+func NewAPIHandler(storer Storer) *APIHandler {
+	return &APIHandler{store: storer}
+}
+
+// HandleGetPuppyByID retrieves puppy by id.
+func (a *APIHandler) HandleGetPuppyByID(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.ParseUint(chi.URLParam(r, "id"), 10, 64)
+	if err != nil {
+		sendErrorResponse(w, http.StatusBadRequest, ErrorEf(ErrInvalid, err, ""))
+		return
+	}
+
+	puppy, err := a.store.ReadPuppy(id)
+	if err != nil {
+		sendErrorResponse(w, http.StatusNotFound, err)
+		return
+	}
+	render.JSON(w, r, puppy)
+}
+
+// HandlePostPuppy adds puppy into the store.
+func (a *APIHandler) HandlePostPuppy(w http.ResponseWriter, r *http.Request) {
+	var p Puppy
+	if err := render.DecodeJSON(r.Body, &p); err != nil {
+		sendErrorResponse(w, http.StatusBadRequest, ErrorEf(ErrInvalid, err, ""))
+		return
+	}
+
+	id, err := a.store.CreatePuppy(p)
+	if err != nil {
+		sendErrorResponse(w, http.StatusBadRequest, err)
+		return
+	}
+
+	p.ID = id
+	render.Status(r, http.StatusCreated)
+	render.JSON(w, r, p)
+}
+
+// HandlePutPuppy updates puppy in the store.
+func (a *APIHandler) HandlePutPuppy(w http.ResponseWriter, r *http.Request) {
+	var p Puppy
+	if err := render.DecodeJSON(r.Body, &p); err != nil {
+		sendErrorResponse(w, http.StatusBadRequest, ErrorEf(ErrInvalid, err, ""))
+		return
+	}
+
+	if err := a.store.UpdatePuppy(p); err != nil {
+		sendErrorResponse(w, http.StatusBadRequest, err)
+		return
+	}
+
+	render.Status(r, http.StatusOK)
+	render.JSON(w, r, "puppy updated")
+}
+
+// HandleDeletePuppy deletes puppy in store by id.
+func (a *APIHandler) HandleDeletePuppy(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.ParseUint(chi.URLParam(r, "id"), 10, 64)
+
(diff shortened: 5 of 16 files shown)


10_rest/kasunfdo/pkg/puppy/rest_test.go
```diff
@@ -0,0 +1,148 @@
...
+	for _, test := range tests {
+		test := test
+		t.Run(test.name, func(t *testing.T) {
+			request, err := http.NewRequest(test.httpMethod, test.url, bytes.NewBuffer(test.payload))
+			require.NoError(t, err)
+
+			router := getTestRouter(t)
+			recorder := httptest.NewRecorder()
+			router.ServeHTTP(recorder, request)
+			response := recorder.Result()
```
golangcibot, line 139: bodyclose: response body must be closed (from `bodyclose`)

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,169 @@
...
+
+URL : `/api/puppy/`
+
+Method : `POST`
+
+Data example :
+
+```json
+{
+
```
juliaogris: unnecessary blank line.

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+	"gopkg.in/alecthomas/kingpin.v2"
+
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy"
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy/store"
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+)
+
+var (
+	out io.Writer = os.Stdout
```
juliaogris: i don't think you are using this anymore. I'm surprised you don't get a linter warning.

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+	config, err := parseArgs(os.Args[1:])
+	if err != nil {
+		panic(err)
+	}
+
+	log.Fatal(initPuppyServer(config))
+}
+
+func parseArgs(args []string) (Config, error) {
+	dataFile := kingpin.Flag("data", "Path to puppy data file").Short('d').Default("../../data/data.json").File()
```
juliaogris: bit of a strange default imo, especially if you intend to distribute the binary. if you insist on a default i'd make it `data.json` or `./data.json`  and collocate binary and datasample.

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+type Config struct {
+	dataFile  io.Reader
+	port      int
+	storeType string
+}
+
+func main() {
+	config, err := parseArgs(os.Args[1:])
+	if err != nil {
+		panic(err)
```
juliaogris: nit: maybe log.Fatalf with a message?

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func createStore(storeType string) (puppy.Storer, error) {
+	switch storeType {
+	case "map":
+		return store.NewMapStore(), nil
+	case "sync":
+		return store.NewSyncStore(), nil
+	default:
+		return nil, puppy.NewError(puppy.ErrInvalid, fmt.Sprintf("given store type %v is invalid", storeType))
```
juliaogris: I'd fold the formatting into a variadic error utility function and call fmt.Sprintf inside it:
```go
puppy.Errorf(puppy.ErrInvalid, "given store type %v is invalid", storeType)
```

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+}
+
+func createPuppies(dataFile io.Reader, store puppy.Storer) error {
+	data, err := ioutil.ReadAll(dataFile)
+	if err != nil {
+		return puppy.NewError(puppy.ErrInternal)
```
juliaogris: Consider wrapping the underlying error for more information rather than discarding it (will also play nicely with go 1.13): in the  puppy package you could add: 
```go
type Error struct {
	Message string
	Code    ErrCode
	Cause   error
}

func (e *Error) Error() string {
	msg := fmt.Sprintf("%s: %s", e.Code.String(), e.Message)
	if e.Cause == nil {
		return msg
	}
	return fmt.Sprintf("%s\n\t%s", msg, e.Cause.Error())
}

func ErrorEf(code ErrCode, cause error, format string, args ...interface{}) *Error {
	return &Error{
		Message: fmt.Sprintf(format, args...),
		Code:    code,
		Cause:   cause,
	}
}
```

10_rest/kasunfdo/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	var puppies []puppy.Puppy
+	if err = json.Unmarshal(data, &puppies); err != nil {
+		return puppy.NewError(puppy.ErrBadFormat)
```
juliaogris: often the error message returned by Unmarshal is useful, don't discard it.

10_rest/kasunfdo/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func TestCreatePuppies(t *testing.T) {
+	s := store.NewMapStore()
+	dataFile, err := os.Open("../../data/data.json")
+	require.NoError(t, err)
+	err = createPuppies(dataFile, s)
+	assert.NoError(t, err)
+}
+
+type mockInvalidReader string
```
juliaogris: `type mockInvalidReader struct {}`

I don't think you need to hold on to a string here, which would justify the string type aliasing, right?

10_rest/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,43 @@
...
+	case ErrInvalid:
+		return "invalid input. %v"
+	case ErrNotFound:
+		return "not found. %v"
+	case ErrBadFormat:
+		return "bad data format"
+	case ErrInternal:
+		return "internal error"
+	default:
+		return "error occurred"
```
juliaogris: unknown error?

10_rest/kasunfdo/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+	render.JSON(w, r, "puppy updated")
+}
+
+// HandleDeletePuppy deletes puppy in store by id.
+func (a *APIHandler) HandleDeletePuppy(w http.ResponseWriter, r *http.Request) {
+	if id, err := strconv.ParseUint(chi.URLParam(r, "id"), 10, 64); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		http.Error(w, NewError(ErrInvalid, "").Error(), http.StatusBadRequest)
+		return
+	} else if err := a.store.DeletePuppy(id); err != nil {
```
juliaogris: happy path align to left, you are returning, no need for the else

10_rest/kasunfdo/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+// HandlePutPuppy updates puppy in the store.
+func (a *APIHandler) HandlePutPuppy(w http.ResponseWriter, r *http.Request) {
+	var p Puppy
+	if err := render.DecodeJSON(r.Body, &p); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		http.Error(w, NewError(ErrInvalid, "").Error(), http.StatusBadRequest)
+		return
+	}
```
juliaogris, line 62: These 5 lines get repeated several times. You could turn them into a helper function and 3 lines:
```go
if hasError(w, render.Decode....) {
    return
}
```
and:
```
func hasError(w http.ResponseWriter, err error) {
    if err == nil {
        return
    }
    w.WriteHeader(http.StatusBadRequest)
    http.Error(w, NewError(ErrInvalid, "").Error(), http.StatusBadRequest)
}
```

or similar

juliaogris (dismissed):
as always: looks really solid, just a bunch of nitpicks.
great work!

10_rest/kasunfdo/cmd/puppy-server/main_test.go
kasunfdo: Yes! Thanks!!

10_rest/kasunfdo/pkg/puppy/rest.go
kasunfdo: `else if` is used to preserve inline variables of if clause.

10_rest/kasunfdo/pkg/puppy/rest.go
kasunfdo: Thank you for suggestion.

10_rest/kasunfdo/pkg/puppy/error.go
kasunfdo: Changed

10_rest/kasunfdo/cmd/puppy-server/main.go
kasunfdo: Noted. :)

10_rest/kasunfdo/cmd/puppy-server/main.go
kasunfdo: This is cool. Thanks!!

10_rest/kasunfdo/cmd/puppy-server/main.go
kasunfdo: Made --data/-d arg required and removed default

10_rest/kasunfdo/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+	storer, err = createStore("sync")
+	require.NoError(t, err)
+	_, ok = storer.(*store.SyncStore)
+	assert.True(t, ok)
+
+	storer, err = createStore("foo")
+	require.Error(t, err)
+	require.Nil(t, storer)
+	assert.Equal(t, "invalid input: given store type foo is invalid", err.Error())
```
camscale: nit: personally I don't feel tests like this are useful. The actual error message is not that important and does not affect functionality, unless that message is defined in some sort of interface. It is cumbersome to want to tweak an error message to have to also change the test to match.

10_rest/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,49 @@
...
+		return "bad data format"
+	case ErrInternal:
+		return "internal error"
+	default:
+		return "unknown error"
+	}
+}
+
+func (e *Error) Error() string {
+	msg := fmt.Sprintf("%s: %s", e.Code.String(), e.Message)
```
camscale: There is no need to call `String()` on `e.Code` here. Since `e.Code` implements the `Stringer` interface, that will be called automatically: https://golang.org/pkg/fmt/#Stringer

Also on that page:
> 5. If an operand implements method String() string, that method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

10_rest/kasunfdo/pkg/puppy/error.go
```diff
@@ -0,0 +1,49 @@
...
+	if e.Cause == nil {
+		return msg
+	}
+	return fmt.Sprintf("%s\n\t%s", msg, e.Cause.Error())
```
camscale: You do not need to call `Error()` here. Also from https://golang.org/pkg/fmt:

> 4. If an operand implements the error interface, the Error method will be invoked to convert the object to a string, which will then be formatted as required by the verb (if any).

10_rest/kasunfdo/pkg/puppy/rest.go
camscale: I agree with Julia here. You are hiding all the actual operations of this method inside a couple of `if` clauses just to keep the scope of a variable shorter, which doesn't really help. Keeping the happy path on the left edge makes it easier to scan-read code - you scan down the left hand side of the function and you can see what is going on:
```
id, err := strconv.ParseUint(chi.URLParam(r, "id"), 10, 64)
if err != nil {
    sendErrorResponse(...)
    return
}
if err := a.store.DeletePuppy(id); err != nil {
    sendErrorResponse(...)
    return
}
render.Status(r, http.StatusOK)
render.Status(r, http.StatusOK)
```

The `DeletePuppy` is the core of this function `HandleDeletePuppy`. In your code, the `DeletePuppy(id)` call is kind of hidden in a block of code, both vertically and horizontally, sandwiched between a couple of `sendErrorResponse` calls. When scanning code, that can make it hard to read. For less significant parts of the code, that may be ok, but as the core of this function, I think it should be a little more obvious.

This is really just my opinion and it's quite ok for you to have your own differing opinion. I'm not asking you to change this, but just pointing out a different way of looking at it. Sometimes I add the extra line of declaration and increase the scope of a variable if it makes the code more readable.

10_rest/kasunfdo/pkg/puppy/rest.go
```diff
@@ -0,0 +1,98 @@
...
+		sendErrorResponse(w, http.StatusNotFound, err)
+		return
+	}
+
+	render.Status(r, http.StatusOK)
+	render.JSON(w, r, "puppy deleted")
+}
+
+// WireRoutes route requests to corresponding REST API handler method.
+func WireRoutes(r chi.Router, apiHandler APIHandler) {
```
camscale: I'm curious as to why you chose to take `apiHandler` by value and not by pointer, particularly when you have all the receivers be by pointer. In this particular case it doesn't matter because `APIHandler` contains only a single member which is an interface, which are references my nature. But if you added another field to `APIHandler` that was a non-reference field, this `WireRoutes` function would bind to functions on a different instance of the `APIHandler` and use different values for those non-reference fields.

I would suggest making `WireRoutes` a method with a pointer receiver:
```
func (a *APIHandler) WireRoutes(r chi.Router) {
    ...
}
```
the same as you do with the HandlerFuncs.

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
+package store
+
+import (
+	"fmt"
+
+	"github.com/anz-bank/go-course/10_rest/kasunfdo/pkg/puppy"
+)
+
+// MapStore is map based implementation of Storer interface
+type MapStore struct {
```
camscale, line 10: This implementation of `MapStore` is not protected by a `sync.Mutex`. While this lab is about adding a REST interface and not implementing the `MapStore`, I'll point out that a REST interface adds concurrency to this program that was not there before, so now you have the real possibility of race conditions on this store corrupting it. I would suggest that while before it may have been justifiable that `MapStore` not be safe for concurrent access, with this lab it no longer is justifiable and should have locking added (or some other method of safe concurrent access).

It is important to understand that Go http handlers are run concurrently and that any application implementing them needs to be safe for concurrent access.

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
...
+	m.nextID++
+	m.store[p.ID] = p
+	return p.ID, nil
+}
+
+// ReadPuppy reads puppy with given id from the store
+// Will return an error if puppy with given id does not exist
+func (m *MapStore) ReadPuppy(id uint64) (puppy.Puppy, error) {
+	p, ok := m.store[id]
+
```
camscale, line 42: nit (very minor): I'd remove this blank line - setting `ok` and testing it should be together, and not separate "paragraphs".

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
...
+	if !ok {
+		return p, puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", id))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
...
+
+// UpdatePuppy update puppy with given id in store if puppy exists
+// Returns nil if puppy with given id exists. Otherwise returns an error
+func (m *MapStore) UpdatePuppy(p puppy.Puppy) error {
+	if p.Value < 0 {
+		return puppy.ErrorEf(puppy.ErrInvalid, nil, "value of puppy is negative")
+	}
+
+	if _, ok := m.store[p.ID]; !ok {
+		return puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", p.ID))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,68 @@
...
+
+	m.store[p.ID] = p
+	return nil
+}
+
+// DeletePuppy deletes puppy with given id from store
+// Returns nil if puppy with given id exists. Otherwise returns an errorZ
+func (m *MapStore) DeletePuppy(id uint64) error {
+	if _, ok := m.store[id]; !ok {
+		return puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", id))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+	id    uint64
+}
+
+func (suite *StorerTest) SetupTest() {
+	suite.id, _ = suite.store.CreatePuppy(puppy.Puppy{Breed: "Labrador", Colour: "Cream", Value: 2999.99})
+}
+
+func (suite *StorerTest) TestCreatePuppy() {
+	id, err := suite.store.CreatePuppy(puppy.Puppy{Breed: "German Shepard", Colour: "Brown", Value: 3499.99})
+	suite.True(id > 1)
```
camscale: Use `Greater` instead of `True`: https://godoc.org/github.com/stretchr/testify/assert#Greater

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+	suite.Nil(err)
+
+	id, err = suite.store.CreatePuppy(puppy.Puppy{Breed: "Terrier", Colour: "White", Value: -3499.99})
+	suite.True(id == 0)
```
camscale: Use `Equal` or `Zero` instead of `True`.

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+	customErr, _ := err.(*puppy.Error)
+	suite.Equal(puppy.ErrInvalid, customErr.Code)
+	suite.Equal("value of puppy is negative", customErr.Message)
+}
+
+func (suite *StorerTest) TestReadPuppy() {
+	p, err := suite.store.ReadPuppy(suite.id)
+
+	suite.Nil(err)
+	suite.Equal(p.ID, suite.id)
```
camscale: These `Equal` tests have their parameters backwards - it is meant to be `Equal(expected, actual)`: https://godoc.org/github.com/stretchr/testify/assert#Equal

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (suite *StorerTest) TestUpdatePuppy() {
+	err := suite.store.UpdatePuppy(puppy.Puppy{ID: suite.id, Breed: "Labrador Retriever", Colour: "Brown", Value: 3999.99})
+
+	suite.Nil(err)
+	p, err := suite.store.ReadPuppy(suite.id)
+
+	suite.Nil(err)
+	suite.Equal(p.ID, suite.id)
```
camscale: These `Equal` tests have their parameters backwards - it is meant to be `Equal(expected, actual)`: https://godoc.org/github.com/stretchr/testify/assert#Equal

However, rather than testing each field, you could assign the puppy you update at the top of the function to a variable and just do one equality test against that variable.

10_rest/kasunfdo/pkg/puppy/store/syncstore.go
```diff
@@ -0,0 +1,79 @@
...
+	return p.ID, nil
+}
+
+// ReadPuppy reads puppy with given id from the store
+// Will return an error if puppy with given id does not exist
+func (s *SyncStore) ReadPuppy(id uint64) (puppy.Puppy, error) {
+	value, ok := s.store.Load(id)
+
+	if !ok {
+		return puppy.Puppy{}, puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", id))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/syncstore.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return *value.(*puppy.Puppy), nil
+}
+
+// UpdatePuppy update puppy with given id in store if puppy exists
+// Returns nil if puppy with given id exists. Otherwise returns an error
+func (s *SyncStore) UpdatePuppy(p puppy.Puppy) error {
+	if p.Value < 0 {
```
camscale: This is the fourth place you've done identical validation. I'd suggest that it doesn't even belong in either of the stores - a puppy should be able to determine if it is valid itself:
```
func (p *Puppy) Validate() error {
    if p.Value < 0 {
        return ErrorEf(puppy.ErrInvalid, nil, "value of puppy is negative")
    }
    return nil
}
```

And then in every store function that takes in a puppy:
```
if err := p.Validate(); err != nil {
    return error  // or return 0, err in Create
}
```

That way, the logic of validation is implemented only once, and can easily be extended later and then every store Create and Update gets the extended validation for free.

10_rest/kasunfdo/pkg/puppy/store/syncstore.go
```diff
@@ -0,0 +1,79 @@
...
+		return puppy.ErrorEf(puppy.ErrInvalid, nil, "value of puppy is negative")
+	}
+
+	s.Lock()
+	defer s.Unlock()
+
+	if _, ok := s.store.Load(p.ID); !ok {
+		return puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", p.ID))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/syncstore.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// DeletePuppy deletes puppy with given id from store
+// Returns nil if puppy with given id exists. Otherwise returns an error
+func (s *SyncStore) DeletePuppy(id uint64) error {
+	s.Lock()
+	defer s.Unlock()
+
+	if _, ok := s.store.Load(id); !ok {
+		return puppy.ErrorEf(puppy.ErrNotFound, nil, fmt.Sprintf("puppy with id: %v is not found", id))
```
camscale: You don't need to use `fmt.Sprintf` here - `puppy.ErrorEf` is variadic and does `fmt.Sprintf` for you.

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,95 @@
...
+	suite.Equal(fmt.Sprintf("puppy with id: %v is not found", suite.id), customErr.Message)
+
+	err = suite.store.DeletePuppy(suite.id)
+	suite.Require().IsType(&puppy.Error{}, err)
+	customErr, _ = err.(*puppy.Error)
+	suite.Equal(puppy.ErrNotFound, customErr.Code)
+	suite.Equal(fmt.Sprintf("puppy with id: %v is not found", suite.id), customErr.Message)
+}
+
+func TestStore(t *testing.T) {
```
camscale: nit: Personally I'd like to see this function before `SetupTest`. It is the entry point to the tests in this file and makes sense to have it first to read top-down.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
+# Puppy Store - An in-memory store for puppies
+
+<!-- MarkdownTOC -->
+
+- [Overview](#overview)
+- [Prerequisites](#prerequisites)
+- [Build, execute, test, lint](#betl)
+    - [Build](#build)
+    - [Execute](#execute)
+        - [Flags](#flags)
```
camscale, line 10: this #flags link does not work, because the title of that section is "Flags(optional)". I suggest you move "(optional)" into some text in that section: "These flags are optional". Or just put in an anchor like you have for others.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+    - [Test](#test)
+    - [Format and Lint](#lint)
+    - [Review unit test coverage](#coverage)
+- [API Documentation](#API)
```
camscale: This link is not working for some reason. Maybe because there is no text under the "API" heading? Perhaps you need a manual anchor like you have for some other headers.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+    go build -o puppystore cmd/puppy-server/main.go
+    
+#### Execute
+
+    ./puppystore -d <path to puppy data file in json format>
+    
+##### Flags(Optional)
+
+- --port, -p: Puppy server port
```
camscale: nit: I would put the flags inside backquotes so they appear as fixed-width text:
- `--port`, `-p`: Puppy server port
- `--store`, `-s`: Puppy store type (`map` / `sync`)

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+- --store, -s: Puppy store type (map/sync)
+
+#### Test
+
+    go test ./...
+
+#### Format and lint <a name="lint"></a>
+
+    gofmt -w .
```
camscale: `gofmt` is not needed as `goimports` also does the formatting.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+#### Review unit test coverage <a name="coverage"></a>
+
+    go test -coverprofile=coverage.out ./... && go tool cover -html=coverage.out
+
+## API
+
+#### Create puppy <a name="create"></a>
+
+Allow users to add puppies into store.
```
camscale: "users" is unnecessary here, and it should be "puppy" (singular) because it adds one puppy:

> Add a puppy to the store.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+```json
+{
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+**Success Response**:
```
camscale: Doing this in bold makes it look the same as the previous heading which makes it look like a separate section from "Create puppy" when it is part of it. I'm not sure what it would look like if you had "##### Request" and "##### Response (success)" headers, but you could try that.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+    "id"     : 1,
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+#### Read puppy <a name="read"></a>
+
+Allow users to retrieve puppies from store.
```
camscale: > Retrieve a puppy from the store.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+    "id"     : 1,
+    "breed"  : "Labrador",
+    "colour" : "White",
+    "value"  : 1200
+}
+```
+
+#### Update puppy <a name="update"></a>
+
+Allow users to add puppies into store.
```
camscale: This description is wrong:

> Update an existing puppy in the store.

10_rest/kasunfdo/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Content example :
+
+```json
+    "puppy updated"
+```
+
+#### Delete puppy <a name="delete"></a>
+
+Allow users to add puppies into store.
```
camscale: This description is wrong:

> Remove a puppy from the store.

camscale (changes requested):
This is an excellent submission. Please don't be put off by the number of comments I have made - I would like you to get "perfection" for your last lab so I am looking very closely.

Thank you for the great docs. API docs are essential and I'm glad to see you've made an effort to document your API this way.

The rest of the code is very well done too, but there are still a few areas for improvement. I'll try to re-review more quickly.

10_rest/kasunfdo/README.md
kasunfdo: Strange! It works for me. Made it lowercase anyway.

10_rest/kasunfdo/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,93 @@
...
+	suite.Run(t, &StorerTest{store: NewSyncStore()})
+}
+
+func (suite *StorerTest) SetupTest() {
+	suite.id, _ = suite.store.CreatePuppy(puppy.Puppy{Breed: "Labrador", Colour: "Cream", Value: 2999.99})
+}
+
+func (suite *StorerTest) TestCreatePuppy() {
+	id, err := suite.store.CreatePuppy(puppy.Puppy{Breed: "German Shepard", Colour: "Brown", Value: 3499.99})
+	suite.Greater(id, suite.id)
```
golangcibot, line 28: suite.Greater undefined (type *StorerTest has no field or method Greater) (from `typecheck`)

camscale:
Thank you for updating testify in response to the golangci-lint error. :+1:

10_rest/kasunfdo/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,78 @@
...
+	p.ID = m.nextID
+	m.nextID++
+	m.store[p.ID] = p
+	return p.ID, nil
+}
+
+// ReadPuppy reads puppy with given id from the store
+// Will return an error if puppy with given id does not exist
+func (m *MapStore) ReadPuppy(id uint64) (puppy.Puppy, error) {
+	p, ok := m.store[id]
```
camscale, line 43: You're going to need to lock the mutex here too. The map itself is not safe for concurrent access. While this operation is looking up the map, a create may be modifying it. This can cause this lookup to panic or crash because it is not atomic.

This was not needed in the `sync.Map` version because that type is safe for concurrent access.

camscale (changes requested):
This is excellent code and docs. It reads very well and is quite understandable. Thanks.

Just one little thing to fix up and you'll be good to merge.

camscale (approved):
LGTM. Good, clear and comprehensive code. Thanks Kasun.

PR #649, opened 2019-09-13
Lab10- Puppy Store Rest layer

### LINKED ISSUE #650: Lab 10 - Puppy REST
- Create directory `10_rest/USERNAME` containing a copy of upstream master `09_json/USERNAME`
- Add file `pkg/puppy/rest.go` implementing:

```
GET    /api/puppy/{id}
POST   /api/puppy/          Payload: Puppy JSON without ID
PUT    /api/puppy/{id}      Payload: Puppy JSON without ID
DELETE /api/puppy/{id}
```

- Use [net/http/httptest](https://golang.org/pkg/net/http/httptest/) for testing
- Add flag `-p PORT` with long flag `--port PORT` to command line flags
- Add flag `-s STORE` with long flag `--store STORE` with accepted values:

```
map, sync, db
```
Document the API in README.md

mohitnag:
Fixes #650

Review of colleague's PR #663 

 - Adds Rest Endpoint for CRUD operations on Puppy Store
 - Adds rest tests


10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,73 @@
...
+		Handler: handler,
+	}
+	srvCh <- srv
+	if err := srv.ListenAndServe(); err != nil {
+		panic(err)
+	}
+}
+
+func initialisePuppyStoreWithFile(storeType string, fileName string) (puppy.Storer, error) {
+	store := createStore(storeType)
```
anzdaddy: Creation and initialisation probably shouldn't be a single function. I would call `createStore` in `main` and pass it to `initialisePuppyStoreWithFile`.

10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,73 @@
...
+	puppies := []puppy.Puppy{}
+	puppiesBytes := readFile(fileName)
+	if err := json.Unmarshal(puppiesBytes, &puppies); err != nil {
+		panic(err)
```
camscale: Since this function can return an error, it would be better to return this error rather than panic. Make sure you wrap it so it is more readable. e.g. something like `return fmt.Errorf("could not unmarshal json: %v", err)`

10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		panic(err)
+	}
+	handler := rest.NewRestHandler(s)
+	srv := &http.Server{
+		Addr:    ":" + (*port),
+		Handler: handler,
+	}
+	srvCh <- srv
+	if err := srv.ListenAndServe(); err != nil {
```
camscale: The docs for `http.ListenAndServe()` says it always returns an error: https://golang.org/pkg/net/http/#ListenAndServe

That makes this `if` test unnecessary. You could just say `panic(srv.ListenAndServe())`. But it is not usual to use panic in these cases, but `log.Fatal`, like `log.Fatal(srv.ListenAndServe())`

I guess you've use `panic` because it is easier to test, but you can use `log.SetOutput(buf)` in a similar way you set `out = buf` in the other labs to capture the output of main. You can verify that the buffer is not empty to see that the `log` path gets called.

10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,73 @@
...
+	s, err := initialisePuppyStoreWithFile(storeType, *fileName)
+	if err != nil {
+		panic(err)
+	}
+	handler := rest.NewRestHandler(s)
+	srv := &http.Server{
+		Addr:    ":" + (*port),
+		Handler: handler,
+	}
+	srvCh <- srv
```
camscale: This will block your program until something receives from the channel, and there is nothing in this program that does that. I know your test case does, but it means this program can only be tested and never actually run for real. That's not a very useful program (but at least you'll know it's been tested and works :) )

Instead I would remove the `make(chan *http.Server)` in the var block and leave it initialised as `nil` and change this to:
```
if srvCh != nil {
    srvCh <- srv
}
```
and in the test code, add `srv = make(chan *http.Server)`

That way the program works find on its own, but the unit tests can inject the synchronisation point.

This also should have a comment on it, because unless you read the tests, this code does not make much sense.

10_rest/mohitnag/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestReadFile(t *testing.T) {
+	assert := assert.New(t)
+	expect := puppy.Puppy{ID: 1, Breed: "dog", Colour: "white", Value: "2"}
+	actual := puppy.Puppy{}
+	buff := readFile(defaultPuppy)
+	err := json.Unmarshal(buff, &actual)
+	assert.NoError(err)
+	assert.Equal(expect, actual)
+
```
camscale: nit: unnecessary blank line

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"strconv"
+
+	"github.com/anz-bank/go-course/10_rest/mohitnag/pkg/puppy"
+)
+
+// MapStore stores Puppy details with Puppy Id as Key and Puppy  as value
+type MapStore map[uint32]puppy.Puppy
+
+func NewMapStore() *MapStore {
```
camscale, line 16: This `MapStore` needs a `sync.Mutex` now. In the previous labs, there was no concurrency, so it didn't really need it. But now you're introduced a `http.Server`, it runs the handlers in goroutines so the is the very real possibility of concurrent access to this `MapStore`. For this code to be correct, it needs to have locking.

10_rest/mohitnag/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,148 @@
...
+	store      Storer
+	makeStorer func() Storer
+	puppy      puppy.Puppy
+}
+
+func (s *storerSuite) SetupTest() {
+	s.store = s.makeStorer()
+	s.puppy = defaultPuppy(s.T())
+	err := s.store.CreatePuppy(s.puppy)
+	if err != nil {
```
camscale: Instead of writing this `if` and using `FailNow`, you should import the `testify/require` package and use `require.NoError(err)`. See https://github.com/stretchr/testify#require-package

`require` will terminate the test with `FailNow`.

I like the idea of using `require` when setting up a test and using `assert` to perform the checks that show the code is correct. Usually the `require` tests should never fail, and if they do, it indicates a problem with the test code, not a problem with the code being tested. It also makes it easy for the reader to understand what is required to set up the test properly and what is actually being tested.

10_rest/mohitnag/pkg/puppy/store/store_test.go
```diff
@@ -0,0 +1,148 @@
...
+	suite.Run(t, &storerSuite{
+		makeStorer: func() Storer { return NewSyncStore() },
+	})
+}
+
+func defaultPuppy(t *testing.T) puppy.Puppy {
+	pup := puppy.Puppy{}
+	buff := readFile("testdata/default-puppy.json")
+	if err := json.Unmarshal(buff, &pup); err != nil {
+		t.Fatalf("Couldn't unmarshall")
```
camscale: This should include the error returned from `Unmarshal` - that will tell the user what exactly was wrong with the data: `t.Fatalf("Couldn't unmarshal default-puppy.json: %v", err)`

10_rest/mohitnag/pkg/puppy/store/storer.go
```diff
@@ -0,0 +1,16 @@
+package store
+
+import (
+	"github.com/anz-bank/go-course/10_rest/mohitnag/pkg/puppy"
+)
+
+// Puppy is an alias for puppy.Puppy
```
camscale: I would remove this comment. It says exactly what the code is already saying, so it does not add any value.

10_rest/mohitnag/pkg/puppy/types.go
```diff
@@ -0,0 +1,17 @@
...
+// Puppy defines the data structure corresponding to a Puppy
+type Puppy struct {
+	ID     uint32 `json:"puppyid"`
+	Breed  string `json:"breed"`
+	Colour string `json:"colour"`
+	Value  string `json:"value"`
+}
+
+// Storer defines the interface on Puppy
+type Storer interface {
```
camscale: You should not be redefining this interface. The code should just use the `store.Storer` interface.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,121 @@
...
+	puppyID, _ := strconv.Atoi(ID)
+	puppy, err := rh.storer.ReadPuppy(uint32(puppyID))
+	if err != nil {
+		renderErrorResponse(w, r, err)
+		return
+	}
+	render.JSON(w, r, puppy)
+}
+
+func (rh *Handler) handleUpdatePuppy(w http.ResponseWriter, r *http.Request) {
```
camscale, line 80: This handler is defined with a route with `{id}`, but it does not use it. Either move it to be under the `/` route, or use the URL parameter.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,121 @@
...
+		renderErrorResponse(w, r, err)
+		return
+	}
+	w.WriteHeader(http.StatusCreated)
+}
+
+func (rh *Handler) handleGetPuppy(w http.ResponseWriter, r *http.Request) {
+	var puppy puppy.Puppy
+	ID := chi.URLParam(r, "id")
+	puppyID, _ := strconv.Atoi(ID)
```
camscale: You need to handle errors from `strconv.Atoi`, because this is unsafe user input. Do not ignore errors that come from parsing user input - that leads to security issues.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,121 @@
...
+	if err != nil {
+		renderErrorResponse(w, r, err)
+		return
+	}
+	render.JSON(w, r, puppy)
+}
+
+func (rh *Handler) handleDeletePuppy(w http.ResponseWriter, r *http.Request) {
+	ID := chi.URLParam(r, "id")
+	puppyID, _ := strconv.Atoi(ID)
```
camscale: Same as `handleGetPuppy` - you need to handle this error and not ignore it.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,121 @@
...
+	err := rh.storer.DeletePuppy(uint32(puppyID))
+	if err != nil {
+		renderErrorResponse(w, r, err)
+		return
+	}
+}
```
camscale, line 107: You don't return a response. Since there is nothing to return, it should emit a 204 No Response status code.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,121 @@
...
+func renderErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
```
camscale, line 109: nit: put a blank line between functions.

10_rest/mohitnag/pkg/rest/rest_test.go
```diff
@@ -0,0 +1,252 @@
...
+		makeStorer: func() puppy.Storer { return store.NewMapStore() },
+	})
+	suite.Run(t, &restSuite{
+		makeStorer: func() puppy.Storer { return store.NewSyncStore() },
+	})
+
+}
+
+func (s *restSuite) TestHandleGetPuppy() {
+	s.T().Run("Get an existing Puppy", func(t *testing.T) {
```
camscale: Rather than using multiple `T().Run()` calls in one test, just split it into multiple test functions: `TestHandleGetPuppy` and `TestHandleGetPuppyMissing`. This makes each test easier to read and removes a level of indentation.

Same with all the other tests in this file.

`t.Run()` is much more useful in a loop with table-driven tests.

camscale (changes requested):
Looking pretty good, Mohit. Still a little more work required, but you're almost there.

camscale:
Also, please remove `WIP` from the title of the pull request. That typically means it is NOT ready for review, but since we have labels for that, you do not need to use "WIP"

10_rest/mohitnag/cmd/puppy-server/main.go
mohitnag: Done. Separated the creation and initialisation of store

10_rest/mohitnag/cmd/puppy-server/main.go
mohitnag: Done

10_rest/mohitnag/cmd/puppy-server/main.go
mohitnag: Well actually, i introduced the `srvCh` to stop the server and to achieve 100% test coverage. I am trying to implement the suggestion

10_rest/mohitnag/cmd/puppy-server/main_test.go
mohitnag: Done

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Done

10_rest/mohitnag/pkg/rest/rest_test.go
mohitnag: Done

10_rest/mohitnag/cmd/puppy-server/main.go
mohitnag: i was not able to use `log.Fatal` as it shut down the server in case on any issue. Instead I have used `log.Panic(srv.ListenServe())`. Please have a look.

10_rest/mohitnag/pkg/puppy/store/mapstore.go
mohitnag: Agree. 
Done

10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,76 @@
+package main
```
golangcibot, line 1: /go/pkg/mod/github.com/stretchr/testify@v1.3.0/assert/assertions.go:173:18: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/mohitnag/pkg/puppy/error.go
```diff
@@ -0,0 +1,28 @@
+package puppy
```
golangcibot, line 1: /usr/local/go/src/strings/strings.go:980:20: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,63 @@
+package store
```
golangcibot, line 1: /usr/local/go/src/strings/strings.go:980:20: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,122 @@
+package rest
```
golangcibot, line 1: /usr/local/go/src/fmt/scan.go:1200:16: DecodeRuneInString not declared by package utf8 (from `typecheck`)

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Done

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Done. Please have a look.

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Done. Please have a look.

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Indeed.

10_rest/mohitnag/pkg/puppy/types.go
mohitnag: This is an honest mistake

10_rest/mohitnag/pkg/puppy/store/storer.go
mohitnag: Done

10_rest/mohitnag/pkg/puppy/store/store_test.go
mohitnag: Done

10_rest/mohitnag/pkg/puppy/store/store_test.go
mohitnag: Done

10_rest/mohitnag/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,76 @@
...
+		Handler: handler,
+	}
+	if srvCh != nil {
+		srvCh <- srv
+	}
+	log.Panic(srv.ListenAndServe())
+}
+
+func initialisePuppyStoreWithFile(store store.Storer, fileName string) error {
+
```
anzboi: Unnecessary extra line

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+	r.Use(middleware.URLFormat)
+	r.Use(render.SetContentType(render.ContentTypeJSON))
+	rh := Handler{
+		storer: s,
+		router: r}
+	rh.handleRequests()
+	return &rh
+}
+
+func (rh *Handler) handleRequests() {
```
anzboi: This name doesn't quite match what the function does. This function does not actually handle requests, it sets up a bunch of routes to tell the router how to route requests to actual handlers

10_rest/mohitnag/pkg/rest/rest_test.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+func TestStorers(t *testing.T) {
+	suite.Run(t, &restSuite{
+		makeStorer: func() store.Storer { return store.NewMapStore() },
+	})
+	suite.Run(t, &restSuite{
+		makeStorer: func() store.Storer { return store.NewSyncStore() },
+	})
+
```
anzboi: Unnecessary extra line

anzboi (dismissed):
LGTM
I have left a couple of comments on some small things you could improve on, but since the course is wrapping up, this is fine.

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,63 @@
...
+	sync.Mutex
+}
+
+func NewMapStore() *MapStore {
+	return &MapStore{m: map[uint32]puppy.Puppy{}}
+}
+
+// CreatePuppy creates a Puppy in mapstore
+func (m *MapStore) CreatePuppy(p puppy.Puppy) error {
+	if _, ok := m.m[p.ID]; ok {
```
camscale: You should not access the map without taking the lock. It needs to be held for both read and write otherwise if one thread is in the middle of a write and you do a read, there is a risk the map is in a bad state because the write is half done, and then the read crashes the application.

You should first do the validation of the puppy and the take the lock before you look in the map.

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	m.Lock()
+	defer m.Unlock()
+	m.m[p.ID] = p
+	return nil
+}
+
+// ReadPuppy reads a Puppy from mapstore
+func (m *MapStore) ReadPuppy(id uint32) (puppy.Puppy, error) {
+	if _, ok := m.m[id]; !ok {
```
camscale, line 39: likewise, you should not look up the map without the lock held.

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,63 @@
...
+		return puppy.Puppy{}, puppy.ErrorF(puppy.NotFound, "puppy with Id %d does not exists", id)
+	}
+	return m.m[id], nil
+}
+
+// UpdatePuppy updates a Puppy in mapstore
+func (m *MapStore) UpdatePuppy(p puppy.Puppy) error {
+	if _, ok := m.m[p.ID]; !ok {
```
camscale, line 49: same again. need to hold the lock

10_rest/mohitnag/pkg/puppy/store/mapstore.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	m.Lock()
+	defer m.Unlock()
+	m.m[p.ID] = p
+	return nil
+}
+
+// DeletePuppy deletes a Puppy from mapstore
+func (m *MapStore) DeletePuppy(id uint32) error {
+	if _, ok := m.m[id]; !ok {
```
camscale, line 60: once more for good luck :) You need to hold the lock here.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+	r := rh.router
+	r.Route("/api/puppy", func(r chi.Router) {
+		r.Route("/", func(r chi.Router) {
+			r.Post("/", rh.handlePostPuppy)
+			r.Put("/", rh.handleUpdatePuppy)
```
camscale, line 43: I know I said in my previous comment that you were not using the `{id}` parameter so either use it or move the handler to a path without the `{id}`. I wasn't thinking very clearly at the time I think, because the right thing to do is to use the `{id}` parameter. The idea behind REST is that the resources you manipulate are given by a URI - since update modifies an existing resource, the URI should identify that resource, so the ID should be in the URL path.

Don't worry about changing this now because I already said you could do it this way, but keep in mind in future when you do some REST programming, if your handler operates on an existing resource, the URI should identify that resource.

10_rest/mohitnag/pkg/rest/rest.go
camscale: See my comment above. I think I made a mistake suggesting you move it to be under the `/` route. But don't worry about fixing that since I've already said to do that.

10_rest/mohitnag/pkg/rest/rest.go
```diff
@@ -0,0 +1,130 @@
...
+		return
+	}
+	render.JSON(w, r, puppy)
+}
+
+func (rh *Handler) handleDeletePuppy(w http.ResponseWriter, r *http.Request) {
+	ID := chi.URLParam(r, "id")
+	puppyID, err := strconv.Atoi(ID)
+	if err != nil {
+		w.WriteHeader(http.StatusNotFound)
```
camscale, line 98: You need a `return` after this line, otherwise you try to delete a puppy with a bad ID.

You might want to look at why your test case did not pick this up. I think you'll find that when you fix this, you won't have test coverage for the `err != nil` case below too, so you might need another test.

10_rest/mohitnag/pkg/rest/rest_test.go
```diff
@@ -0,0 +1,251 @@
...
+	assert.NoError(err)
+	defer resp.Body.Close()
+	actual, err := ioutil.ReadAll(resp.Body)
+	assert.NoError(err)
+	actualPuppy := puppy.Puppy{}
+	err = json.Unmarshal(actual, &actualPuppy)
+	assert.NoError(err)
+	assert.Equal(200, resp.StatusCode)
+	assert.Equal(expectedPuppy, actualPuppy)
+
```
camscale: nit: unnecessary blank line

10_rest/mohitnag/pkg/rest/rest_test.go
```diff
@@ -0,0 +1,251 @@
...
+	assert.NoError(err)
+	defer resp.Body.Close()
+	assert.Equal(http.StatusNoContent, resp.StatusCode)
+	resp, err = http.Get(puppyServer.URL + "/api/puppy/1")
+	assert.NoError(err)
+	defer resp.Body.Close()
+	assert.Equal(404, resp.StatusCode)
+}
+
+func (s *restSuite) TestHandleDeleteMissingPuppy() {
```
camscale: I would call this `TestHandleDeleteBadID`, and when you write the next test, that should be called `TestHandleDeleteMissingPuppy`

camscale (changes requested):
A couple more fixups Mohit. Don't worry about the end of the course. I'll keep reviewing this until it is done so you can get your final lab in. You've done great work throughout the course.

10_rest/mohitnag/cmd/puppy-server/main.go
mohitnag: Done

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Renamed to `handleRoutes`.

10_rest/mohitnag/pkg/rest/rest_test.go
mohitnag: Deleted

10_rest/mohitnag/pkg/puppy/store/mapstore.go
mohitnag: Agree. This is my ignorance and the hurry to Merge the lab. Thanks for pointing it out

10_rest/mohitnag/pkg/puppy/store/mapstore.go
mohitnag: Done

10_rest/mohitnag/pkg/puppy/store/mapstore.go
mohitnag: Done

10_rest/mohitnag/pkg/puppy/store/mapstore.go
mohitnag: Done.

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Agree

10_rest/mohitnag/pkg/rest/rest.go
mohitnag: Done

10_rest/mohitnag/pkg/rest/rest_test.go
mohitnag: Deleted

10_rest/mohitnag/pkg/rest/rest_test.go
mohitnag: Done

camscale (approved):
LGTM. Great work, Mohit. Well done on making it to the end of lab 10.

PR #653, opened 2019-09-13
Add lab 6 - CRUD Puppy with interface

### LINKED ISSUE #652: Lab 6 - CRUD Puppy with interface

- Create an executable go program in directory 06_puppy/USERNAME (see hints)
- Implement a Puppy struct containing ID, Breed, Colour, Value
- Create Storer interface with crud methods for Puppy
- Write a MapStore implementation of Storer backed by a map
- Write a SyncStore implementation of Storer backed by a sync.Map
- Keep all implementation files in the same folder and in package main
- Test against the Storer interface and run in suite with both implementations

alextmz:
Fixes #652

Review of colleague's PR #443

 - Add lab 6 - CRUD Puppy with interface, tests.


06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,44 @@
...
+
+type MapStore map[uint16]*Puppy
+
+func NewmapStore() *MapStore {
+	a := MapStore{}
+	return &a
+}
+
+func (m *MapStore) CreatePuppy(p *Puppy) error {
+	if _, ok := (*m)[p.ID]; !ok {
```
sirigithub: @alextmz keen to understand why you choose to access the puppy ID as (*m)[p.ID] instead of using p.ID?

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,44 @@
...
+		return nil
+	}
+	return errors.New("puppy ID being created already exists")
+}
+
+func (m *MapStore) ReadPuppy(id uint16) (*Puppy, error) {
+	if v, ok := (*m)[id]; ok {
+		return v, nil
+	}
+	return nil, errors.New("puppy ID being read does not exist")
```
sirigithub: would be good to indicate the id being read in the error e.g  puppy with id: %d"

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (m *MapStore) UpdatePuppy(id uint16, p *Puppy) error {
+	if _, ok := (*m)[id]; ok {
+		(*m)[id] = p
+		return nil
+	}
+	return errors.New("puppy ID being updated does not exist")
```
sirigithub: same as above more info in error message.

06_puppy/alextmz/syncstore.go
```diff
@@ -0,0 +1,51 @@
...
+	_, ok := m.Load(p.ID)
+	if !ok {
+		m.Store(p.ID, *p)
+		return nil
+	}
+	return errors.New("puppy ID being created already exists")
+}
+
+func (m *SyncStore) ReadPuppy(id uint16) (*Puppy, error) {
+	r, ok := m.Load(id)
```
sirigithub: nit pick , you could combine this to one line :
if r, ok := m.Load(id); ok {
puppy := r.(Puppy)
}

06_puppy/alextmz/syncstore.go
```diff
@@ -0,0 +1,51 @@
...
+	if ok {
+		puppy := r.(Puppy)
+		return &puppy, nil
+	}
+	return nil, errors.New("puppy ID being read does not exist")
+}
+
+func (m *SyncStore) UpdatePuppy(id uint16, p *Puppy) error {
+	_, ok := m.Load(id)
```
sirigithub: similar as above can be optimised to one line

06_puppy/alextmz/storer_test.go
```diff
@@ -0,0 +1,96 @@
...
+func (su *storerSuite) SetupTest() {
+	su.p = nil
+	var p = []Puppy{
+		{ID: 256, Breed: "Dogo", Colour: "White", Value: 500},
+		{ID: 512, Breed: "Mastiff", Colour: "Brindle", Value: 700},
+		{ID: 768, Breed: "Fila", Colour: "Golden", Value: 900},
+	}
+	su.p = append(su.p, p...)
+	err := su.st.CreatePuppy(&su.p[0])
+	su.Nil(err)
```
sirigithub: @alex if this Create function fails the setup would fail consider using suite.FailNow ?

06_puppy/alextmz/mapstore.go
sirigithub: suggestion :you might have to handle the case when id !=p.Id as an exception or call out the assumption that Update overrides a puppy with given id with that provided by user.

sirigithub (commented):
@alextmz good work. A few suggestions for you to consider.

06_puppy/alextmz/mapstore.go
sirigithub: Some of the CO's pointed out in my review that a good design choice would be to get the API handle the puppy id creation instead of the user providing the ID. I think this is a good design approach to follow.

alextmz: @sirigithub Not sure if I understand your question - would not make sense to use p.ID as it is the item received, whereas (*m) is the Mapstore we're adding the item to. This specific statement checks if our mapstore already has a item with the same ID as the item received.
Indeed the API handling the ID would be a better approach. I'll look into it, thanks!

06_puppy/alextmz/mapstore.go
alextmz: Yes. Will check that.

06_puppy/alextmz/storer_test.go
alextmz: Yes, that makes a lot more sense, thanks!

06_puppy/alextmz/mapstore.go
sirigithub: @alextmz oops my bad i mis read the syntax, pl ignore my first comment.

06_puppy/alextmz/syncstore.go
alextmz: Thats OK, I usually don't like to combine statements but it does increase readability in this one

06_puppy/alextmz/main.go
```diff
@@ -0,0 +1,17 @@
+package main
```
golangcibot, line 1: /usr/local/go/src/bytes/reader.go:98:18: DecodeRune not declared by package utf8 (from `typecheck`)

06_puppy/alextmz/main.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	ps := NewmapStore()
```
juliaogris: typically: `NewMapStore`

06_puppy/alextmz/main_test.go
```diff
@@ -0,0 +1,20 @@
...
+	"bytes"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMain(t *testing.T) {
+	want := "Read Puppy by ID: {1 Dogo white 50}\n"
+
+	t.Run("main test", func(t *testing.T) {
```
juliaogris: no need for t.Run if there is only a single testcase. unless i'm missing something.

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,57 @@
...
+	switch {
+	case p == nil:
+		return fmt.Errorf("error, puppy pointer is nil")
+	case p.ID == 0:
+		p.ID = len(m) + 1
+		m[p.ID] = *p
+		return nil
+	default:
+		return fmt.Errorf("trying to create a puppy that already has an ID %d", p.ID)
+	}
```
juliaogris, line 23: no reference found, but i'd typically write this as two ifs:
```go
if p == nil {
    return fmt.Errof("...")
}
if p.ID != 0 {
    return fmt.Errof("...")
}
p.ID = len(m) + 1
m[p.ID] = *p
return nil

```

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (m MapStore) ReadPuppy(id int) (Puppy, error) {
+	if id < 0 {
+		return Puppy{}, fmt.Errorf("invalid puppy id %d", id)
+	}
```
juliaogris: i wouldn't test for this at all separately as it should be covered by the `ok` check, but if you insist, maybe just work with `uint`?

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,57 @@
...
+	if v, ok := m[id]; ok {
+		return v, nil
```
juliaogris: typically bail on error first, i.e. `!ok` ... `return err`

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	return Puppy{}, fmt.Errorf("puppy ID %d being read does not exist", id)
+}
+
+func (m MapStore) UpdatePuppy(p Puppy) error {
```
juliaogris, line 38: same comments as `ReadPuppy`

06_puppy/alextmz/mapstore.go
```diff
@@ -0,0 +1,57 @@
...
+		return fmt.Errorf("invalid puppy id %d", p.ID)
+	}
+	if _, ok := m[p.ID]; ok {
+		m[p.ID] = p
+		return nil
+	}
+	return fmt.Errorf("puppy ID %d being updated does not exist", p.ID)
+}
+
+func (m MapStore) DeletePuppy(id int) error {
```
juliaogris, line 46: same comments as `ReadPuppy`

06_puppy/alextmz/storer_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// can we create without error?
+	p1 := Puppy{Breed: "Dogo", Colour: "White", Value: 500}
+	// this value copy is here, not below, to guarantee that it is not going
+	// to be modified before being used on the relevant test
+	expected := p1
+	err := su.st.CreatePuppy(&p1)
+	su.NoError(err)
+	// do we error when creating something that already exists?
+	err = su.st.CreatePuppy(&p1)
+	su.NotNil(err)
```
juliaogris: `su.Error(err)`

06_puppy/alextmz/storer_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// what we create and what we read back match?
+	actual, _ := su.st.ReadPuppy(p1.ID)
+	actual.ID = 0
+	su.Equal(expected, actual)
+	// do we error when trying to create a puppy from a nil pointer?
+	var p4 *Puppy
+	err = su.st.CreatePuppy(p4)
+	su.NotNil(err)
```
juliaogris: `su.Error(err)`

06_puppy/alextmz/storer_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// do we error when trying to create an already identified Puppy?
+	p2 := Puppy{Breed: "Mastiff", Colour: "Brindle", Value: 700}
+	p2.ID = IDPuppyDoesNotExist
```
juliaogris: personally i'd prefer the number inlined, or defined here.

06_puppy/alextmz/storer_test.go
```diff
@@ -0,0 +1,131 @@
...
+	err = su.st.CreatePuppy(&p2)
+	su.NotNil(err)
+	// do we error when trying to create a puppy that has ID < 0?
+	p3 := Puppy{ID: -1, Breed: "Fila", Colour: "Golden", Value: 900}
+	err = su.st.CreatePuppy(&p3)
+	su.NotNil(err)
```
juliaogris: `su.Error...`
below as well.

06_puppy/alextmz/syncstore.go
```diff
@@ -0,0 +1,77 @@
...
+	sync.Mutex
+	size int
+}
+
+func NewSyncStore() *SyncStore {
+	a := SyncStore{}
+	return &a
+}
+
+func (m *SyncStore) CreatePuppy(p *Puppy) error {
```
juliaogris, line 19: same comments as mapstore

juliaogris (changes requested):
please add link to review of colleague's pr in your own pr description

06_puppy/alextmz/main.go
alextmz: Yup, agreed.

06_puppy/alextmz/main_test.go
alextmz: Agree. 
I guess I just like to see the tree in a hierarchical  way (Testfunction/file, Test function, individual tests inside the function) but I agree that there's no point to it beside cosmetics which aren't worth the price of a harder to read code, thus should be taken off. Will watch this trait as I feel I'm probably doing this frequently.

before:
```
=== RUN   TestMain/main_test
--- PASS: TestMain (0.00s)
    --- PASS: TestMain/main_test (0.00s)
```
changed to:
```
=== RUN   TestMain
--- PASS: TestMain (0.00s)
```

06_puppy/alextmz/mapstore.go
alextmz: Yes, I wrote if's but then I stumbled on https://tour.golang.org/flowcontrol/11 and it just seemed more straightforward to read and to see the code flow. 
Specifically besides the use of switch/if, I used the cases in this order as reading it goes like "nil is an error, zero is okay, and everything else is error" - nothing else to think about, you know all the error and ok conditions of flow. 
Using your test ordering one would have to read it as "nil is error, not zero is error, and everything else is okay" - leaving to the programmer to deduce further that the only okay value is zero through a double-inversion, which is straightforward to be fair, but more work than not thinking at all as in the former code, as it's explicit. 
Just seems more intuitive and I'd like to keep it that way if the language standards don't have a problem with it :-) 
Would be really curious to hear from more people about this if my reasoning makes any sense or it's some kind of beginner mistake and I don't see it, which is highly probable...

06_puppy/alextmz/mapstore.go
alextmz: Agree.
Was a bit puzzled by this one as it also affects my `switch` above, then I found https://golang.org/doc/effective_go.html#if, https://news.ycombinator.com/item?id=16678209, https://danp.net/2015/11/02/reducing-go-nesting.html. So despite not having an impact on readability of my code as there's no nesting it seems a good habit to have.

06_puppy/alextmz/mapstore.go
alextmz: Changed my mind, changing this to `if`, due to:
1. Use of switches like that is useful but gain is marginal with that small of a function.
2. Rest of the code uses `if` and it's going to look odd. 

Hah, btw found a reference: https://golang.org/doc/effective_go.html#switch

06_puppy/alextmz/mapstore.go
alextmz: I'm ok with taking those off.

06_puppy/alextmz/storer_test.go
alextmz: done

06_puppy/alextmz/storer_test.go
alextmz: done

06_puppy/alextmz/storer_test.go
alextmz: Any specific reason? They look good to me in being 2 constants that save 7 values from being hard-coded, but I agree that being globals is not ideal.

06_puppy/alextmz/storer_test.go
alextmz: done

06_puppy/alextmz/syncstore.go
alextmz: done

06_puppy/alextmz/storer_test.go
alextmz: nm - I ended up taking those off to make the tests more isolated, hope this was the reason :-)

06_puppy/alextmz/mapstore.go
juliaogris: awesome, thanks for the link to the reference.

juliaogris (approved):
lgtm

PR #615, opened 2019-08-27
Implement lab7 custom errors

nicko-lee:
Build upon lab6 + add custom errors + add thread safety in sync.map
+ write tests

Fixes #614 

Review of colleague's PR #573 

#### Changes proposed in this PR:

Implement lab7 custom errors 
Build upon lab6 + add custom errors + add thread safety in sync.map + write tests

PR #655, opened 2019-09-14
Add lab 7 - Errors

### LINKED ISSUE #654: Lab 7 - Errors
Lab 7 - Errors

- Create errors for:

        - Value < 0

alextmz:
Fixes #654

Review of colleague's PR #615

- Lab 7 - Errors


07_errors/alextmz/mapstore.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+	return nil, NewError(ErrIDBeingReadDoesNotExist)
+}
+
+func (m *MapStore) UpdatePuppy(id PuppyID, p *Puppy) error {
+	if _, ok := (*m)[id]; ok {
+		if p.Value < 0 {
+			return NewError(ErrValueLessThanZero)
+		}
+		(*m)[id] = p
```
kasunfdo: What if id is not set in p (puppy)?
It would be better if id is set as p.ID
Otherwise, remove id arg and just use p.ID

07_errors/alextmz/mapstore.go
```diff
@@ -0,0 +1,46 @@
...
+type MapStore map[PuppyID]*Puppy
+
+func NewmapStore() *MapStore {
+	a := MapStore{}
+	return &a
+}
+
+func (m *MapStore) CreatePuppy(p *Puppy) error {
+	if _, ok := (*m)[p.ID]; !ok {
+		if p.Value < 0 {
```
kasunfdo: This check could be done first as error should be returned regardless of puppy with given id already exists or not. 

```
if p.Value < 0{
    return NewError(ErrValueLessThanZero)
}

if _, ok := (*m)[p.ID]; !ok {
    .....
// Add to store
}


```

07_errors/alextmz/errors.go
```diff
@@ -0,0 +1,50 @@
...
+	Code    uint16
+}
+
+const (
+	_ = iota
+	ErrGeneric
+	_
+	ErrValueLessThanZero
+	_
+	ErrIDBeingCreatedAlreadyExists
```
kasunfdo: nit: IMO error codes can be simplified.
eg:
```
ErrInvalid   
ErrDoesNotExist
ErrInternal
```

07_errors/alextmz/storer_test.go
```diff
@@ -0,0 +1,101 @@
...
+// - If we error when trying to update what doesn't exist
+// - If the updated content matches what we expect
+func (su *storerSuite) TestUpdatePuppy() {
+	err := su.st.UpdatePuppy(su.p[0].ID, &su.p[1])
+	su.Nil(err)
+	err = su.st.UpdatePuppy(su.p[2].ID, &su.p[2])
+	su.NotNil(err)
+	err = su.st.UpdatePuppy(su.p[0].ID, &su.p[3])
+	su.NotNil(err)
+	pa, _ := su.st.ReadPuppy(su.p[0].ID)
```
kasunfdo: 🔥🔥🔥
Here you are fetching puppy by id: 256
But you are getting puppy with id 512
`{ID: 512, Breed: "Mastiff", Colour: "Brindle", Value: 700}`

07_errors/alextmz/syncstore.go
```diff
@@ -0,0 +1,61 @@
...
+	a := SyncStore{}
+	return &a
+}
+
+func (m *SyncStore) CreatePuppy(p *Puppy) error {
+	if _, ok := m.Load(p.ID); !ok {
+		if p.Value < 0 {
+			return NewError(ErrValueLessThanZero)
+		}
+		m.Lock()
```
kasunfdo: Locking should be done before `if _, ok := m.Load(p.ID); !ok {....}`

07_errors/alextmz/syncstore.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func NewSyncStore() *SyncStore {
+	a := SyncStore{}
+	return &a
+}
+
+func (m *SyncStore) CreatePuppy(p *Puppy) error {
+	if _, ok := m.Load(p.ID); !ok {
+		if p.Value < 0 {
```
kasunfdo: This check could be done first as error should be returned regardless of puppy with given id already exists or not.

07_errors/alextmz/syncstore.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	return nil, NewError(ErrIDBeingReadDoesNotExist)
+}
+
+func (m *SyncStore) UpdatePuppy(id PuppyID, p *Puppy) error {
+	if _, ok := m.Load(id); ok {
+		if p.Value < 0 {
+			return NewError(ErrValueLessThanZero)
+		}
+		m.Lock()
```
kasunfdo: Lock() before Load(...)

07_errors/alextmz/syncstore.go
```diff
@@ -0,0 +1,61 @@
...
+		m.Store(id, *p)
+		m.Unlock()
+		return nil
+	}
+	return NewError(ErrIDBeingUpdatedDoesNotExist)
+}
+
+func (m *SyncStore) DeletePuppy(id PuppyID) error {
+	_, ok := m.Load(id)
```
kasunfdo, line 63: Lock() before Load(...)

07_errors/alextmz/syncstore.go
kasunfdo: Also use defer for unlock.

```go
func (m *SyncStore) CreatePuppy(p *Puppy) error {
    if p.Value < 0 {
        return NewError(ErrValueLessThanZero)
    }

    m.Lock()
    defer m.Unlock()

    if _, ok := m.Load(p.ID); !ok {
        m.Store(p.ID, *p)
        return nil
    }

    return NewError(ErrIDBeingCreatedAlreadyExists)
}
```

07_errors/alextmz/types.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+type PuppyID uint64
+type PuppyVal int32
```
kasunfdo: Nit: This should be float type. Because prices are like 999.99 😁😁

07_errors/alextmz/storer_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+type storerSuite struct {
+	suite.Suite
+	p  []Puppy
+	st Storer
+}
+
+func (su *storerSuite) SetupTest() {
+	su.p = nil
+	var p = []Puppy{
```
kasunfdo:
```go
su.p = []Puppy{
		{ID: 256, Breed: "Dogo", Colour: "White", Value: 500},
		{ID: 512, Breed: "Mastiff", Colour: "Brindle", Value: 700},
		{ID: 768, Breed: "Fila", Colour: "Golden", Value: 900},
		{ID: 1024, Breed: "Wolfhound", Colour: "Gray", Value: -50},
	}
```

07_errors/alextmz/storer_test.go
```diff
@@ -0,0 +1,101 @@
...
+		{ID: 256, Breed: "Dogo", Colour: "White", Value: 500},
+		{ID: 512, Breed: "Mastiff", Colour: "Brindle", Value: 700},
+		{ID: 768, Breed: "Fila", Colour: "Golden", Value: 900},
+		{ID: 1024, Breed: "Wolfhound", Colour: "Gray", Value: -50},
+	}
+	su.p = append(su.p, p...)
```
kasunfdo: Nit: Better if you just assign to su.p without appending.

kasunfdo (commented):
Well organized code. Only concerns I'm having are following
- Update method implementation and test
- Locking implementation of syncmap
- You have mistakenly committed a binary file into the repo

07_errors/alextmz/mapstore.go
alextmz: Done, I'll make it a value not a pointer, refactoring a lot of stuff anyway.

07_errors/alextmz/mapstore.go
alextmz: Agree.

07_errors/alextmz/errors.go
alextmz: Agree, reading Lab 10 makes more sense to use HTTP-like codes.

07_errors/alextmz/storer_test.go
alextmz: Yes, does look strange. I'll refactor storer to deal with the ID, having it externally passable is troublesome.

07_errors/alextmz/syncstore.go
alextmz: Thanks, in the new code this I won't need to do a Load and the Locks will be in the proper places.

07_errors/alextmz/syncstore.go
alextmz: Agree

07_errors/alextmz/syncstore.go
alextmz: Yes :)

07_errors/alextmz/syncstore.go
alextmz: :+1:

07_errors/alextmz/types.go
alextmz: It's not a nit, it's a real bug man! :+1:

07_errors/alextmz/storer_test.go
alextmz: :+1:

07_errors/alextmz/storer_test.go
alextmz: Yes I don't know why I did it like that...

07_errors/alextmz/errors.go
```diff
@@ -0,0 +1,40 @@
...
+	Message string
+	Code    int
+}
+
+const (
+	Err400BadRequest     = 400
+	Err404NotFound       = 404
+	Err409Conflict       = 409
+	Err500InternalError  = 500
+	Err501NotImplemented = 501
```
juliaogris: use meaningful errors / error classification and map to http status code if/when/where needed. http status codes should not be your dimension of classification.

juliaogris (changes requested):
colleague PR review please
remove binary from commit and fix up commit history so it's removed from the history too please.
please pull all change requests from lab6 over here too.
no need for switch statements.
https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

07_errors/alextmz/errors.go
```diff
@@ -0,0 +1,47 @@
+package main
```
golangcibot, line 1: /usr/local/go/src/fmt/scan.go:1200:16: DecodeRuneInString not declared by package utf8 (from `typecheck`)

juliaogris (approved):
Looks also good to me.
I'm not 100% sure about the error codes you have chosen as the map to closely to HTTP statuses 
IMO it should be more application specific and then map to HTTP status ie internal, JSONParseError, InvalidInput, etc....
However, I'm happy to let go of this as it really is mostly a matter of taste :)

PR #661, opened 2019-09-15
Lab11 - Puppy Notifiaction

### LINKED ISSUE #658: Lab 11 - Puppy Notifications

willshen8:
Fixes #658 

Review of colleague's PR #625 

##  - Adding a lost puppy server which:
- has a single endpoint of `POST   /api/lostpuppy/          Payload: { id: PUPPY_ID }`
-  returns HTTP status 201 for even IDs
- returns HTTP status 500 for odd IDs
- Update Puppy Delete method to notify lostpuppy-service in a goroutine and log response code asynchronously.


11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,143 @@
...
+		return
+	}
+	render.JSON(w, r, "Puppy is deleted successfully")
+	go notifyLostPuppy(id, w)
+}
+
+// notifyLostPuppy sends a POST response to puppy server when a puppy is deleted
+func notifyLostPuppy(id int, w http.ResponseWriter) {
+	payload := []byte(fmt.Sprintf(`{"ID":%d}`, id))
+	resp, err := http.Post(lostPuppyServer, "application/json", bytes.NewReader(payload))
```
golangcibot: bodyclose: response body must be closed (from `bodyclose`)

willshen8: I have full code coverage, but adding `bodyclose` fails one of my tests. I will amend.

11_notify/willshen8/cmd/lostpuppy-service/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/tylerb/graceful"
+)
+
+var (
+	// graceful library shuts down the server for testing purposes
+	server = &graceful.Server{
+		Timeout: 3 * time.Second,
+	}
+	stop     chan bool
```
camscale: Since you don't care about the actual value passed on the channel (as you do not receive into a variable), it is normal in Go for this to be `chan struct{}`. That indicates to the reader that the value passed on the channel is not relevant, just the presence of a value.

11_notify/willshen8/cmd/lostpuppy-service/main.go
```diff
@@ -0,0 +1,42 @@
...
+	testMode = false
+)
+
+// runMain() stops the server after 3 seconds under test mode
+func runMain() {
+	// start the stop channel
```
camscale: this comment is incorrect - this is not "starting" a channel - it is simply making it. "start" is not a concept that applies to channels.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+type Config struct {
+	file   *os.File
+	port   int
+	storer string
+}
+
+func main() {
+	cmdConfigs, err := parseCmdArgs(args)
+	if err != nil {
+		fmt.Println(err)
```
camscale: I think you need to exit after this point. Either this should be a `panic` or you should call `os.Exit(1)`.

11_notify/willshen8/pkg/puppy/mapstore.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+// NewMapStore initialise a new MapStore
+func NewMapStore() *MapStore {
+	return &MapStore{ms: map[uint32]Puppy{}}
+}
+
+// CreatePuppy create a new puppy and store in mapStore.
+func (m *MapStore) CreatePuppy(p *Puppy) (uint32, error) {
+	i, err := strconv.Atoi(p.Value)
```
camscale: This `p.Value` validation could be simplified with the duplication removed:
```
if i, err := strconv.Atoi(p.Value); err != nil || i < 0 {
    return 0, Errorf(ErrInvalidInput, ErrInvalidInput.String())
}
```

11_notify/willshen8/pkg/puppy/error.go
```diff
@@ -0,0 +1,40 @@
...
+
+type Error struct {
+	Message string  `json:"message"`
+	Code    ErrCode `json:"code"`
+}
+
+type ErrCode uint32
+
+func (e *Error) Error() string {
+	return fmt.Sprintf("Error code %d: %s",
```
camscale: Put this all on one line. It hurts readability splitting it like this.

11_notify/willshen8/pkg/puppy/mapstore.go
```diff
@@ -0,0 +1,65 @@
...
+	if err != nil {
+		return 0, Errorf(ErrInvalidInput, ErrInvalidInput.String())
+	}
+	if i < 0 {
+		return 0, Errorf(ErrInvalidInput, ErrInvalidInput.String())
+	}
+
+	m.nextID++
```
camscale: nit: since you increment `nextID` before you use it, it really isn't the next ID, is it? It is the `lastID`. To really make it `nextID`, you should increment it after you use it.

I've labelled this comment a "nit", but it might be more important than that. At first it seems trivial, but subtle mis-naming like this leads to subtle misunderstandings which leads to bugs. Best to be precise and accurate in your naming.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/chi/middleware"
+	"github.com/go-chi/render"
+	"github.com/sirupsen/logrus"
+)
+
+const lostPuppyServer = "http://localhost:8888/api/lostpuppy/"
```
camscale, line 15: This should be passed in as configuration from the command line. You can use that value as the default, but it should not be the only possible address of the lost puppy server. You should pass this configuration to `NewRestHandler`.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+	var puppies []puppy.Puppy
+	jsonData, _ := ioutil.ReadAll(r) // error already handled by kingpin
+	if err := json.Unmarshal(jsonData, &puppies); err != nil {
+		return nil, err
+	}
+	return puppies, nil
+}
+
+// parsePortFlag check for valid port number entered from commandLine
+func parsePortFlag(c Config) (string, error) {
```
camscale: Technically this should be called `validatePortFlag` because it is not "parsing" it. However, you should be able to get rid of this altogether by using a Uint16 kingpin type when you define the flag: https://godoc.org/gopkg.in/alecthomas/kingpin.v2#ArgClause.Uint16 - that way kingpin will validate it for you.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+	if c.port < 0 || c.port > 65535 {
+		return "0", errors.New("invalid port number entered")
+	}
+	return strconv.Itoa(c.port), nil
+}
+
+// createPuppies takes an array of puppies and saves it to a puppy storer
+func createPuppies(s puppy.Storer, puppies []puppy.Puppy) (puppy.Storer, error) {
+	for _, jsonPuppy := range puppies {
+		jsonPuppy := jsonPuppy // fixes linter issue
```
camscale: dont write code just to fix linter issues and silence the linter. Either the linter is pointing out a genuine problem so you should write code to address that problem, or the linter has made a mistake or isn't smart enough to tell if something really is a problem, or you just want to do something in a way the linter doesn't like. In that case, use a `//nolint:<linter>` comment to silence the linter.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+	if storerErr != nil {
+		panic(storerErr)
+	}
+	decodedPuppies, parseErr := parseJSONPuppies(cmdConfigs.file)
+	if parseErr != nil {
+		panic(parseErr)
+	}
+	portFlag, portErr := parsePortFlag(cmdConfigs)
+	if portErr != nil {
+		fmt.Println(portErr)
```
camscale: You also need to exit after this point too. If the port was invalid, you should not use it and continue.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	port := ":" + portFlag
```
camscale: `port` is an unnecessary variable and too far away from where it is used. I would just say `ListenAndServe(":"+portFlag, router)` where you use `port` instead. The way it is now, the reader needs to remember what `port` is while reading through the `createPuppies`, `NewRestHandler`, `SetupRoute{r,s}` before seeing it used, or forget and look back when they see the `ListenAndServe`. Make it easy for people reading your code.

11_notify/willshen8/cmd/puppy-server/main.go
```diff
@@ -0,0 +1,107 @@
...
+
+	puppyStorer, _ := createPuppies(newStorer, decodedPuppies)
+	fmt.Fprintln(out, puppyStorer)
```
camscale: This looks like leftover debugging that should be removed. I can't see that it makes much sense for the application to print this out normally.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+func NewRestHandler(storer Storer) *RestHandler {
+	return &RestHandler{store: storer}
+}
+
+// HandleGet gets the puppy by id and displays the results.
+func (rh *RestHandler) HandleGet(w http.ResponseWriter, r *http.Request) {
+	id, err := strconv.Atoi(chi.URLParam(r, "id"))
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		http.Error(w, http.StatusText(http.StatusBadRequest)+": "+ErrInvalidInput.String(), http.StatusBadRequest)
```
camscale: These two lines are repeated a lot throughout this code, and in `HandlePost` it looks like you forgot to do a `w.WriteHeader(...)`. You should make a helper function that does all this for you:
```
func httpError(w http.ResponseWriter, code int, err error) {
    w.WriteHeader(code)
    http.Error(w, fmt.Sprintf("%s: %s", http.StatusText(code), err), code)
}
```
and then simplify the error sites:
```
if err != nil {
    httpError(w, http.StatusBadRequest, ErrInvalidInput)
    return
}
```
That makes it rather clear for the new reader of this code to have a good idea what is going on, and saves you from a lot of repetition and the occasional mistake.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+		return
+	}
+	puppy, readErr := rh.store.ReadPuppy(uint32(id))
```
camscale: just use `err` instead of `readErr`. It doesn't matter that you've used that name earlier in the function. Only use another name if you want to manipulate to error values together. If you're done with the last one (and you are), just keep using the name `err`.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+		http.Error(w, http.StatusText(http.StatusNotFound)+": "+ErrNotFound.String(), http.StatusNotFound)
+		return
+	}
+	render.JSON(w, r, puppy)
+}
+
+// HandlePost post a new puppy and then display the results.
+func (rh *RestHandler) HandlePost(w http.ResponseWriter, r *http.Request) {
+	var p Puppy
+	if decodeErr := render.DecodeJSON(r.Body, &p); decodeErr != nil {
```
camscale: Just call this `err` instead of `decodeErr`. Unless you have a good reason, you should call your error variables `err` when they are just used within a few lines of them being assigned.

Please do this throughout the code. I won't mention it on the other cases in this file, but please do fix them.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+		w.WriteHeader(http.StatusUnprocessableEntity)
+		http.Error(w, http.StatusText(http.StatusUnprocessableEntity)+": "+ErrInvalidInput.String(),
+			http.StatusUnprocessableEntity)
+		return
+	}
+
+	updateErr := rh.store.UpdatePuppy(uint32(id), &p)
+	if updateErr != nil {
+		// if updateError is due to invalid id
+		if updateErr.(*Error).Code == ErrNotFound {
```
camscale: This is a little bit dangerous. You are assuming that the only error type returnable by `UpdatePuppy` is `puppy.Error`. If it is not, you will panic here. You may say that `puppy.Error` is the only type returned, but that is now. Code changes over time, and this sort of code is the cause of future bugs. It is best to make sure you know what you are working with and have a default case for the unknown type of error:
```
puppyErr, ok := err.(Error)
if ok && puppyErr.Code == ErrNotFound {
    httpError(w, http.StatusNotFound, ErrNotFound)
} else if ok && puppyErr.Code == ErrInvalidInput {
    httpError(w, http.StatusBadRequest, ErrInvalidInput)
} else {
    // be careful in real code returning internal errors (err) to callers. It may leak sensitive info.
    httpError(w, http.InternalServerError, err)
}
return
```
You'll notice that by using the `httpError` helper I defined earlier and moving the `return` out of each branch since it is common to all, it is now handling more error cases with fewer lines of code, which (to me) makes this a lot more readable.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+		w.WriteHeader(http.StatusNotFound)
+		http.Error(w, http.StatusText(http.StatusNotFound)+": "+ErrInvalidInput.String(), http.StatusNotFound)
+		return
+	}
+	go rh.notifyLostPuppy(id)
+	render.JSON(w, r, "Puppy is deleted successfully")
+}
+
+// notifyLostPuppy sends a POST response to puppy server when a puppy is deleted
+func (rh *RestHandler) notifyLostPuppy(id int) {
```
camscale, line 110: This should not be a method on `RestHandler` as it does not need anything from it. Make it a pure function:
```
func notifyLostPuppy(id int) { ... }
```

I know you are using `rh.client` here, but that is an empty string because nothing ever sets it from what I can see. If it really is needed, you can just pass it in as another function argument.

11_notify/willshen8/pkg/puppy/rest.go
```diff
@@ -0,0 +1,144 @@
...
+	if err != nil {
+		logrus.Error(err)
+		return
+	}
+	defer resp.Body.Close()
+	logrus.Infof("Lost puppy with id of %d has a status code of %d", id, resp.StatusCode)
+}
+
+// SetupRoutes provides the routes for this REST API.
+func SetupRoutes(r chi.Router, rh RestHandler) {
```
camscale: It is odd that you are taking the `RestHandler` parameter by value. This makes a copy of it and if it has any non-reference fields and they change in the `RestHandler` that is passed in after this call, the handlers you set up here will operate on the wrong ones. I suggest this should be a pointer, and maybe even make this function a method on `RestHandler`. If `rh` were an interface, I might suggest leaving it as is, because that might make it easier to test by passing in a fake implementation of the interface, but it is not so I think a method makes more sense.

11_notify/willshen8/pkg/puppy/syncstore.go
```diff
@@ -0,0 +1,75 @@
...
+	return &SyncStore{syncStore: sync.Map{}}
+}
+
+// CreatePuppy create a new puppy and store in mapStore.
+func (s *SyncStore) CreatePuppy(p *Puppy) (uint32, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	i, err := strconv.Atoi(p.Value)
+	if err != nil {
+		return 0, Errorf(ErrInvalidInput, "Puppy value can't be negative")
```
camscale: this is the wrong error for this error case. The error message should be "puppy value not a number" or something.

11_notify/willshen8/pkg/puppy/syncstore.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	if i < 0 {
+		return 0, Errorf(ErrInvalidInput, "Unrecognised input")
```
camscale: this error should be "Puppy value can't be negative."

11_notify/willshen8/pkg/puppy/syncstore.go
```diff
@@ -0,0 +1,75 @@
...
+// NewMapStore initialise a new SyncStore
+func NewSyncStore() *SyncStore {
+	return &SyncStore{syncStore: sync.Map{}}
+}
+
+// CreatePuppy create a new puppy and store in mapStore.
+func (s *SyncStore) CreatePuppy(p *Puppy) (uint32, error) {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	i, err := strconv.Atoi(p.Value)
```
camscale: You do this same validation in four places. That really should be moved to a single place, especially as the mapstore version uses different error messages. I suggest you make the validation be a method on Puppy:
```
func (p* Puppy) Validate() error {
    if i, err := strconv.Atoi(p.Value); err != nil {
        return Errorf(ErrInvalidInput, "Puppy value not a number")
    } else if i < 0 {
        return Errorf(ErrInvalidInput, "Puppy value can't be negative")
    }
    return nil
}
```
and then use it in the store methods:
```
if err := p.Validate(); err != nil {
    return 0, err
}
```

11_notify/willshen8/pkg/puppy/syncstore.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		return 0, Errorf(ErrInvalidInput, "Puppy value can't be negative")
+	}
+	if i < 0 {
+		return 0, Errorf(ErrInvalidInput, "Unrecognised input")
+	}
+	s.nextID++
```
camscale: Same comment on `nextID` that I made on `MapStore` - this is not the next id if you increment it before using it.

11_notify/willshen8/pkg/puppy/mapstore.go
```diff
@@ -0,0 +1,65 @@
+package puppy
+
+import (
+	"strconv"
+)
+
+type MapStore struct {
```
camscale, line 5: This implementation of `MapStore` really needs to have a `sync.Mutex`. We haven't worried about it before because there has not been any concurrency in the program, but since lab 10 with the REST interface, the http handlers are called in goroutines so you now have concurrency, and this store can become corrupted without proper locking. It is not hard to add, so I suggest you do. I didn't review your lab 10, but I would have asked for it there. But as this is your last lab, I'd like to see it done correctly.

camscale (changes requested):
Looking good, Will. A lot of comments from me, because I always find a lot to comment on (perhaps I just talk too much). I'm really very pleased to have seen your progress from lab 1 through to this lab 11. I feel you're writing quite good Go code.

I haven't yet made any comments on your tests - if I have enough comments on the main code, I prefer to see them addressed before reviewing tests. But I will attempt to re-review not too long after you re-send this for your next review. Please ping me on #go-course if you like when you're ready.

11_notify/willshen8/pkg/puppy/rest.go
willshen8: Thanks, I agree. I have created a field inside `Config` struct to parse this from commandline

11_notify/willshen8/pkg/puppy/mapstore.go
willshen8: Yes, I agree. I have amended it so that initialising of the mapstore start at id=1, and only increment once ID has been used!

11_notify/willshen8/pkg/puppy/error.go
willshen8: Done

11_notify/willshen8/pkg/puppy/mapstore.go
willshen8: Thanks for your advice, it makes code lots concise and neat. Done.

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: Done, I have used `panics`

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: thanks, I changed the function name. But will stick with current implementation of int. As I don't want kingpin to throw an error when the port number is invalid, but rather return port 0/an open port available. Or we don't tolerate user entering an incorrect port?

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: I struggled with this one, but sure what is complaining. So have to just silence it here.

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: Done

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: I agree this is a good point about declaring a variable and use it straight away. Done

11_notify/willshen8/cmd/puppy-server/main.go
willshen8: The originally thought was to use it for testing, but no longer needed and have deleted it.

11_notify/willshen8/pkg/puppy/rest.go
willshen8: Fantastic suggestion, saved quite few lines of code.

11_notify/willshen8/pkg/puppy/rest.go
willshen8: ok, I guess it's okay since you returned earlier, and that error doesn't get used anymore. I get a weird feeling for reusing the same `err` values in the same function though, basically overwriting it for another use.

The benefits could be you don't need another error variable which could have some performance gain?

11_notify/willshen8/pkg/puppy/rest.go
willshen8: All done

11_notify/willshen8/pkg/puppy/rest.go
willshen8: This is another great pickup, Cam. I agree, there would be a default error statement just like switch statements.

willshen8: I implemented based the suggestion, but can't reach test coverage, as `updatePuppy` method doesn't return any other error types. So I'll keep this in mind in the future.

11_notify/willshen8/pkg/puppy/rest.go
willshen8: Based on your previous reviews, the `restHandler` now stores the lost puppy server as `RestHandler.client`. Therefore you can specify the lost puppy server address from command line and gets stored in there.

11_notify/willshen8/pkg/puppy/rest.go
willshen8: Thank you, making it a method of `RestHandler' is a good option.

11_notify/willshen8/pkg/puppy/syncstore.go
willshen8: Opps, mixed up with error below when refactoring it before.

11_notify/willshen8/pkg/puppy/syncstore.go
willshen8: Corrected also

11_notify/willshen8/pkg/puppy/syncstore.go
willshen8: You are a champion in code simplicity and reuse, Cam.

11_notify/willshen8/pkg/puppy/syncstore.go
willshen8: Corrected also, thanks.

11_notify/willshen8/pkg/puppy/mapstore.go
willshen8: Absolutely agree, and thanks for pointing it out.

11_notify/willshen8/cmd/lostpuppy-service/main.go
willshen8: Corrected, thanks.

11_notify/willshen8/cmd/lostpuppy-service/main.go
willshen8: This is the only comment that I am not sure of, don't I need the value to be true/false to stop the server?

11_notify/willshen8/cmd/puppy-server/main.go
camscale: It is complaining about that you are taking a reference to the loop variable and passing that reference to a function. It is described in more detail here: https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

but the gist of it is that there is only one variable for the loop variable and it changes each iteration. Because you are calling `CreatePuppy` through an interface, there is no way to know what it is doing with that reference. If it stores it or passes it to a closure or goroutine, then the value may change on the next loop iteration.

Your solution was correct - `jsonPuppy := jsonPuppy` - this creates a new variable for each iteration of the loop, so on the next iteration, the variable will not be overwritten.

My comment was with regards to the comment `// fixes linter issue`. Our goal is not to fix linter issues, our goal is to write correct code. When the linter points out an issue, we should examine what the linter is saying, and if it is correct, change the code so that our code is correct. If the linter is being too cautious and the code is correct, then we can add a `//nolint:scopelint` comment to tell the linter that the code is ok.

When I said `//nolint:<linter>` I didn't mean that exact string. You're meant to replace `<linter>` with the name of the linter that is reporting the problem (`scopelint` in this case).

So the best way to change your code is to put `jsonPuppy := jsonPuppy` back and remove the comments. Correct code needs no comments (unless it is doing something really tricky, but this idiom inside a loop in Go is quite common and most Go coders will learn to recognise what is happening there).

camscale: Another link to read: https://www.evanjones.ca/go-gotcha-loop-variables.html, or just search for [golang loop variable reference] to see a lot more that has been written about this.

11_notify/willshen8/pkg/puppy/lostpuppy.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"net/http"
+	"time"
+
+	"github.com/go-chi/chi"
+	"github.com/go-chi/render"
+)
+
+// Payload struct is the format that lost puppy server expects to see
+type Payload struct {
```
camscale: The name of this struct could be better. "Payload" is such a generic name, and when looked at fully qualified with the package name (`puppy.Payload`), it's kind of misleading because it is not the payload of a puppy at all. I'd suggest something like `LostPuppyRequest`, because that's kind of what it is - a request to the lost puppy service.

11_notify/willshen8/pkg/puppy/lostpuppy.go
```diff
@@ -0,0 +1,35 @@
...
+	ID int `json:"id"`
+}
+
+// HandleLostPuppy processes the payload and determine the response based on id field
+func HandleLostPuppy(w http.ResponseWriter, r *http.Request) {
+	time.Sleep(2 * time.Second)
+	var jsonData Payload
+	if decodeErr := render.DecodeJSON(r.Body, &jsonData); decodeErr != nil {
```
camscale: Just use `err` instead of `decodeErr`. Always use `err` for error variables unless you need to have multiple at once. As you get more experience with reading Go code, you expect to see `err` for errors and when you do, it is very obvious at a glance what is happening. When you see a different name, you look deeper to try to understand why it is different to usual and it breaks your reading flow when there is no reason for it.

11_notify/willshen8/pkg/puppy/lostpuppy.go
```diff
@@ -0,0 +1,35 @@
...
+		w.WriteHeader(http.StatusUnprocessableEntity)
+		http.Error(w, http.StatusText(http.StatusUnprocessableEntity), http.StatusUnprocessableEntity)
+		return
+	}
+	if jsonData.ID%2 == 0 {
```
camscale: nit: I would use `if jsonData%2 == 1` - get use to handling error paths inside `if` statements, and handling the normal flow at the left edge of the function: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

In this case, it is very trivial, but I think it is still good practice to stick to that.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func runTest(t *testing.T, payload []byte, expectedBody string, expectedHTTPCode int) {
+	req, err := http.NewRequest("POST", "/api/lostpuppy/", bytes.NewBuffer(payload))
+	if err != nil {
```
camscale, line 16: Since you are already using `stretchr/testify`, I would import `require` as well as `assert` and use that to verify that the tests are set up properly:
```
require.NoError(t, err)
```

it does a `t.Fatal()` under the covers, so the test will not continue, and you get a nice symmetry where test setup is verified with `require` (which aborts the test) and `assert` for verifying the results (which does not abort)`.

I think if you are not using `testify`, you can use the basic Go `testing.T` functions, but if you are using `testify`, you may as well use all its features and not mix it with basic Go testing.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
```diff
@@ -0,0 +1,71 @@
...
+	// Setup recorder
+	rr := httptest.NewRecorder()
+
+	// Execute
+	router.ServeHTTP(rr, req)
+
+	// Get actual response
+	resp := rr.Result()
+	defer resp.Body.Close()
+	body, _ := ioutil.ReadAll(resp.Body)
```
camscale: Don't ignore errors, especially in tests - it can lead you to false positives for the test. Use `require.NoError(t, err)` instead.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
```diff
@@ -0,0 +1,71 @@
...
+		testName    string
+		payload     []byte
+		HTTPCode    int
+		expectedMsg string
+	}{
+		{
+			testName:    "Test POST with nil payload",
+			payload:     nil,
+			HTTPCode:    http.StatusUnprocessableEntity,
+			expectedMsg: "Unprocessable Entity\n"},
```
camscale: nit: put the `},` at the end on a line of its own. That way, every line inside the struct body is the same - they all end with a comma. If you later add another field to the struct, the you just add a single line to the body of these tests, rather than needing to modify the line with `},` as well. This makes the diff more readable when reviewing.

It also gives visual symmetry that makes it easier to scan the code.

camscale (changes requested):
Thanks Will - looking good. I've just focussed on reviewing the lost puppy stuff that this lab adds on top of the previous one. Just some pretty simple comments really. I'm happy to re-review whenever you get a chance to work on it - I know you're on holidays. Don't worry, we'll let you get it merged and won't close the course on you. You've done a great job and come a long way from lab 1. I'm really pleased to see the quality improve as it has over the last couple of months. Thanks so much for your effort.

11_notify/willshen8/cmd/puppy-server/main_test.go
```diff
@@ -0,0 +1,163 @@
...
+// Unit test for createPuppies()
+func TestSavePuppiesToStore(t *testing.T) {
+	var buf bytes.Buffer
+	out = &buf
+	var puppies []puppy.Puppy
+	puppies = append(puppies, puppy.Puppy{ID: 1, Breed: "Jack Russell Terrier", Color: "White and Brown", Value: "1500"})
+	storer := puppy.NewMapStore()
+	returnedPuppyStorer, _ := createPuppies(storer, puppies)
+
+	expected := "&{map[1:{1 Jack Russell Terrier White and Brown 1500}] 2 {0 0}}\n"
```
camscale, line 106: Making a string from these store and comparing that is not really the best way to do this test:

1) You are vulnerable to the internal details of the storer affecting the test, as seen by the change you needed to make for this test to pass, just because you added a mutex to the storer,
2) It may be wrong. What if, for example, the dog breed stored was "Jack Russell" and the colour was "Terrier White and Brown" - it looks identical but is different. This is not likely to happen in this specific case but it does give an example of how this sort of testing can go wrong.

Better would be to compare the contents of two stores directly, or to test that the resulting store contains what you are expecting it to contain. Consider a storer that is implemented with an external SQL database - you cannot just compare the store - all you can do is check it contains what you expect it to. It may contain other things too, but your storer_test can do deeper tests that the storer does the right thing in that regard.

The best way to think about testing a function (in this case, `createPuppies`) is to look at what it does and nothing else. In this case, given a slice of puppies, it puts them in a store. So the validation is that the puppies in the slice are in the store. If there is something else in the store, that is irrelevant - you are not testing that the store works properly, just that `createPuppies` does what it claims.

No need to change this though - it's from a previous lab - I just wanted to comment and give some feedback.

camscale (commented):
Just an additional comment on the code from previous labs... No need to change this.

11_notify/willshen8/pkg/puppy/lostpuppy.go
willshen8: Cheers, I agree.

11_notify/willshen8/pkg/puppy/lostpuppy.go
willshen8: Got it!

11_notify/willshen8/pkg/puppy/lostpuppy.go
willshen8: Yes, I agree. I have amended it

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
willshen8: Thanks, good to know.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
willshen8: Thanks, I tend not wanting to swallow the error. But when I was doing it, I didn't know how to test the error. But now I know.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
willshen8: 100% agree it's more readable this way.

11_notify/willshen8/cmd/puppy-server/main_test.go
willshen8: Thanks very much for this great input. I will take care of it in the future. If I would do it again, I would test this way (based on your feedbacks) - there is no error when creating puppies and puppies are stored successfully. Agree that using these internal string representations of store is not ideal.

11_notify/willshen8/pkg/puppy/lostpuppy_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func runTest(t *testing.T, payload []byte, expectedBody string, expectedHTTPCode int) {
+	req, err := http.NewRequest("POST", "/api/lostpuppy/", bytes.NewBuffer(payload))
+	if err != nil {
+		require.NoError(t, err)
```
camscale, line 17: nit: There is no need to put the `require.NoError(t, err)` inside the `if err != nil` block. Just:
```
req, err := http.NewRequest(...)
require.NoError(t, err)
```

camscale (approved):
LGTM. Just one little nit that you can ignore and merge if you want.

Congratulations on getting all the labs and the bonus lab complete!

PR #671, opened 2019-09-29
Bump go lint versions

camscale:
WIP: Testing to see if this fixes problems.

DO NOT MERGE


camscale:
Travis is failing because the linter is OOMing. I had to increase the deadline for the linter because it was no longer completing in 1 minute. Now it won't complete with the amount of memory travis gives to it.

I'm going to abandon this PR. We'll stick to the current Go and lint versions, and just ignore the bogus golangci.com lint errors.

PR #674, opened 2019-10-01
Add lab1 main and test

smartshark27:
Fixes #

Review of colleague's PR #

## -
Add lab1 main and test

### DIFF
--- a/01_fib/smartshark27/main.go
+++ b/01_fib/smartshark27/main.go
@@ -1,0 +1,25 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func fib(n int) {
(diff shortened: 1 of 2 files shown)


camscale (commented):
I'm sorry smarkshark27, but the course has closed to new PRs now (mid September).

But some comments nonetheless:
1) You've skipped parts that are required - the PR template would have given you an inkling of that. See https://github.com/anz-bank/go-course/wiki/Software-Engineering-with-Go for more details.

2) You can declare multiple vars at once and not use the `var` keyword: `prev, num := 0, 1`

3) You can do multiple assignments at once, so you don't need `tmp`: `prev, num = num, prev+num`

4) Your unit tests are inadequate. You should have tests for the `fib` function.

juliaogris:

PR #605, opened 2019-08-23
Implement Lab6  CRUD puppystore

sirigithub:
Fixes #606

Review of colleague's PR: #653

#### Changes proposed in this PR:
- Rework create logic to handle puppy ID creation as per PR review suggestion.
- Updated error messages as per code review comments.
- Created an executable go program in directory 06_puppy/sirigithub
- Implemented a Puppy struct containing ID, Breed, Colour, Value
- Created Storer interface with crud methods for Puppy
- Created a MapStore implementation of Storer backed by a map
- Created a SyncStore implementation of Storer backed by a sync.Map
- Written tests against the Storer interface for both implementations
- Implemented a main class and basic tests for both Storer implementations.

PR #679, opened 2019-10-01
Implementation of puppy store 

### LINKED ISSUE #680: Lab 6 - CRUD puppy with interface
Implemented CRUD interface 
type Storer interface {
	CreatePuppy(p *Puppy) (uint32, error)
	ReadPuppy(ID uint32) (*Puppy, error)
	UpdatePuppy(ID uint32, Puppy *Puppy) error
	DeletePuppy(ID uint32)
}
and two ways to store it using map and sync.map. Accompanied by unit tests using testify suit.

jimbosoft:
CRUD interface and storage using map and sync.map

Fixes #680 

Review of colleague's PR #605 and #533 

Implementation of the puppy CRUD interface and storing it in map and sync.map


06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// NewPuppyStorer constructor creates the map
+func NewPuppyStorer() MapStore {
+	return MapStore{}
+}
+
+// checkCreation might not be required if it is not possible
+// to create a Mapstore without calling the above constructor
+func checkCreation(s MapStore) error {
```
camscale: rather than making this a function, i think the code would be clearer if you just did the test at the top of the function where this is called. You're then replacing one `if` with another, so it's not more code. You can use a global error so you do not need to repeat the text. This will also make the test case easier and remove the need for `t.Run()` in that test case:

```go
var ErrNotConstructed = errors.New("store not created")
...
func (s MapStore) CreatePuppy(p *Puppy) (int32, error) {
    if s == nil {
        return 0, ErrNotConstructed
    }
    ...
}
```

To me, that is a lot clearer what is going on - i see you comparing `s` to `nil`, I see the error value `ErrNotConstructed`, and there is less code all around. I don't need to look somewhere else to see what this `checkCreation` thing is.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+	s.Range(func(key interface{}, value interface{}) bool {
+		length++
+		return true
+	})
+	return length
+}
+
+// CreatePuppy add a puppy to storage
+func (s MapStore) CreatePuppy(p *Puppy) (uint32, error) {
+	if err := checkCreation(s); err != nil {
```
camscale: I suggest this is not really needed - if the caller has not created the map, that's their problem. It may be worth adding a `func NewMapStore() MapStore` constructor function so the caller doesn't need to know about the implementation which would allow you to change it to a struct later if you needed to add a mutex for example. The caller would not need to change. The constructor could then be responsible for creating the `map`.

If you want the safety of ensuring that the constructor is called, you can make `MapStore` be `mapStore`. If that were in a different package, a user of that package could not create a `mapStore` directly - they would have to use the constructor function to create it.

See http://www.golangpatterns.info/object-oriented/constructors for a small example.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+		return 0, err
+	}
+	p.ID = uuid.New().ID()
+	s[p.ID] = p
+	return p.ID, nil
+}
+
+// CreatePuppy threadsafe adding a puppy to storage
+func (s *SyncMapStore) CreatePuppy(p *Puppy) (uint32, error) {
+	p.ID = uuid.New().ID()
```
camscale: I like this. A nice simple stateless way to generate IDs.

camscale: Since this is mutating the value that is passed in by pointer, I would document that in the method docstring

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+	s.Store(p.ID, p)
```
camscale: You should make a copy of `p` before storing it. Otherwise the caller can logically corrupt the store by changing the ID, for instance in their copy, which is also the `MapStore`s copy. Not storing a copy also goes against the purpose of the `UpdatePuppy` method - it's not needed if you can just change the puppy via the pointer.

But most importantly, not storing a copy violates the semantics of the interface and allows subtly different  implementations of the interface. Consider a `Storer` implementation that stores the puppies in a remote SQL database. If the caller changed the value of the puppy through the pointer, it obviously would not change what was in the SQL database. But if you were using a `MapStore`, it would change it. The application could then behave differently depending on which implementation of the interface is used, even though it uses the interface in the same way.

This applies to the `MapStore` too.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+
+// MapStore stores puppies
+type MapStore map[uint32]*Puppy
+
+// SyncMapStore stores puppies threadsafe
+type SyncMapStore struct {
+	sync.Map
+}
+
+type mapCheck interface {
```
camscale: nit: I'd call this `mapTest`, just to be clear it is for testing. A comment to that effect would also not go astray. Then the reader does not need to be concerned about the poor performance of one of the implementations of this interface.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+	length() int
+}
+
+// length used for testing
+func (s MapStore) length() int {
+	return len(s)
+}
+
+func (s *SyncMapStore) length() int {
```
camscale: If you're going to bung all this in one file, I personally would find it more readable if the implementations were kept together - all `MapStore` methods, then all `SyncMapStore` methods. But consider splitting this into three files: `storer.go` that defines the `Storer` interface and the `Puppy` struct (and probably the `mapCheck` interface), and `mapstore.go` and `syncmapstore.go` for the implementations of each. Then for testing you would have `storer_test.go` which tests each implementation via the storer interface. Then you can add a separate `mapstore_test.go` if necessary to test specific facets of `MapStore`, and similarly for `SyncMapStore`.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+	var length int
+	s.Range(func(key interface{}, value interface{}) bool {
```
camscale: This doesn't really give a true length WRT concurrency. The map can change while you're counting the length, so it may not be accurate. Maybe it should have a comment noting this so future users do not assume it is accurate.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+
+// UpdatePuppy update your puppy store
+func (s MapStore) UpdatePuppy(id uint32, puppy *Puppy) error {
+	if err := checkCreation(s); err != nil {
+		return err
+	}
+	if res := s[id]; res == nil {
+		return fmt.Errorf("no puppy with ID %v found", id)
+	}
+	s[id] = puppy
```
camscale: This should either validate that `id == puppy.ID` or enforce it: `puppy.ID = id`. If you do enforce it, what if it changes? An update then would be to remove the old ID and add the new ID.

06_puppy/jimbotech/mapStore.go
```diff
@@ -0,0 +1,138 @@
...
+	delete(s, id)
+}
+
+// DeletePuppy threadsafe removal of the puppy from store
+func (s *SyncMapStore) DeletePuppy(id uint32) {
+	s.Delete(id)
+}
+
+// NewPuppyStorer constructor creates the map
+func NewPuppyStorer() MapStore {
```
camscale: This should be called `NewMapStore` - it is standard for basic Go constructor functions to be named `New<type>`

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+type storesSuite struct {
+	suite.Suite
+	store  Storer
+	mapper mapCheck
+}
+
+func TestStorer(t *testing.T) {
+	m := NewPuppyStorer()
```
camscale: Rather than creating the stores here, instead store a constructor function in the suite, and add a `SetupTest` method on the suite that calls the function to create the store. That way it has a consistent and known state at the start of each test.

Or, you can leave these here as is, and in `SetupTest`, re-create them based on a type switch:
```
func (s *storeSuite) SetupTest() {
    switch s.store.(type) {
    case MapStore:
        s.store = NewMapStore()
    case *SyncMapStore:
        s.store = &SyncMapStore{}
    default:
        s.Fail("Unknown Storer implementation")
    }
    s.mapper = s.store.(mapCheck)
}
```

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+	_, err = puppyStore.ReadPuppy(1)
+	t.Run("ReadPuppy test", f)
+	puppyStore.DeletePuppy(1)
+}
+
+// TestCreateSuccess add to the store and verify
+// by reading that it is in the store and then remove it
+// to leave the store in the original state
+func (s *storesSuite) TestCreateSuccess() {
+	asserter := assert.New(s.T())
```
camscale: You don't need `asserter` - your `suite` embed an `assert.Assertions` so you can just use `s.Nil` or `s.Equal`, etc. You get all the testify/assert methods.

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+	if success, pup := create(s); success {
```
camscale: I like to use `testify/require` for this - require makes the test fail right away, so I use it to verify that the test is set up properly and not continue if not.
```
success, pup := create(s)
require.True(success)
```

... now I've looked at `create()` I see that you probably don't want to use assert/require here. If you use `require` in `create()` you can assume it always returns a puppy, or does not return at all.

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+		// now check by reading the value back and compare
+		pup2, err2 := s.store.ReadPuppy(pup.ID)
+		if asserter.Nil(err2, "Read store should work") {
```
camscale: Use `require.NoError(...)` to test for no errors. No need for the comment either. The error should say what went wrong.

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+			asserter.Equal(*pup, *pup2)
```
camscale: No need to dereference the pointers - it will compare deeply:
```
s.Equal(pup, pup2)
```

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+		}
+		s.store.DeletePuppy(pup.ID)
```
camscale: No need to delete the puppy if you set up a fresh store for each test using `SetupTest()`

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func create(s *storesSuite) (bool, *Puppy) {
+	asserter := assert.New(s.T())
+	pup := Puppy{1, "kelpie", "brown", "indispensable"}
+	id, err := s.store.CreatePuppy(&pup)
+	if asserter.Nil(err, "Create on the store should have worked") {
```
camscale: If you use `require.NoError(err)` here, you don't need the nesting, and you can get rid of the bool return value.

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// TestMapStoreWithoutContructor may not be required as it may not be
+// possible to create an instance of that type without using the constructor
+func TestMapStoreWithoutContructor(t *testing.T) {
+	var puppyStore MapStore
+	pup := Puppy{1, "kelpie", "brown", "indispensable"}
+
+	_, err := puppyStore.CreatePuppy(&pup)
+	f := func(t *testing.T) {
```
camscale: If you make the change I suggested earlier with the `checkCreation()` function, this test case would become a lot easier to read - you would not need this anonymous function or use t.Run(). Instead you'd just:
```
_, err := puppyStore.CreatePuppy(&pup)
assert.Equal(ErrNotConstructed, err)
err = puppyStore.UpdatePuppy(1, &pup)
assert.Equal(ErrNotConstructed, err)
...
```

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+func TestStorer(t *testing.T) {
+	m := NewPuppyStorer()
+	ms := &SyncMapStore{}
+	suite.Run(t, &storesSuite{store: m, mapper: m})
+	suite.Run(t, &storesSuite{store: ms, mapper: ms})
+}
+
+// TestMapStoreWithoutContructor may not be required as it may not be
+// possible to create an instance of that type without using the constructor
+func TestMapStoreWithoutContructor(t *testing.T) {
```
camscale: If you split up the files as I suggested earlier, this function would belong in `mapstore_test.go` - assuming you even keep the `ErrNotConstructed` error.

06_puppy/jimbotech/mapStore_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+//TestUpdateFailure checks the error returned when updating with an invalid id
+func (s *storesSuite) TestUpdateFailure() {
+	asserter := assert.New(s.T())
+	if success, pup := create(s); success {
+		pup2 := Puppy{1, "kelpie", "black", "indispensable"}
+		err := s.store.UpdatePuppy(1, &pup2)
+		if asserter.NotNil(err, "Update on id 1 should have failed") {
+			st := fmt.Sprintf("no puppy with ID %v found", 1)
+			asserter.Equal(st, err.Error())
```
camscale: This starts getting quite nested. Generally good Go code keeps the happy path aligned to the left edge: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

06_puppy/jimbotech/puppy.go
```diff
@@ -0,0 +1,18 @@
...
+	"fmt"
+	"io"
+	"os"
+)
+
+var out io.Writer = os.Stdout
+
+func main() {
+	var puppyStore Storer = NewPuppyStorer()
+	pup := Puppy{1, "kelpie", "brown", "indispensable"}
```
camscale: `indispensable` is not a colour, and the evidence I saw showed that the brown kelpie was not indispensable.

More seriously, the order of these parameters does not match the order declared in the struct. I suggest you use named parameters:
```
pup := Puppy{Breed: "kelpie", Colour: "brown", breed: "indispensable"}
```
That means you are now correct, and you get to leave out `ID`, which it does not make sense to put it because you are not meant to set that.

Since you've made this mistake throughout your code, you should be asking why your tests did not catch this, and how you can update your tests to catch it.

hint:
```
s.Equal("brown", p.Colour)
...
```
instead of
```
s.Equal(pup1, pup2)
```

06_puppy/jimbotech/puppy_test.go
```diff
@@ -0,0 +1,17 @@
+package main
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestMain2(t *testing.T) {
```
camscale: I wonder what ever happened to `TestMain1` ? Perhaps this could take its place, and even be called `TestMain`?

camscale (commented):
Just a few comments, less than I usually make. Some things I commented on occurs multiple times and I did not comment on each occurrence, so do check the rest of the code to apply the same comments elsewhere.

06_puppy/jimbotech/mapStore.go
jimbosoft: Great thanks. I did try that, but then the linter attacked me ... I thought my idea was right, thanks for confirming. Now I just have to shut the linter up somehow ...

camscale: Shut up the linter with `//nolint:<linter>` - replace `<linter>` with the name of the linter complaining (in brackets at the end of the lint message). Put that comment at the end of the line that it is complaining about, or for block-level ignores, put it in the block aligned with the code.

06_puppy/jimbotech/mapStore.go
jimbosoft: Done

06_puppy/jimbotech/mapStore.go
jimbosoft: Next time ...

06_puppy/jimbotech/mapStore.go
jimbosoft: Done

06_puppy/jimbotech/mapStore.go
jimbosoft: Both done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: yeap, much simpler, done

06_puppy/jimbotech/mapStore.go
jimbosoft: Done

06_puppy/jimbotech/mapStore.go
jimbosoft: Done

06_puppy/jimbotech/mapStore.go
jimbosoft: I quoted the go doc for "range" in my comment.

06_puppy/jimbotech/mapStore.go
jimbosoft: Also forgot to copy the puppy ... all resolved now

06_puppy/jimbotech/mapStore.go
jimbosoft: Done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Noice. Spend a long time trying different things and failing, till I saw this comment. Done.

06_puppy/jimbotech/mapStore_test.go
jimbosoft: That's what you get if you copy code and not really understand it. Fixed now.

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Because I did not see this comment first, I spent a lot of time left aligning my code. Now with this comment I see that I wasted all that effort as this more or less left aligns it anyway. I've gone crazy with require now. Originally I assumed that is how assert works, as assert in other languages terminates the test. It's require for me all the way now ... and I thought I was so clever for testing the response from assert to avoid continuing ... kill me now, will this project ever end?

06_puppy/jimbotech/mapStore_test.go
jimbosoft: I guess that's what errors do, so why is there an option for a message ... Done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Done

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Split doen

06_puppy/jimbotech/mapStore_test.go
jimbosoft: Why did I not hear about "require" before I did this ... Done

06_puppy/jimbotech/puppy_test.go
jimbosoft: It is now, done

06_puppy/jimbotech/puppy.go
jimbosoft: I think you were getting tired ... there is a lot of breeding going on here ... apart from that the order was correct ... but it is a much better idea to use name parameters as you suggest, so you win again. Hurray!

06_puppy/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,64 @@
...
+		return true
+	})
+	return length
+}
+
+// CreatePuppy threadsafe adding a puppy to storage
+// but will modify the member ID.
+func (s *SyncMapStore) CreatePuppy(p *Puppy) (uint32, error) {
+	p.ID = uuid.New().ID()
+	sp := p
```
camscale: This is not making a copy of the Puppy - you're just assigning the pointer. You fixed this in mapStorem but not here by the looks of it.

Maybe you should add test cases to ensure that the caller cannot change puppy in the store through either a pointer passed to the store or a pointer returned from the store.

06_puppy/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,64 @@
...
+	return p.ID, nil
+}
+
+// ReadPuppy threadsafe retrieval of your puppy.
+func (s *SyncMapStore) ReadPuppy(id uint32) (*Puppy, error) {
+	val, found := s.Load(id)
+	if !found {
+		return nil, fmt.Errorf("no puppy with ID %v found", id)
+	}
+	return val.(*Puppy), nil
```
camscale: This returns a pointer to the puppy in the store.

06_puppy/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// UpdatePuppy threadsafe update your puppy store.
+func (s *SyncMapStore) UpdatePuppy(id uint32, puppy *Puppy) error {
+	_, found := s.Load(id)
+	if found {
+		puppy.ID = id
+		sp := puppy
```
camscale: This is not copying the puppy

06_puppy/jimbotech/storer_test.go
```diff
@@ -0,0 +1,112 @@
...
+type storesSuite struct {
+	suite.Suite
+	store  Storer
+	mapper mapTest
+}
+
+func TestSuite(t *testing.T) {
+	var ms Storer = NewMapStore()
+	var sms Storer = &SyncMapStore{}
+	suite.Run(t, &storesSuite{store: ms, mapper: ms.(mapTest)})
```
camscale: You could make this `suite.Run(t, &storesSuite{store: NewMapStore()})` - you don't need the var because you don't need to initialise the `mapper` field as that is initialised by the `SetupTest` function. `SetupTest` just needs the `store` field to be initialised. Just a thought - no need to make this change:

```
func TestSuite(t *testing.T) {
    suite.Run(t, &storesSuite{store: NewMapStore()})
    suite.Run(t, &storesSuite{store: &SyncMapStore{}})
}
```

I find shorter functions with no extraneous guff easier to read.

camscale (dismissed):
Minor comments. You know what you're doing here, so i'll approve for merging. You can choose to address the comments or not.

06_puppy/jimbotech/sync_mapstore.go
```diff
@@ -49,13 +49,13 @@ func (s *SyncMapStore) ReadPuppy(id uint32) (*Puppy, error) {
 // UpdatePuppy threadsafe update your puppy store.
 func (s *SyncMapStore) UpdatePuppy(id uint32, puppy *Puppy) error {
 	_, found := s.Load(id)
-	if found {
-		puppy.ID = id
-		sp := *puppy
-		s.Store(id, &sp)
-		return nil
+	if !found {
```
camscale, line 52: minor nitpick: you could change this to `if _, found := s.Load(id); !found {`

And while `found` is fine as a name, the name `ok` is used far more often, so much so that it is called the "comma ok idiom" - `if _, ok := s.Load(id); !ok {` - for a while "found" seems like a better name because it is more descriptive, but after a while, "ok" becomes standard (it is also used when receiving from channels: `v, ok := <-ch`). So you'll see "ok" used a lot here.

camscale (approved):
LGTM

PR #681, opened 2019-10-27
Lab7 - Errors

### LINKED ISSUE #682: Lab 7 - Errors
Used Lab 6 code and added:
Error type with fields message and code
mutex to avoid access conflicts

jimbosoft:
Fixes #682

Review of colleague's PR #

Lab 7 - adding error type and added mutex locking to the sync store


07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,48 @@
...
+	s.Require().Equal(ErrIDNotFound, err)
+}
+
+func (s *storesSuite) TestError() {
+	var err = &Error{}
+	err = nil
+	res := err.Error()
+	s.Assert().Equal("<nil>", res)
+}
+
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion

```

07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,45 @@
+package main
+
+func (s *storesSuite) TestReadFailure() {
```
camscale, line 3: These test functions should be `storer_test.go` along with the other tests. Splitting it up like this just removes the context needed to understand the tests. Also, keeping the successful and failure tests together makes it easier to understand the unit under test from the test cases.

07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,45 @@
...
+	pup2, err := s.store.ReadPuppy(1)
```
camscale, line 4: Why is this `pup2`? Where is `pup` (pup1)? Same applies to the other tests in this file. You use `pup2` when there is no original pup.

This matters because it can be confusing. When I saw `pup2`, I assumed that the test fixture was setting up pup1 for all the tests in the suite, so I went to look at that to see what pup1 was and could not find it. I'm stil not 100% that there is not a pup1 hiding somewhere else (another downside to splitting these tests into a different file - now I don't know how many files I need to look at to find what I'm looking for) and how it may affect these tests.

07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,45 @@
...
+	s.Require().Nil(pup2)
+	s.Require().Equal(ErrIDNotFound, err)
+	s.Require().NotEmpty(err.Error())
```
camscale, line 7: This is an odd thing to test. If you want to test that the `ErrIDNotFound` presents as a non-empty string, that test is completely separate from the store test. Furthermore, why is "empty" the useful test for this? If the string was "skdjjfhaksdjafs" - would that be OK?

07_errors/jimbotech/error_test.go
camscale, line 3: My style for naming tests would be to name this `TestReadIDNotFound`, and the next would be `TestReadValueBelowZero`. The tests share the same prefix (`TestRead`) so I know they are grouped, and the "failure" is specific about exactly which failure is being tested.

This is just my style though, but I think it helps.

07_errors/jimbotech/storer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	s.True(pup.Colour == black)
+	s.NotEqual(pup, pup2)
+}
+
+func create(s *storesSuite) *Puppy {
+	pup := Puppy{Breed: "kelpie", Colour: brown, Value: "indispensable"}
+	id, err := s.store.CreatePuppy(&pup)
+	s.Require().NoError(err)
+	s.Require().NotEqual(pup.ID, uint32(1))
+	s.Require().Equal(id, pup.ID, "Pup id must be set to actual id")
```
camscale: This require/assert belongs in `TestCreateSuccess` as that is an invariant of the `CreatePuppy` method. This function is a text fixture called from many places, which means if the storer returned a bad id such that this fails, every test that uses this fixture function will also fail even though what they are testing themselves may well be correct.

07_errors/jimbotech/storer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	s.True(pup2.Colour == brown)
+	s.True(pup.Colour == black)
+	s.NotEqual(pup, pup2)
+}
+
+func create(s *storesSuite) *Puppy {
+	pup := Puppy{Breed: "kelpie", Colour: brown, Value: "indispensable"}
+	id, err := s.store.CreatePuppy(&pup)
+	s.Require().NoError(err)
+	s.Require().NotEqual(pup.ID, uint32(1))
```
camscale: this is non-obvious. Why can't `pup.ID` be `1`? I assume that this fixture function is intended to be called after a test has already created a pup? If so, at the very least this function needs a comment saying so. But I would even say this test is not needed at all - that the storer does not reuse IDs is a test on the `CreatePuppy` function and that test belongs in a test specifically for it.

[time passes as review continues]

oh wow. you generate IDs using `rand`. This makes less sense now. Why can `rand` not return `1`?

07_errors/jimbotech/storer_test.go
```diff
@@ -0,0 +1,116 @@
...
+	// Now modify the original and make sure the
+	// value in the store will not change
+	pup.Colour = black
+	// now check by reading the value back and compare
+	pup2, err2 := s.store.ReadPuppy(pup.ID)
+	s.Require().NoError(err2)
+	s.Equal("kelpie", pup2.Breed)
+	s.Equal(brown, pup2.Colour)
+	s.Equal("indispensable", pup2.Value)
+	s.True(pup2.Colour == brown)
```
camscale: Use `S.Equal` here. And all the other places you use `s.True(x == y)`

07_errors/jimbotech/storer_test.go
```diff
@@ -0,0 +1,116 @@
...
+const brown = "brown"
+const black = "black"
+const grey = "grey"
+
+func TestSuite(t *testing.T) {
+	suite.Run(t, &storesSuite{store: NewMapStore()})
+	suite.Run(t, &storesSuite{store: &SyncMapStore{}})
+}
+
+//SetupTest creates the correct empty map for each test
```
camscale: nitpick: space after double-slash please.

07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (s *storesSuite) TestDeleteFailure() {
+	err := s.store.DeletePuppy(1)
+	s.Require().Equal(ErrIDNotFound, err)
+}
+
+func (s *storesSuite) TestError() {
+	var err = &Error{}
+	err = nil
```
camscale, line 42: these two lines can be replaced with `var err *Error` - you specify the type and by not initialising it, you get the zero value, which is nil.

07_errors/jimbotech/error_test.go
```diff
@@ -0,0 +1,45 @@
...
+	err := s.store.DeletePuppy(-1)
+	s.Require().Equal(ErrValueBelowZero, err)
+}
+
+func (s *storesSuite) TestDeleteFailure() {
+	err := s.store.DeletePuppy(1)
+	s.Require().Equal(ErrIDNotFound, err)
+}
+
+func (s *storesSuite) TestError() {
```
camscale, line 40: Should be called `TestErrorNil` or something.

07_errors/jimbotech/mapstore.go
```diff
@@ -0,0 +1,74 @@
...
+	return len(s)
+}
+
+// CreatePuppy add a puppy to storage
+// but will modify the member ID.
+func (s MapStore) CreatePuppy(p *Puppy) (int32, error) {
+	if s == nil {
+		return 0, ErrNotConstructed
+	}
+	p.ID = rand.Int31()
```
camscale: wat. i congratulate you on trying to leave land mines for the future developers. I curse you as a possible future developer.

07_errors/jimbotech/mapstore.go
```diff
@@ -0,0 +1,74 @@
...
+	return &retVal, nil
+}
+
+// UpdatePuppy update your puppy store.
+func (s MapStore) UpdatePuppy(id int32, puppy *Puppy) error {
+	if s == nil {
+		return ErrNotConstructed
+	}
+	_, err := s.ReadPuppy(id)
+	if err != nil {
```
camscale: Since you do not need the value of the first return value (you're using `_`), idiomatic Go would put this on one line:
```
if _, err := s.ReadPuppy(id); err != nil {
```

07_errors/jimbotech/mapstore.go
```diff
@@ -0,0 +1,74 @@
...
+	return nil
+}
+
+// DeletePuppy remove the puppy from store.
+func (s MapStore) DeletePuppy(id int32) error {
+	if s == nil {
+		return ErrNotConstructed
+	}
+	_, err := s.ReadPuppy(id)
+	if err != nil {
```
camscale: ditto.

07_errors/jimbotech/storer.go
```diff
@@ -0,0 +1,57 @@
...
+func (e *Error) Error() string {
+	if e == nil {
+		return "<nil>"
+	}
+	return e.Message + " Error Code: " + strconv.Itoa(e.Code)
+}
+
+const (
+	// ErrNegativeID error number if id is negative
+	ErrNegativeID = -1
```
camscale: These codes should probably be named as `ecNegativeID` or something (ec == error code). Vars starting with `Err` are usually expected to be error types.

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
+package main
+
+import (
+	"math/rand"
+	"sync"
+)
+
+// SyncMapStore stores puppies threadsafe.
+type SyncMapStore struct {
+	mux sync.Mutex
```
camscale: `mux` is usually the name for a multiplexer, not a mutex. You see that name used a lot for `http.ServeMux`. I think `m` is fine as a name, as it is always used with `Lock()` or `Unlock()` after it - it is clear from usage that `m` is a mutex. Otherwise be verbose and call it `mutex`.

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
...
+	return &retPup, nil
+}
+
+// UpdatePuppy threadsafe update your puppy store.
+func (s *SyncMapStore) UpdatePuppy(id int32, puppy *Puppy) error {
+	_, err := s.ReadPuppy(id)
+	if err != nil {
+		return err
+	}
+	s.mux.Lock()
```
camscale: since `sync.Map` is already concurrency safe for individual operations, the only purpose of having a mutex as well is to provide concurrency safety over multiple operations on the map. Yet here you do not take the lock until after you do `ReadPuppy`. That takes and releases the lock, meaning the lock is not held over the duration of all the operations on the `sync.Map`.

The only way this can work then is to not use `ReadPuppy` in this method - you will need to take the lock and directly retrieve from the `sync.Map`

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
...
+	return nil
+}
+
+// DeletePuppy threadsafe removal of the puppy from store.
+func (s *SyncMapStore) DeletePuppy(id int32) error {
+	_, err := s.ReadPuppy(id)
+	if err != nil {
+		return err
+	}
+	s.mux.Lock()
```
camscale: ditto

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
...
+	})
+	return length
+}
+
+// CreatePuppy threadsafe adding a puppy to storage
+// but will modify the member ID.
+func (s *SyncMapStore) CreatePuppy(p *Puppy) (int32, error) {
+	s.mux.Lock()
+	defer s.mux.Unlock()
+	p.ID = rand.Int31()
```
camscale: you're evil.

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"math/rand"
+	"sync"
+)
+
+// SyncMapStore stores puppies threadsafe.
+type SyncMapStore struct {
+	mux sync.Mutex
+	sync.Map
```
camscale: by embedding the map, you are making the map methods directly accessible to users of `SyncMapStore`, allowing them to bypass any checks/validation you have. You should put the `sync.Map` in as an unexported field.

07_errors/jimbotech/sync_mapstore.go
```diff
@@ -0,0 +1,81 @@
+package main
+
+import (
+	"math/rand"
+	"sync"
+)
+
+// SyncMapStore stores puppies threadsafe.
+type SyncMapStore struct {
```
camscale: the filename should be `syncmapstore.go`. But I would call the type `SyncStore` because shorter is better. From the outside, no-one cares that it is a map - it's a store. So don't name it with "Map". Which means the filename could be `syncstore.go`

07_errors/jimbotech/storer.go
camscale: Don't set the error code to -1. Just use `iota` and leave the assignment off subsequent error codes. That let's Go assign the numbers since the actual numbers themselves are not relevant.

07_errors/jimbotech/storer.go
```diff
@@ -0,0 +1,57 @@
...
+
+// ErrValueBelowZero error generated if the calu is below zero
+var ErrValueBelowZero = &Error{Message: "id below 0", Code: ErrNegativeID}
+
+// ErrIDNotFound error if the requested ID is not in the store
+var ErrIDNotFound = &Error{Message: "id not found", Code: ErrNotFound}
+
+// ErrNotConstructed returned if the interface was called without
+// first constructing the underlaying structure.
+var ErrNotConstructed = errors.New("store not created")
```
camscale: Why is this not an `Error` like the other ones?

camscale (commented):
Some comments for you... I'll be back in a few months :)

It would be nice if you could squash your commits into a cohesive set. Leaving old broken commits is like renaming a function to `foo_old()`, fixing it as `foo()` and leaving `foo_old()` in the code.

Also, I edited your PR message to remove the space between `# 682` - that way, github creates it as a link.

07_errors/jimbotech/error_test.go
jimbosoft: Done

07_errors/jimbotech/error_test.go
jimbosoft: Done

07_errors/jimbotech/error_test.go
jimbosoft: Removed

07_errors/jimbotech/storer_test.go
jimbosoft: Deleted

07_errors/jimbotech/error_test.go
jimbosoft: Renamed

07_errors/jimbotech/storer_test.go
jimbosoft: Moved

07_errors/jimbotech/storer_test.go
jimbosoft: Done

07_errors/jimbotech/storer_test.go
jimbosoft: Fixed

07_errors/jimbotech/error_test.go
jimbosoft: Done

07_errors/jimbotech/error_test.go
jimbosoft: Renamed

07_errors/jimbotech/mapstore.go
jimbosoft: Previously I used uuid.New().ID() But unfortunately the uint will not turn into a int. This is in my view the next best thing. Similar in functionality. What is your concern? A repeat of the random number? The doco says: "Random numbers are generated by a Source. Top-level functions, such as Float64 and Int, use a default shared Source that produces a deterministic sequence of values each time a program is run." So I expect no repeats during during a run. Maybe I am misunderstanding your objection? What do you suggest?

jimbosoft: Added uniqueness check

07_errors/jimbotech/mapstore.go
jimbosoft: Done

07_errors/jimbotech/mapstore.go
jimbosoft: Done

07_errors/jimbotech/storer.go
jimbosoft: Done

07_errors/jimbotech/sync_mapstore.go
jimbosoft: Done

07_errors/jimbotech/sync_mapstore.go
jimbosoft: Not sure what motivated me to do the code this way. Too long ago. The ReadPuppy is called in order to produce the required errors. If I do what you suggest and directly retrieve from the the sync.Map I would have to replicate a lot of code. I guess I put the lock where I did as I wanted to hold it for the shortest possible time, but as you pointed out, another thread can get in between Read and the change. I will move the lock to the top of the function and separate out the error generating retrieval function into unsafeRetrievePuppy.

07_errors/jimbotech/sync_mapstore.go
jimbosoft: see above

07_errors/jimbotech/sync_mapstore.go
jimbosoft: Good call.
Done

07_errors/jimbotech/sync_mapstore.go
jimbosoft: Done

07_errors/jimbotech/storer.go
jimbosoft: Because god willed it so ... however if you will something else, that's fine with me too. May the gods argue with each other ... Done

07_errors/jimbotech/storer.go
jimbosoft: Interesting ... Done

07_errors/jimbotech/sync_mapstore.go
jimbosoft: As above, the gravity of my wickedness is still a mystery to me ... added uniqueness check ...

anzdaddy:
Please merge or close.

